PR #1, opened 2016-11-02
Extract queue_job module from connector

guewen:
* Replace ConnectorSession by odoo.api.Environment
* Rework job classes
* Jobs are now applied on Model methods

```python
  class MyModel(models.Model):
     _name = 'my.model'

     @api.multi
     @job
     def my_method(self, a, k=None):
         _logger.info('executed with a: %s and k: %s', a, k)


  class MyOtherModel(models.Model):
      _name = 'my.other.model'

      @api.multi
      def button_do_stuff(self):
          self.env['my.model'].with_delay().my_method('a', k=2)
```

See history on https://github.com/OCA/connector/pull/212


queue_job/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+ 'external_dependencies': {'python': ['requests'
+                                      ],
+                           },
+ 'data': ['security/security.xml',
+          'security/ir.model.access.csv',
+          'views/queue_job_views.xml',
+          'data/queue_data.xml',
+          ],
+ 'installable': True,
+ 'application': True,
```
lmignon: Is-it really an application?

lmignon:
@guewen What a great refactoring!!!!! No time for a full review yet. After a first read I have not found anything to complain. I need to investigate a little further to completly understand what is behind the changes and for exemple at https://github.com/OCA/queue/pull/1/files#diff-8b63facccbd0e181392cdc06745d034eR615

queue_job/__manifest__.py
guewen: Isn't it an application if we have a top-level menu for accessing the queue models (Channels, Job functions and Jobs). Corollary, would we want a top-level menu? (I'd say yes, and we need to find an icon BTW).

guewen:
I still have to test in _in real_ with `base_import_async` because it's only tested with the python tests for now :)

leorochael:
As a test, I've tried to migrate a little project that doesn't try to use the whole connector to use just this PR, but quickly ran into an issue: We also use the connector events subsystem...

It seems to me that many uses of the queue without the connector will use (or could benefit from the use of) the events as well.

Should we try to move it along into `queue` or should it perhaps have its own repo?

leorochael:
Speaking of moving modules into the queue repo, I think base_import_async would be a good candidate, as it uses solely the job queue.

Rewriting base_import_async to use the new model method delay would be a good way to taste what that api should feel like.

leorochael:
And speaking of model methods delay, if we add a method to the global BaseModel object that allows delaying methods, why not simply allow delaying all methods (with serializable parameters) by default?

That's what ERP5 does, by the way.

You could then just add decorators to the methods where you want to pass different queue parameters than the defaults (like priority, or channel name).

guewen:
> It seems to me that many uses of the queue without the connector will use (or could benefit from the use of) the events as well.

You might be right. However, it might be done in a subsequent pull request.

guewen:
> why not simply allow delaying all methods (with serializable parameters) by default?

I thought about that. Actually, I am _restricting_ to delay methods with the `@job` decorator so it would just need to remove this restriction. 

We should just think about the security. The `with_delay` does a `getattr` on the model, so I used the `@job` decorator as a whitelist of the jobs to run. Before removing the restriction, we must be sure that we don't open a security hole.

The other point is that currently we show the "job functions" in a view so we can configure the channels. It might be weird to have only some of them, but maybe that's just a matter of improving the way we present and name them.

lmignon:
> It seems to me that many uses of the queue without the connector will use (or could benefit from the use of) the events as well.

IMO the events subsystem is also a good candidate for a generic module not related at all to the connector module nor the queue_job module.

leorochael:
@lmignon I agree, but even as an independent module, should it be in a separate repo?

lmignon:
@leorochael IMO yes it should be in a separate repo with its own release management/cycle.

queue_job/__manifest__.py
blaggacao: Although they are, I would argue they shouldn't. I feel coining it "application" should account mainly for the business end user perspective. So, If I can't magically hire the best people in the world with this or similar stuff, then it's probably not an "application". This implies abstaining from top-level menu. Maybe it's better classified (hidden) in the "technical" settings menu.

guewen: @blaggacao if the user has not the accesses rights for the queue, he won't see the top-level menu, so for a technical end-user, it can make sense to have a top-level menu (jobs is a menu you open often as you regularly want to check them, it should not be hidden in an obscure place).

lmignon: @guewen your POV makes sense :+1:

leorochael: Even if it does have a top level menu, I'm not sure it's enough to call it an application.

Personally, I'd be ok if the Queue menu was somewhere inside Technical instead of being top-level.

If I need to keep an eye on the jobs queue all the time so that a top-level menu makes sense, then I'd argue something is wrong in my application...

guewen: Then you never tried to synchronize with Magento :joy: .
Joke apart, I'm not sure what exactly is supposed to be an Application. So given the controversy, I removed the flag.

blaggacao: Convincing and true. I was wrong.

adrienpeiffer:
Hi @guewen,

Thanks for the work !
I did some functional test on this one. I just created a PR (https://github.com/guewen/queue/pull/2) on your own branch to fix some things that appeared during my tests.

We can discuss on each of my commits if you want.

Thanks,

Adrien

guewen:
@adrienpeiffer @lmignon @sbidoul @leorochael @blaggacao  The last 2 commits rework the final bits of the jobs and related actions implementation. Tests are green on my computer. Waiting for the Travis build. Hopefully merged in 2016? :-)

guewen:
Aaaand it's green!

parthivgls:
👍 +1

queue_job/controllers/main.py
```diff
@@ -0,0 +1,132 @@
...
+
+        # TODO: set_started should be done atomically with
+        #       update queue_job set=state=started
+        #       where state=enqueid and id=
+        job.set_started()
+        job.store()
+        http.request.env.cr.commit()
+
+        _logger.debug('%s started', job)
+        job.perform()
```
sbidoul, line 62: Don't we need some kind of sudo(job.user_id) here?

queue_job/controllers/main.py
```diff
@@ -0,0 +1,132 @@
...
+            _logger.debug('%s postponed', job)
+
+        except (FailedJobError, Exception):
+            buff = StringIO()
+            traceback.print_exc(file=buff)
+            _logger.error(buff.getvalue())
+            job.env.clear()
+            with odoo.api.Environment.manage():
+                with odoo.registry(job.env.cr.dbname).cursor() as new_cr:
+                    new_env = api.Environment(new_cr, job.env.uid,
```
sbidoul: job.env.uid is always SUPERUSER_ID

queue_job/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+
+{'name': 'Job Queue',
+ 'version': '10.0.1.0.0',
+ 'author': 'Camptocamp,Acsone,Odoo Community Association (OCA)',
```
sbidoul: s,Acsone,ACSONE SA/NV,

queue_job/models/base.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @api.model_cr
+    def _register_hook(self):
+        """ register marked jobs """
+        super(Base, self)._register_hook()
+        job_methods = [method for __, method
+                       in inspect.getmembers(self, predicate=inspect.ismethod)
+                       if getattr(method, 'delayable', None)]
+        for job_method in job_methods:
+            self.env['queue.job.function']._register_job(job_method)
```
sbidoul, line 23: @guewen nice!

queue_job/controllers/main.py
guewen: the sudo is done in `perform()` that executes `self.func()`, the latter being the method call on the sudoed recordset: https://github.com/guewen/queue/blob/43bb875adc421f6773a07ae6ba143014d7a0e600/queue_job/job.py#L419

queue_job/controllers/main.py
guewen: Do you mean we should replace `job.env.uid` by `SUPERUSER_ID`? I replaced by `job.env(cr=new_cr)` which returns the new env with only the `cr` modified, avoiding some repetition.

queue_job/__manifest__.py
guewen: Fixed

sbidoul (approved):
Great job!

leorochael:
Hi @guewen, one thing I'm missing in this PR is the configuration file effort I did for `connector` on OCA/connector#213.

Since there wasn't a maintained 10.0 branch of connector yet, I ended up not porting it over before your queue_job extraction effort, even though it was already landed on 9.0 and it would obviously be trivial to make it work on 10.0.

That PR, along with odoo/odoo#13685, which got squash merged on odoo/odoo@5ee4344, means it should be possible to configure and launch a `queue_job` enabled odoo 10.0 entirely from `odoo.cfg` without having to resort to environment variables.

I'd like to see the config file work land on queue_job before this PR is merged, but I'm not sure when I'm going to find the [round tuits](https://en.wiktionary.org/wiki/round_tuit) to make it happen on my side, and I don't want to hold this PR hostage, so feel free to land it without.

The worse that could happen is that I'll make another PR later...

adrienpeiffer:
Hi Guewen,
Can you take a look on https://github.com/guewen/queue/pull/3 ?

adrienpeiffer:
Hi,

One last thing, I noticed that when an object in memory that is in parameter of the delayed method, and that this object is modified (like a dictionary). When the job fails, changes made to the dictionary remain persistent. Any comment ?

queue_job/README.rst
```diff
@@ -0,0 +1,167 @@
...
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to ``Job Queue`` menu
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/{repo_id}/{branch}
```
pedrobaeza: Put this information

queue_job/README.rst
```diff
@@ -0,0 +1,167 @@
...
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
```
pedrobaeza: Remove this

guewen:
@adrienpeiffer could you give an example?

leorochael:
@adrienpeiffer, this is normal behaviour in Python,  if you pass mutable objects into a function/method and the function mutates the object, the object remains mutated outside of the scope of the function, even if the function raises an exception.

Only ORM records know to revert state when a transaction is aborted.

The solution is either not to mutate the object in the delayed call, or to make a copy if you need to.

leorochael:
@guewen: guewen/queue#5 for the config file work.

adrienpeiffer:
@leorochael Of course. But in this case, these mutable objects are stored in the database when the job failed. So, I think that it's not really clean.

@guewen. For instance, I give a dictionary that contains all data necessary the create a new partner. When the job failed (for any reason) the dictionary (args, kwargs) of the delayed method contains other key 
in the dictionary like ```message_followers_ids``` or something like that.

Personally, I just make a deep copy of the dictionary in the delayed method when the job is running.

guewen:
@adrienpeiffer Nice catch! Thanks for the explanation, I could fix it in efddad9.

@leorochael Thanks for bringing this feature back. I added a comment on the pull request regarding the naming of the section `options-queue_job`.

sbidoul:
And when https://github.com/guewen/queue/pull/5 is merged, I'll do a PR for the sequential channels feature.

adrienpeiffer (approved):
:+1: Great work !

sbidoul:
@guewen here is the last piece of the puzzle: https://github.com/guewen/queue/pull/6

guewen:
Thanks @sbidoul. I'm eager to see this merged :)

sbidoul:
Voilà @guewen. À toi l'honneur du tweet.

blaggacao:
@guewen Congratulations! This is actually really exciting work. To give people like me a headstart, consider a possible use cases section in the readme? Like advertising it a little bit, in the words of a seasoned contributor?

leorochael:
We should consider adding a github-pages or ReadTheDocs auto-generated site next.

Also, we should probably port [base_import_async](https://github.com/OCA/connector-interfaces/tree/8.0/base_import_async) (and perhaps move it into this repo) as an example of what can be done with `queue_job`.

PR #7, opened 2017-02-01
Rationalize Job properties

leorochael:
- Extract the Job storage model attribute and use it consistently.

 - Turn `.dbrecord` and `.uuid` into their own auto-cached properties. Make them slightly better performant in the process.

I suggest evaluating one commit at a time.

### DIFF
--- a/queue_job/__manifest__.py
+++ b/queue_job/__manifest__.py
@@ -3,7 +3,7 @@
 
 
 {'name': 'Job Queue',
- 'version': '10.0.1.0.0',
+ 'version': '10.0.2.0.0',
  'author': 'Camptocamp,ACSONE SA/NV,Odoo Community Association (OCA)',
  'website': 'https://github.com/OCA/queue/queue_job',
  'license': 'AGPL-3',
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -180,6 +180,8 @@ class Job(object):
 
     """
 
+    job_model_name = 'queue.job'
+
     @classmethod
     def load(cls, env, job_uuid):
         """ Read a job from the Database"""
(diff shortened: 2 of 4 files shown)


queue_job/job.py
```diff
@@ -437,12 +467,10 @@ def description(self):
         else:
             return '%s.%s' % (self.model_name, self.func.__name__)
 
-    @property
+    @cached_truthy_property
```
guewen, line 470: Isn't usually called a lazy property? Couldn't the decorator be `lazy_property`?

queue_job/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 
 {'name': 'Job Queue',
- 'version': '10.0.1.0.0',
+ 'version': '10.0.2.0.0',
```
guewen, line 6: Wouldn't be sufficient to change to `10.0.1.1.0`?

guewen (commented):
> Move it to a class attribute so it's overridable in subclasses.

Only for my curiosity, do you have any use case / plans for that change?

leorochael:
@guewen wrote:
> > Move it to a class attribute so it's overridable in subclasses.
> Only for my curiosity, do you have any use case / plans for that change?

Not specifically, but I felt that the model name spread as a string literal throughout the code was a bug magnet.

I'd be content with a module global, but since it was already assigned as an instance attribute for some reason I could not determine, I felt it appropriate to keep it as a class attribute.

Someone could, in theory, create a subclass of `Job` to store jobs in a subclass of `queue.job` if they felt like it, but I have no such plans for now.

queue_job/job.py
leorochael: Well, since it only caches truthy values, `lazy_property` would not reflect all it does. Perhaps `lazy_truthy_property`? `truthy_memoize`? I'm open to suggestion, even `lazy_property` if you prefer.

BTW, there is an actual [`lazy_property`](https://pypi.python.org/pypi/lazy-property) Python distribution, but it behaves slightly differently regarding the cached value storage, and it stores anything, even non-truthy values. It would be nice to avoid confusion.

queue_job/__manifest__.py
leorochael: Well, it doesn't require a database migration, so perhaps `1.1.0` is sufficient.

I just wanted to call attention to the change of method into an attribute which could break code that depends on `queue_job`, but I can set it to `10.0.1.1.0` if you prefer.

PR #8, opened 2017-03-22
Stop parsing psycopg exceptions

leorochael:
Instead of trying to parse error messages from postgres, check the
existence of the module table with a non-exception-raising query.

The `unicode(err)` cast failed with postgres in locales that generated
error messages with non-ascii chars.

### DIFF
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -219,16 +219,14 @@ class Database(object):
 
     def _has_queue_job(self):
         with closing(self.conn.cursor()) as cr:
-            try:
-                cr.execute("SELECT 1 FROM ir_module_module "
-                           "WHERE name=%s AND state=%s",
-                           ('queue_job', 'installed'))
-            except psycopg2.ProgrammingError as err:
-                if unicode(err).startswith('relation "ir_module_module" '
-                                           'does not exist'):
-                    return False
-                else:
-                    raise
+            cr.execute("SELECT 1 FROM pg_tables WHERE tablename=%s",
+                       ('ir_module_module',))
+            if not cr.fetchone():
+                return False
+            cr.execute(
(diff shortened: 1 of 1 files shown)


queue_job/jobrunner/runner.py
```diff
@@ -219,16 +219,14 @@ def close(self):
...
-                else:
-                    raise
+            cr.execute("SELECT to_regclass('ir_module_module')")
+            [ir_module_module_exists] = cr.fetchone()
+            if not ir_module_module_exists:
+                return False
+            cr.execute(
+                "SELECT 1 FROM ir_module_module WHERE name=%s AND state=%s",
+                ('queue_job', 'installed')
+            )
```
guewen, line 229: Alternatively, we could check postgresql's error code, it's 42P01 for an undefined table. (https://www.postgresql.org/docs/8.1/static/errcodes-appendix.html). It's one query less, but also one error more in the logs. 

`to_regclass` is only available since PG9.4 though and I think lesser versions should be supported in Odoo 10.0

leorochael: What is the minimum version of Postgres we plan to support on `queue_job`. I tried to look up quickly minimum version requirements for Odoo but could not find a definitive answer.

guewen: Same... I could find a comment speaking about 9.3 used with Odoo 10.0.

On Travis, the default version is 9.1 (https://github.com/OCA/maintainer-quality-tools/issues/432#issuecomment-281580935).

I think that Odoo 10.0 is still compatible with 9.0+ and we should aim for the same compatibility.

leorochael: Well, the last version that is supported by PostgreSQL itself is 9.2.

I've changed the query to support older versions.

guewen (approved):
Thanks for the change!

leorochael:
Rebasing to squash the `fixup!`

PR #9, opened 2017-03-29
[10.0] [IMP] Allow to override the channel to use when delaying a job

lmignon:
The idea is to be able to specify the channel to use when delaying the job. 
```python
delayable = self.env['test.queue.job'].with_delay(channel='root.sub.sub')
```
If a channel is given as parameter to the `with_delay`method, it's used in place of the one declared as default_channel on the function to delay.


guewen (approved):
Nice one. Thanks

rousseldenis (approved):
:+1: Code Review + tested

PR #10, opened 2017-04-19
Allow to use datetime and date in job arguments

guewen:
Add JSON encoder and decoder for datetime and date values in the jobs so
we can use these types in job arguments.


lmignon (approved):
Thank you @guewen 
:+1: Code review

pedrobaeza (approved):

leemannd (approved):
:+1:

guewen:
Thanks for the fast review and merge :)

PR #16, opened 2017-06-29
[10.0][FIX] module name in web_icon

lmignon:
@guewen @pedrobaeza


pedrobaeza (approved):
Ouch!

pedrobaeza:
Fast-tracking as obvious bug

PR #17, opened 2017-07-17
[10.0][FIX] do not always load args of a job with sudo but with the user of the job

Cedric-Pigeon:
Because of the initial environnement all recordset  loaded by the job are with user Admin even if the job is planned by another user.

It can cause some failures in a multi-company context.

This PR suggest a fix for this.

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -195,6 +195,9 @@ class Job(object):
             raise NoSuchJobError(
                 'Job %s does no longer exist in the storage.' % job_uuid)
 
+        if stored.user_id != env.user:
+            stored = stored.sudo(stored.user_id)
+
         args = stored.args
         kwargs = stored.kwargs
         method_name = stored.method_name


Cedric-Pigeon:
@guewen Could you have a look at this please ?

guewen:
It should normally be handled here: https://github.com/OCA/queue/blob/10.0/queue_job/job.py#L440-L444
Can you see why it doesn't apply?

Cedric-Pigeon:
Yes, sorry you are right.

self.env.user is correct but on the import_batch method of connector magento, the recordset for the backend given in args is with sudo and I don't understand why.

guewen:
I see. The issue is in ``load``: https://github.com/OCA/queue/blob/10.0/queue_job/job.py#L198-L199
the args are read with the initial env. They should be read with the user context like we do for the `func` in a `@property`.

Cedric-Pigeon:
Thanks, I'll change my PR in this way.

Cedric-Pigeon:
@guewen Change done.

guewen:
Are we sure the user can always read all the fields of the job? _A priori_ I would say yes, but I'm not 100% sure. Are you? :)

Good catch by the way!

Cedric-Pigeon:
To do it the user must be member of queue manager group.

guewen:
To do what? A job can be delayed on behalf of any user and it should still work.

guewen:
@Cedric-Pigeon I pushed an alternative correction on #18 which is more accurate, because it keeps the user for every recordset used in args:

```
record.with_delay().export(backend.sudo(), partner)
```

will correctly deserialize backend as admin and partner as the current user

PR #18, opened 2017-07-18
Load job recordsets in args with original user

guewen:
When a job is enqueued with recordsets in args/kwargs, they are
serialized as json. When they are deserialized, until now the recorsets
were read using the admin user.  With this change, the original user is
included in the json, so we can read the recordset with the original user.


lmignon (approved):
@guewen Good catch! Thank your for the fix! (Code review)

Cedric-Pigeon (approved):

PR #19, opened 2017-08-09
[DOC] Improve README by showing user they need to import job

gdgellatly:
Hi @guewen 

I assume the import example is correct, I copied it from the test_queue_job module after my own module errored.


guewen:
It is, thanks!

PR #20, opened 2017-08-21
[ADD] queue_job module: migration from 9.0

cubells:
These changes are needed to upgrade connector module to 10.0

cc @Tecnativa

### DIFF
--- a/queue_job/migrations/10.0.1.0.0/pre-migration.py
+++ b/queue_job/migrations/10.0.1.0.0/pre-migration.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    # In order to migrate connector from v. 9.0 to v. 10.0, we need to set
+    # connector module state to 'to upgrade'
+    cr.execute("""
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

rafaelbn (approved):

lasley:
Curious - what's the rationale behind including the major version migrations here vs OpenUpgrade? If it's something fully supported by OCA, we migrate in the module?

pedrobaeza:
It's simple, this should be executed when this module gets upgraded. We can't include this in OpenUpgrade if we want to keep the modularity, or we will fill OpenUpgrade with a lot of `if openupgrade.is_module_installed(...` and the logic will be more complicated, as we must know where is the proper place to perform the upgrade.

lasley:
Totally makes sense, thanks Pedro

lmignon:
@pedrobaeza @cubells @rafaelbn You must revert this merge! The queue_job addon doesn't depend on 'connector' and don't exist in 9.0.

pedrobaeza:
This is for 10.0, not for 9.0.

guewen:
@pedrobaeza yes but how would that work? queue_job is a new addon in 10.0 so you'd install it and the migration won't be called. And if it is called you might not have the `connector` addon in your setup and you shouldn't be forced to use it because you are using `queue_job`.

pedrobaeza:
This will only be activated on migrations, not on regular installation, so there's no problem.

PR #24, opened 2017-08-29
[IMP] queue_job: OpenUpgrade migration script from 9.0

pedrobaeza:
Same as #20 with a renaming complement.

### DIFF
--- a/queue_job/migrations/10.0.1.0.0/pre-migration.py
+++ b/queue_job/migrations/10.0.1.0.0/pre-migration.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    # In order to migrate connector from v. 9.0 to v. 10.0, we need to set
+    # connector module state to 'to upgrade'
+    cr.execute("""
+        UPDATE ir_module_module
+        SET state='to upgrade'
+        WHERE name='connector'
+    """)
+    from openupgradelib import openupgrade
+    openupgrade.rename_xmlids(
+        cr, [
+            ('queue_job.group_connector_manager',
+             'queue_job.group_queue_job_manager',)
+        ],
+    )
--- a/queue_job/migrations/10.0.1.0.0/pre-migration.py
+++ b/queue_job/migrations/10.0.1.0.0/pre-migration.py
@@ -17,10 +17,13 @@ def migrate(cr, version):
         SET state='to upgrade'
         WHERE name='connector'
     """)
-    from openupgradelib import openupgrade
-    openupgrade.rename_xmlids(
-        cr, [
-            ('queue_job.group_connector_manager',
-             'queue_job.group_queue_job_manager',)
-        ],
-    )
+    try:
+        from openupgradelib import openupgrade
+        openupgrade.rename_xmlids(
+            cr, [
+                ('queue_job.group_connector_manager',
+                 'queue_job.group_queue_job_manager',)
+            ],
+        )
+    except ImportError:
+        pass
--- a/queue_job/migrations/10.0.1.0.0/pre-migration.py
+++ b/queue_job/migrations/10.0.1.0.0/pre-migration.py
@@ -7,9 +7,9 @@ def migrate(cr, version):
     if not version:
         return
     # In order to migrate connector from v9 to v10, we need to set
-    # connector module state to 'to upgrade'. If this is not done, all 
-    # connector.* xmlids are removed due to the module renaming done by 
-    # OpenUpgrade. In the future the approach sketched in 
+    # connector module state to 'to upgrade'. If this is not done, all
+    # connector.* xmlids are removed due to the module renaming done by
+    # OpenUpgrade. In the future the approach sketched in
     # https://github.com/OCA/queue/pull/23#issuecomment-325706811
     # may provide a more generic solution.
     cr.execute("""
--- a/queue_job/migrations/10.0.1.0.0/pre-migration.py
+++ b/queue_job/migrations/10.0.1.0.0/pre-migration.py
@@ -6,8 +6,12 @@
 def migrate(cr, version):
     if not version:
         return
-    # In order to migrate connector from v. 9.0 to v. 10.0, we need to set
-    # connector module state to 'to upgrade'
+    # In order to migrate connector from v9 to v10, we need to set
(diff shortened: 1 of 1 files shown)


queue_job/migrations/10.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    # In order to migrate connector from v. 9.0 to v. 10.0, we need to set
+    # connector module state to 'to upgrade'
```
sbidoul: I suggest adding, 'If this is not done, all connector.* xmlids are removed due to the module renaming done by OpenUpgrade. In the future the approach sketched in https://github.com/OCA/queue/pull/23#issuecomment-325706811 may provide a more generic solution.'

sbidoul (approved):

sbidoul:
Thanks Pedro. Can you remove the trailing whitespace to make pylint happy?

queue_job/migrations/10.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,26 @@
...
+    # connector.* xmlids are removed due to the module renaming done by 
+    # OpenUpgrade. In the future the approach sketched in 
+    # https://github.com/OCA/queue/pull/23#issuecomment-325706811
+    # may provide a more generic solution.
+    cr.execute("""
+        UPDATE ir_module_module
+        SET state='to upgrade'
+        WHERE name='connector'
+    """)
+    from openupgradelib import openupgrade
```
simahawk: isn't this going to fail if you don't have `openupgradelib` in your python env?

simahawk (changes requested):

queue_job/migrations/10.0.1.0.0/pre-migration.py
pedrobaeza: Of course, but it's a requirement for OpenUpgrade

simahawk: I'd say we should not assume that ppl are going to use it to do the migration. Can't we wrap it in a try/except ImportError and log the error?

hbrunn: finally somebody who [agrees with me](https://github.com/OCA/knowledge/pull/143#issuecomment-313422925)

pedrobaeza: My explanation continues the same that in that moment. The script is not going to be executed if you are not migrating.

hbrunn: and you're still mistaken in assuming that OpenUpgrade is the only way to migrate things

pedrobaeza: Yes, because this is OCA, and it's the official OCA migration engine. If you use another thing (any of us does), just install openupgradelib as a requirement. It's not a big deal.

simahawk: that's why I said "assume that ppl are going to use it to do the **migration**". I can do the migration myself or pay somebody to do it (even Odoo SA). 

You cannot force ppl to use this tool - even if it's the greatest tool in the world - especially if the price to pay is to use a try/except :) Plus, this is already a convention when you use external libs in odoo modules.

Frankly, I've see so many PRs reviews claiming for missing white spaces or commas or whatever fancy bit lacking...
this one is not one of them IMHO.

hbrunn: I totally agree that installing openupgradelib is not a big deal. And I don't say it shouldn't be used in migrations (I do think we should use it a lot more actually), I only say that if we use it in the sense that stuff explodes if it's not there, it's a requirement, and requirements belong in the module manifest and by extension into requirements.txt. Also no big deal to have it there ;-)

A nice side effect is that when you already have your module requiring openupgradelib, it's also safe to use it in migration scripts for minor updates.

pedrobaeza: Odoo is not going to migrate OCA modules. This is not going to fail unless you are migrating OpenUpgrade way. It's even inside the method after the check of the version.

But OK, this is peanuts. I'm going to put it for passing to the next picky claim.

pedrobaeza: @hbrunn the problem is then you are requiring that library for the normal fresh installation of the module where most people are not going to migrate in life (bad for them).

simahawk (approved):

queue_job/migrations/10.0.1.0.0/pre-migration.py
hbrunn: true. But I envision a world where basically any OCA module depends on openupgradelib, then we can harness its full power when radically changing the database layout somewhere in the lifecycle of a release. Actually, I sometimes find myself avoiding big changes just because I'd hate the idea to write a script like that without openupgradelib.

hbrunn:
thanks btw, if that's not clear from the discussion above :wink:

queue_job/migrations/10.0.1.0.0/pre-migration.py
pedrobaeza: Yeah, me too, but I don't want to modify the external dependencies of this module as it has been a total pain and they will put problems for that. I won't contribute anymore to this repository.

guewen:
Thanks all. I'm happy we reached a common agreement and that we learned more about openupgrade in the process.

queue_job/migrations/10.0.1.0.0/pre-migration.py
sbidoul: FWIW I personally think that
1. using openupgradelib in migration scripts is tremendously helpful
2. it's a lightweight dependency, I see no issue to depend on it explicitly and we must put it in external dependencies when we use it

guewen:
> Yeah, me too, but I don't want to modify the external dependencies of this module as it has been a total pain and they will put problems for that. I won't contribute anymore to this repository.

Sorry you see it like that.

queue_job/migrations/10.0.1.0.0/pre-migration.py
sbidoul: > I won't contribute anymore to this repository.

@pedrobaeza I really don't understand why you react like that. This repo is one of the best managed OCA repos. And the debate was nothing else than people (not exactly newbies btw) trying to get a deep understanding of a complex issue. And a good idea emerged of the discussion.

PR #25, opened 2017-09-07
[ADD] Take into account HTTP response codes when assigning work

guewen:
Port forwarding of #259


lmignon (approved):
Thank you @guewen for the forward port. 
I merge it since it's already approved in 8.0 https://github.com/OCA/connector/pull/259/files

PR #27, opened 2017-09-18
[FIX] queue_job: Complete migration script

pedrobaeza:
I don't want to let the migration script half-finished with a case I have found in one DB, but please don't be so picky.

Rationale: As there's a new computed field called `channel_method_name`, if the computation is done on module upgrade, maybe some models are not loaded yet, so we pre-create the column for avoiding this calculation and make it manually at the end of the process. We redo the computing code relaxing the constraints for avoiding also errors on missing models or methods.

@Tecnativa

### DIFF
--- a/queue_job/migrations/10.0.1.0.0/end-migration.py
+++ b/queue_job/migrations/10.0.1.0.0/end-migration.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2017 Tecnativa - Pedro M. Baeza
+# Copyright 2017-2018 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, SUPERUSER_ID
@@ -12,13 +12,18 @@ def migrate(cr, version):
     if not version:
         return
     env = api.Environment(cr, SUPERUSER_ID, {})
-    jobs = env['queue.job'].search([])
-    for job in jobs:
-        if job.model_name not in env:
+    QueueJob = env['queue.job']
+    groups = QueueJob.read_groups(
+        [], ['model_name', 'method_name'], ['model_name', 'method_name'],
+        lazy=False,
+    )
+    for group in groups:
+        if group['model_name'] not in env:
             continue
-        model = env[job.model_name]
-        method = getattr(model, job.method_name, False)
-        if method:
-            job.channel_method_name = '<%s>.%s' % (
+        model = env[group['model_name']]
+        method = getattr(model, group['method_name'], False)
+        QueueJob.search(group['__domain']).write({
+            'channel_method_name': '<%s>.%s' % (
                 method.im_class._name, method.__name__,
-            )
+            ),
+        })
--- a/queue_job/migrations/10.0.1.0.0/end-migration.py
+++ b/queue_job/migrations/10.0.1.0.0/end-migration.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, SUPERUSER_ID
+
+
+def migrate(cr, version):
+    """Now that everything has been loaded, compute the value of
+    channel_method_name.
+    """
+    if not version:
+        return
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    jobs = env['queue.job'].search([])
+    for job in jobs:
+        if job.model_name not in env:
+            continue
+        model = env[job.model_name]
+        method = getattr(model, job.method_name, False)
+        if method:
+            job.channel_method_name = '<%s>.%s' % (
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@guewen this one is ready for merging, tested on several migrations.

This completes the migration for being bullet-proof against missing models in the new version.

Please check.

guewen (approved):
Can't test but looks good to me. Thanks!

queue_job/migrations/10.0.1.0.0/pre-migration.py
```diff
@@ -25,5 +28,12 @@ def migrate(cr, version):
...
         )
+        # Pre-create column for avoiding the computation over non loaded models
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        openupgrade.add_fields(env, [
+            ('channel_method_name', 'queue.job', 'queue_job', 'char', False,
+             'queue_job'),
+        ])
     except ImportError:
-        pass
+        # Create at least the column, although without XML-ID
```
lmignon: @pedrobaeza Why do you mute this ImportError? IMO openupgradelib should be mandatory if the migration script runs.

pedrobaeza: @lmignon yeah I prefer that way, but others like @hbrunn or @jbeficent  have told me in other occasions that they don't want to have it as mandatory

hbrunn: that must have been a misunderstanding. I think it's fine to mandate openupgradelib, but then it must be a python dependency of the module and a requirement

pedrobaeza: Yeah, but I didn't want to put that dependency in the manifest as it's not required for normal module installation, but only on upgrades, so we are in a circular loop...

hbrunn (approved):

queue_job/migrations/10.0.1.0.0/pre-migration.py
hbrunn: ah, that was the point then. Yes, I firmly believe if it's not a requirement, it can't be required for anything. Positive review added ;-)

lmignon: I've no problem to not declare the dependency into the manifest since it's only used in some specific UC (migration from 9. to 10) but if it is not present in such circumstances the error must be raised.

pedrobaeza: OK, @lmignon, I have raise an exception if `openupgradelib` is not found.

lmignon (commented):
Thank you @pedrobaeza! We'll test the script.

lmignon (approved):
Functional test OK ping @guewen

simahawk (approved):

rousseldenis (approved):
Tested.

queue_job/migrations/10.0.1.0.0/end-migration.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def migrate(cr, version):
+    """Now that everything has been loaded, compute the value of
+    channel_method_name.
+    """
+    if not version:
+        return
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    jobs = env['queue.job'].search([])
```
StefanRijnhart: Please don't iterate over the full history of jobs. Even on my smallest customer using the connector I have over 250000 jobs. Can you compute the string for each model once and then assign in SQL?

pedrobaeza: No, each one is unique depending on its model and arguments. If we want to have a proper value, it should be done this way. We can fill with a default value, but then new jobs will be different from old ones.

rousseldenis: @pedrobaeza Maybe using a read_group on model_name. This can drastically improve assignation, no ?

pedrobaeza: The problem is that they are not ORM/SQL fields, but you can see at the code, Python attributes, got dynamically. I will try to improve this grouping by `model_name, method_name` anyway.

pedrobaeza:
@StefanRijnhart @rousseldenis look if it's better now.

rousseldenis:
Seems ok! Thanks

queue_job/migrations/10.0.1.0.0/end-migration.py
```diff
@@ -0,0 +1,29 @@
...
+
+def migrate(cr, version):
+    """Now that everything has been loaded, compute the value of
+    channel_method_name.
+    """
+    if not version:
+        return
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    QueueJob = env['queue.job']
+    groups = QueueJob.read_groups(
```
rousseldenis, line 16: @pedrobaeza read_group without 's'

StefanRijnhart (approved):
Thanks!

pedrobaeza:
I have fixed the typo in https://github.com/OCA/queue/commit/e116cff10c8463f66c866df39c3dfa10b70d522f

PR #30, opened 2017-10-02
v11 migration

lepistone:


guewen (approved):
We worked as a pair of this one, so we mutually reviewed each other work during the process. Thanks :)

guewen:
We still have a few travis issues to fix.

gdgellatly (approved):
@lepistone @guewen I've statically reviewed in detail and there is nothing that I can see that warrants even a comment.  All looks good.  Pretty straightforward port, adapting to a few changes in v11 which we knew about and py3.  I'd still like to see it all green obviously, and to give it a proper manual functional test, but my use cases involve modules that aren't ported to v11 yet so its a bit of cart before horse.

I think we should consider a third review on this if possible, only because self-review (i.e. guewen both porting and approving) is not really a good practice.

queue_job/jobrunner/channels.py
```diff
@@ -133,7 +134,7 @@ class ChannelJob(object):
 
     Here are some examples.
 
-    j1 comes before j2 before it has a smaller date_created
+    j1 comes before j2 because it has a smaller date_created
```
gdgellatly: @lepistone a smaller -> an earlier - OK one comment non blocking

lepistone: fixed, thanks!

simahawk (approved):
code LGTM

guewen:
Travis is green! :smiley:

guewen:
We'll rebase to remove the fixups before the merge

pedrobaeza:
Congratulations

sbidoul:
It's now on pypi.

PR #31, opened 2017-10-03
Migrate v11 queue job subscribe

guewen:
~~Depends of #30~~

### DIFF
--- a/queue_job_subscribe/__manifest__.py
+++ b/queue_job_subscribe/__manifest__.py
@@ -2,7 +2,7 @@
 # Copyright 2016 Cédric Pigeon
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {'name': 'Queue Job Subscribe',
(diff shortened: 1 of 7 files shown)


gdgellatly (approved):
@guewen I only reviewed files in these commits https://github.com/OCA/queue/pull/31/files/85af69bf6cc064243aff0756848d28b2116297d1..6eac2b205d4db38afb8551d8306a8ee3ab9e0897 .

In terms of the port it looks fine, obviously its a simple one, although I noticed in res_users that we have Jobs Notifications as the string.  To be honest I don't know if its right or not, but looks wrong given the help string refers to Job Notifications.

guewen:
@gdgellatly Thanks! I added a fixup commit, changing the string to "Job Notifications". I don't know either which was correct but at least it's consistent.

lepistone (approved):

atchuthan (approved):

simahawk (approved):
runbot build re-triggered

PR #35, opened 2017-10-13
[8.0][IMP]: 8.0 backports from 10.0 branch

shepilov-vladislav:
Requied only patch original Odoo to support 'base' model

After https://github.com/odoo/odoo/blob/8.0/openerp/models.py#L592
```
        # all models except 'base' implicitly inherit from 'base'
        if name != 'base':
            parents = list(parents) + ['base']

```

And after https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/ir/ir_model.py#L64

```
#
# IMPORTANT: this must be the first model declared in the module
#
class Base(models.AbstractModel):
    """ The base model, which is implicitly inherited by all models. """
    _name = 'base'
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ addons:
 
 env:
   global:
-  - VERSION="10.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
+  - VERSION="8.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
   - TRANSIFEX_USER='transbot@odoo-community.org'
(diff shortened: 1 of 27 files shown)


shepilov-vladislav:
Can you create new branch?

guewen:
In 8.0, the queue job was integrated with the connector (https://github.com/OCA/connector/tree/8.0) addon. Thanks for your proposal, but there is no plan to support a backport of the `queue` module in 8.0 as it would conflict with the `connector` addon. Besides, addons should not need patches to work.

PR #41, opened 2017-12-02
[10.0]  retry_postpone using new cursor

2zx:
I've experienced this error 

> _current transaction is aborted, commands ignored until end of transaction block_

trying to requeue a job in case of a failure due to SQL errors (typically concurrency errors)
I think the problem can be solved with the use of a new cr as it is done in the case of a failed job

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -71,10 +71,15 @@ class RunJobController(http.Controller):
         env = http.request.env(user=odoo.SUPERUSER_ID)
 
         def retry_postpone(job, message, seconds=None):
-            job.postpone(result=message, seconds=seconds)
-            job.set_pending(reset_retry=False)
-            job.store()
-            env.cr.commit()
+            env.cr.rollback()
+
+            with odoo.api.Environment.manage():
+                with odoo.registry(job.env.cr.dbname).cursor() as new_cr:
+                    job.env = job.env(cr=new_cr)
+                    job.postpone(result=message, seconds=seconds)
+                    job.set_pending(reset_retry=False)
+                    job.store()
+                    new_cr.commit()
 
         job = self._load_job(env, job_uuid)
         if job is None:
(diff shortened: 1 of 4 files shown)


queue_job/controllers/main.py
```diff
@@ -71,10 +71,16 @@ def runjob(self, db, job_uuid, **kw):
         env = http.request.env(user=odoo.SUPERUSER_ID)
 
         def retry_postpone(job, message, seconds=None):
-            job.postpone(result=message, seconds=seconds)
-            job.set_pending(reset_retry=False)
-            job.store()
-            env.cr.commit()
+
+            job.env.cr.rollback()
```
simahawk, line 75: shouldn't you rollback only in case of errors? /cc @guewen

guewen: I think it's there to prevent a conflict between the job.env's transaction and the new one. The job record might have been updated by the first one and will be updated by the second one? It's not called in the FailedJobError case though so why is it necessary here?

queue_job/jobrunner/runner.py
```diff
@@ -188,7 +188,8 @@ def urlopen():
         try:
             # we are not interested in the result, so we set a short timeout
             # but not too short so we trap and log hard configuration errors
-            response = requests.get(url, timeout=1)
+            # CGT: timeout must be hight enought to wait for server startup
+            response = requests.get(url, timeout=15)
```
guewen: Why is it required for you? What's your setup?

guewen (commented):
Do you use this in production? Since when and do you see any side-effect?

yvaucher (approved):
Tried it and works fine

I didn't experience anymore

`current transaction is aborted, commands ignored until end of transaction block`

guewen:
Great thanks @yvaucher for the test and @2zx for the patch!
@2zx would you mind to remove the change on `response = requests.get(url, timeout=15)` so we keep only the correction related to the `current transaction is aborted, commands ignored until end of transaction block` error?

yvaucher:
@2zx Could you make the autovacuum improvment on an other PR?

By the way there is still an open question from @guewen about the changes on the timeout.

Are you willing to finalize this Pull request?

2zx:
@yvaucher all unnecessary commits to this pull request have been reverted, sorry for the delay

guewen (approved):
Thanks a lot for your contribution!

guewen:
Note: to be squashed on merge

PR #46, opened 2018-01-24
Add default related action to open related records in view

guewen:
Add default related action to open related records.

The related action button now opens by default the record on which the job is working. It uses the default views of the job. The exact behavior depends of the number of records in the job:

* No record (job called on a  model): display a message indicating that there is no available action
* One record: open the default view in form
* Several records: open the default view in list

The related action can still be overridden by the 'related_action' decorator.

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -528,15 +528,20 @@ class Job(object):
             self.result = result
 
     def related_action(self):
-        if not hasattr(self.func, 'related_action'):
-            return None
-        if not self.func.related_action:
-            return None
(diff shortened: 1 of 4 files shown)


guewen:
thoughts:
* set it as default related action when it's not set?
* add keyword arguments to configure for instance a default xmlid for the view (e.g. invoices have several views)?

guewen:
This one is ready :)

lmignon (approved):
Than you @guewen It's a nice usability improvement!

simahawk (approved):
LG

PR #47, opened 2018-01-26
[IMP] avoid base_sparse_field --load

simahawk:
`base_sparse_field` patches odoo fields
in order to be able to use `fields.Serialized`.

To "pay" this shortcut you have to load it at server load
which makes harder to test it with external tools (ie: pytest)
and disorient first users as they always miss this step.

Now we load the field straight from the module: easy, neat, works.

### DIFF
--- a/queue_job/README.rst
+++ b/queue_job/README.rst
@@ -76,7 +76,7 @@ Configuration
 
   - if ``xmlrpc_port`` is not set: ``ODOO_QUEUE_JOB_PORT=8069``
 
- * Start Odoo with ``--load=web,base_sparse_field,queue_job``
+ * Start Odoo with ``--load=web,queue_job``
    and ``--workers`` greater than 1. [1]_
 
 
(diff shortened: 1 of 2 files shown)


guewen (approved):
Yeaaah! :fireworks: 

Thanks!

simahawk:
NOTE: runbot is yellow for:
```
2018-01-26 08:50:33,923 166 INFO openerp_test odoo.models: Storing computed values of queue.job.function.channel
<string>:98: (WARNING/2) Cannot analyze code. No Pygments lexer found for "none".
2018-01-26 08:50:35,029 166 WARNING openerp_test odoo.addons.base.module.module: docutils' system message present: <system_message level="2" line="98" source="<string>" type="WARNING"><paragraph>Cannot analyze code. No Pygments lexer found for "none".</paragraph><literal_block xml:space="preserve">.. code-block:: none
```

PR #49, opened 2018-02-06
[FIX] Get attributes on the class instead of recordsets

Garamotte:
Even if the recordset is empty, this is still an instance of the class, so the @property methods are called. As this type of method is usually run on a single record, they need self.ensure_one(), which crashes when evaluated on a recordset, preventing the server to start when queue_job is installed.

Getting the attributes on the class itself instead of using an empty recordset avoids evaluating property methods.


simahawk (approved):
just code review: LGTM

guewen (approved):

lmignon (approved):
Thank you @sylvain-garancher

PR #50, opened 2018-02-06
[FIX] Get attributes on the class instead of recordsets

Garamotte:
Even if the recordset is empty, this is still an instance of the class, so the @property methods are called. As this type of method is usually run on a single record, they need self.ensure_one(), which crashes when evaluated on a recordset, preventing the server to start when queue_job is installed.

Getting the attributes on the class itself instead of using an empty recordset avoids evaluating property methods.

This reverts commit 49d8f37b856dc2159ea4cc6e8c0cf113efd6f5a9, which was a hack for a single attribute name instead of a global fix.

### DIFF
--- a/queue_job/models/base.py
+++ b/queue_job/models/base.py
@@ -16,18 +16,11 @@ class Base(models.AbstractModel):
     def _register_hook(self):
         """ register marked jobs """
         super(Base, self)._register_hook()
-        job_methods = set()
-        for attr_name in dir(self):
-            # _cache on models is a lazy_property which raises an
-            # AssertionError when called on a empty model, just skip it.
-            if attr_name == '_cache':
-                continue
-            try:
-                attr = getattr(self, attr_name)
-            except AttributeError:
-                continue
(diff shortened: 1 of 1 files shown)


guewen (approved):
Clever! Frustrated I didn't thought about it, but you make my day 😂

lmignon (approved):

queue_job/models/base.py
```diff
@@ -16,18 +16,11 @@ class Base(models.AbstractModel):
...
-                continue
-            try:
-                attr = getattr(self, attr_name)
-            except AttributeError:
-                continue
-            if inspect.ismethod(attr) and getattr(attr, 'delayable', None):
-                job_methods.add(attr)
+        job_methods = [
+            method for __, method
+            in inspect.getmembers(self.__class__, predicate=inspect.ismethod)
```
guewen, line 21: In the class, the methods are not yet methods but functions, because what we are inspecting at this point is the decorated function I guess.

PR #56, opened 2018-02-09
queue_job_cron: do not require read access to ir.cron

### LINKED ISSUE #48: User in cron-job needs read-right ir.cron document after install of queue_job_cron
Before installing queue_job_cron the user running the cron job could be anyone who had the access rights to the object of the job. After installing the user also needs read rights to ir.cron. This is irrespective of how the job is running, job_queue or not. Is this on purpose?

sbidoul:
fixes #48

### DIFF
--- a/queue_job_cron/models/ir_cron.py
+++ b/queue_job_cron/models/ir_cron.py
@@ -41,7 +41,7 @@ class IrCron(models.Model):
 
     @api.model
     def _callback(self, model_name, method_name, args, job_id):
-        cron = self.env['ir.cron'].browse(job_id)
+        cron = self.env['ir.cron'].sudo().browse(job_id)
         if cron.run_as_queue_job:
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Shouldn't the job should be run with the user specified in the cron?

sbidoul:
@pedrobaeza I think it does, see https://github.com/OCA/queue/issues/48#issuecomment-364058572.

This sudo is only to read the ir.cron properties (ie the channel name), not to run it.

pedrobaeza (approved):
OK, thanks for the clarification. As always, it would be desirable a version bump

simahawk (approved):

guewen (approved):

Cedric-Pigeon (approved):

sbidoul:
Merged and bumped version.

PR #59, opened 2018-02-22
[10.0] [ADD] ability to set jobs to done in batch

ecino:
In our production database, we have a lot of failed jobs that we only need to put as done, because there is no real other action to do. For us, it's boring to go through all the form views and mark the jobs as done.

We therefore propose to add a small wizard to mark as done in batch, as it already exists for requeing jobs.

### DIFF
--- a/queue_job/views/queue_job_views.xml
+++ b/queue_job/views/queue_job_views.xml
@@ -150,7 +150,22 @@
                 </group>
                 <footer>
                     <button name="requeue" string="Requeue" type="object" class="oe_highlight"/>
-                    or
+                    <button string="Cancel" class="oe_link" special="cancel"/>
+                </footer>
+            </form>
+        </field>
+    </record>
+
+    <record id="view_set_jobs_done" model="ir.ui.view">
+        <field name="name">Requeue Jobs</field>
+        <field name="model">queue.jobs.to.done</field>
+        <field name="arch" type="xml">
+            <form string="Set jobs done">
+                <group string="The selected jobs will be set to done.">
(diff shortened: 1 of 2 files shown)


ecino:
@guewen do you think this could be needed to all?

queue_job/views/queue_job_views.xml
```diff
@@ -157,6 +157,23 @@
...
+        <field name="name">Requeue Jobs</field>
+        <field name="model">queue.jobs.to.done</field>
+        <field name="arch" type="xml">
+            <form string="Set jobs done">
+                <group string="The selected jobs will be set to done.">
+                    <field name="job_ids" nolabel="1"/>
+                </group>
+                <footer>
+                    <button name="set_done" string="Set to done" type="object" class="oe_highlight"/>
+                    or
```
yvaucher: The `or` is the old fashion design of wizards in version 7.0 :)

yvaucher (approved):
Make sense, and you can anyway requeue any jobs in done state, thus it gives you a better control.

Current work around are either to edit state in DB or to install mass_update module.

In batch changes I've also find myself doing a lot of `eta` edition.  But that would be for an other feature.

simahawk (approved):

guewen (approved):
Thanks. If you can remove the `or` in the wizard it'd be very nice.

lmignon (approved):
@ecino Thank you for this useful improvement. Can you remove the 'or' so we can merge your changes?

ecino:
@lmignon @guewen Done!

PR #61, opened 2018-02-27
[10.0][FIX] queue_job:Issue when create channel from UI.

vjd8866:
- [ FIX ]

### DIFF
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -267,8 +267,8 @@ class JobChannel(models.Model):
     @api.depends('name', 'parent_id.complete_name')
     def _compute_complete_name(self):
         for record in self:
-            # if not record.name:
-            #     return  # new record
+            if not record.name:
+                return  # new record
             channel = record
             parts = [channel.name]
             while channel.parent_id:


guewen:
Hi,

Thanks for the fix! I think a better fix would be to have `required=True` on the name field.

guewen (changes requested):
The correct solution is to add a required on the name

guewen:
> The correct solution is to add a required on the name

Wrong :-/

@vdo-serpentcs did you sign the CLA?

queue_job/models/queue_job.py
```diff
@@ -267,8 +267,8 @@ class JobChannel(models.Model):
     @api.depends('name', 'parent_id.complete_name')
     def _compute_complete_name(self):
         for record in self:
-            # if not record.name:
-            #     return  # new record
+            if not record.name:
+                return  # new record
```
pedrobaeza, line 271: Should it be `continue`? Or better to use `self.filtered('name')` in the loop

guewen: Good catch :) (however continue is better than `self.filtered` as anyway have to loop on the records, `filtered` would use 2 loops)

guewen:
Considering the change needed and my uncertainty on the CLA, I'm opening #95.

Thanks @vdo-serpentcs!

vjd8866:
@guewen No i have not signed CLA

vjd8866:
> > The correct solution is to add a required on the name
> 
> Wrong :-/
> 
> @vdo-serpentcs did you sign the CLA?

@guewen I have signed CLA but how can I know it's approved, I don't get any confirmation.

Issue #57, opened 2018-02-15
Blueprint: limit redundant jobs


guewen:
Draft document

# Introduction
The idea would be to prevent multiple jobs to be created many times to do the same thing, even if the thing  to do is idempotent. Read the thread #52 for the original discussion.
An earlier tentative to get this was done in https://github.com/OCA/connector/pull/13
This blueprint is a tentative to get the feature right.

# Proposal
Currently, if I create 2 jobs doing the same thing, it will create 2 jobs which will both be executed.
```python
record.with_delay().do_stuff()
record.with_delay().do_stuff()
# => do stuff
# => do stuff
```
The use case would be that if I run `record.with_delay().do_stuff()`, even in different transaction, more than once only one job would be created. We should probably consider only `pending`, `enqueued` or `started` jobs, which means that if a job has been executed (state = done) and we call again `record.with_delay().do_stuff()`, then a new job will be created.

A sensible approach is to create a hash assigned to a job, built from whatever parts of the job constitute a unique key to identify this job. This identity key should be defined by the caller, or be part of the job definition:

```python
record.with_delay(identity_key='e294e8444453b09d59bdb6efbfec1323').do_stuff()
record.with_delay(identity_key='e294e8444453b09d59bdb6efbfec1323').do_stuff()
# => do stuff
```

```python

from odoo.addons.queue_job import job, build_hash

def generate_identity(func, *args, **kwargs):
    return build_hash(args[0].id, args[1]['a_field'])

@job(identity_key=generate_identity)
def do_something(self, data):
    print(data)
```

Then, at the creation of the job, a check would be done on `(job name, identity key)` and prevent the creation of the job if it exists. 
There is a drawback on performance, but which should be limited if the field job name + identity key are indexed, furthermore no check will be done if no identity key is defined so it doesn't change if we don't use the feature.

One remaining issue is that 2 concurrent transactions will not be able to see that each other create the same job and we will end up with 2 jobs with the same identity key.

guewen:
Done in https://github.com/OCA/queue/pull/66 and https://github.com/OCA/queue/pull/99

Issue #52, opened 2018-02-08
[Feature]Prevent duplicate Task in the queue


flotho:
Hi,

From time to time, jobs with the same Task signature are created by different methods.
This cause the queue to increase in an useless way.
Won't it be a good idea to prevent to create jobs if another one is existing in either 'pending' or 'started' state?
@guewen , what's your point of view on this one?

regards

guewen:
@flotho The logic whether a job should be created or not depending if another job exists or not changes from job to job. At best we could add a way for devs to provide a function attached to the job, that would define this logic.

Keep in mind that jobs should be idempotent, and in some cases they can leave early if the work has been done already. I know this is not the same thing, but it works well enough usually in the use cases I had to implement. Anyway you have to take that in account because you could only merge "pending"  "enqueued" or "started" jobs, so a identical job may already have been executed. On one side you lose some time executing a job that's already be done, on the other side, you slow down the creation of the job with extra checks. I prefer a fast delaying of jobs in the users transactions at the cost of a slower execution of those jobs in the background. 

Do you have examples where you would really benefit from this feature? If yes, I'd welcome a pull request implementing a smart solution, which keep the current behavior by default, but allow a developer to define merge conditions for pending jobs.

flotho:
Hi @guewen ,

Great thanks for this answer. Actually I understand your point of view that create fast and deliver fast will be the best. 
As an example I managed 2 instance in odoo 8 and odoo9 with prestashop connector. (BTW I made some little proposal in the OCA repo that could be usefull).
The context is really simple, my customer succeed in its business and the number of order is increasing a lot. In the same way they develop their B2B activity and also have sale.order that have many, many lines... (more than a hundred).
So we have 2 channels, one coming from PS and the other form Odoo itself, and each validation triggered a job to update the stock on PS. If you work with 5 workers for the main channel, during high activities, products coming from the B2C channel could contain products coming from B2B channel so that in the queue you'll have 2 jobs and you could also have more than duplicates etc...
Additionally, if you consider that the products are created in PS with huge category tree structure, each product will have to check/create dependencies so you'll have a high volume of jobs.
All this stuff is slowing down all my process and from time to time the queue has to be reset manually.
In this case, I'm wondering if it won't be a good proposal to make the test in the run method so that each job won't run itself if he find another one having the same signature. In this case, it will skip itself and mostly set it to the done state.
What does it sounds like to you ?

guewen:
Closed by https://github.com/OCA/queue/pull/66

PR #66, opened 2018-04-26
[10.0] Add identity key on job to allow limiting redundant execution

TDu:
This to implement a solution to the issue raised on issue #57

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -52,13 +52,15 @@ class DelayableRecordset(object):
     """
 
     def __init__(self, recordset, priority=None, eta=None,
-                 max_retries=None, description=None, channel=None):
+                 max_retries=None, description=None, channel=None,
+                 identity_key=None):
         self.recordset = recordset
         self.priority = priority
         self.eta = eta
         self.max_retries = max_retries
         self.description = description
         self.channel = channel
+        self.identity_key = identity_key
 
     def __getattr__(self, name):
         if name in self.recordset:
@@ -82,7 +84,8 @@ class DelayableRecordset(object):
                                max_retries=self.max_retries,
                                eta=self.eta,
                                description=self.description,
-                               channel=self.channel)
+                               channel=self.channel,
+                               identity_key=self.identity_key)
         return delay
 
     def __str__(self):
@@ -185,6 +188,12 @@ class Job(object):
         The complete name of the channel to use to process the job. If
         provided it overrides the one defined on the job's function.
 
+    .. attribute::identity_key
+
+        A key referencing the job, multiple job with the same key will not
+        be added to a channel if the existing job with the same key is not yet
+        started or executed.
+
     """
 
     @classmethod
@@ -233,22 +242,38 @@ class Job(object):
         job_.max_retries = stored.max_retries
         if stored.company_id:
             job_.company_id = stored.company_id.id
+        job_.identity_key = stored.identity_key
         return job_
+
+    @classmethod
+    def exist_duplicate_job(cls, env, identity_key):
+        """Check if a job to be executed with the same key exists."""
+        jobs = env['queue.job'].search(
+            [('identity_key', '=', identity_key),
+             ('state', 'in', [PENDING, ENQUEUED])],
+            limit=1
+        )
+        return bool(len(jobs))
 
     @classmethod
     def enqueue(cls, func, args=None, kwargs=None,
                 priority=None, eta=None, max_retries=None, description=None,
-                channel=None):
+                channel=None, identity_key=None):
         """Create a Job and enqueue it in the queue. Return the job uuid.
 
         This expects the arguments specific to the job to be already extracted
         from the ones to pass to the job function.
 
         """
+
+        recordset = func.im_self
+        env = recordset.env
+        if identity_key and cls.exist_duplicate_job(env, identity_key):
+            return
         new_job = cls(func=func, args=args,
                       kwargs=kwargs, priority=priority, eta=eta,
                       max_retries=max_retries, description=description,
-                      channel=channel)
+                      channel=channel, identity_key=identity_key)
         new_job.store()
         _logger.debug(
             "enqueued %s:%s(*%r, **%r) with uuid: %s",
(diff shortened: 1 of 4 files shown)


yvaucher (approved):

queue_job/job.py
```diff
@@ -185,6 +188,11 @@ class Job(object):
         The complete name of the channel to use to process the job. If
         provided it overrides the one defined on the job's function.
 
+    .. attribute::identity_key
+
+        A key referencing the job, multiple job with the same key will not
+        be added to a channel.
```
guewen: Worth to add that it concerns only jobs not yet executed

queue_job/models/base.py
```diff
@@ -60,4 +60,5 @@ def with_delay(self, priority=None, eta=None,
                                   eta=eta,
```
guewen, line 63: Can you add the argument in the docstring?

queue_job/job.py
```diff
@@ -233,22 +241,38 @@ def load(cls, env, job_uuid):
...
+        job_.identity_key = stored.identity_key
         return job_
 
+    @classmethod
+    def exist_duplicate_job(cls, env, identity_key):
+        """Check if a job to be executed with the same key exists."""
+        jobs = env['queue.job'].search(
+            [('identity_key', '=', identity_key),
+             ('state', 'in', [PENDING, ENQUEUED, STARTED])],
+            limit=1
```
guewen: We must have a partial index on the state and identity key:

```
CREATE INDEX queue_job_identity_key_state_partial_index ON queue_job (identity_key) WHERE state in ('pending', 'enqueued', 'started');
```
It can be created in the `init` method of the module.

guewen (changes requested):
Great, I didn't expected to have a PR on this so soon, thanks for your work :)

We should document somewhere the limitations (and the link to the blueprint #57): job is still created if another job with the same identity key is done or failed and it won't prevent another job to be created with the same identity key in another transaction if they happen at the same time. And then ultimately jobs must still be idempotent.

@flotho Could you have a look as you reported #52

simahawk (approved):
beside Guewen's remarks LGTM

queue_job/job.py
```diff
@@ -233,22 +241,38 @@ def load(cls, env, job_uuid):
...
             job_.company_id = stored.company_id.id
+        job_.identity_key = stored.identity_key
         return job_
 
+    @classmethod
+    def exist_duplicate_job(cls, env, identity_key):
+        """Check if a job to be executed with the same key exists."""
+        jobs = env['queue.job'].search(
+            [('identity_key', '=', identity_key),
+             ('state', 'in', [PENDING, ENQUEUED, STARTED])],
```
yvaucher: tl; dr: Please remove `STARTED` from the search

----

On a second though, I'm wondering if we want all states. If the process is quite long. Wouldn't we have an issue when we dismiss the creation of a new job if one is already `STARTED`?

What I have in mind is for an instance:
- an update on record X creates a job to export
- later the job A starts
- while job A is in started state, a second update is made on record X
- it doesn't trigger a new export job

Thus you will have a mismatch between local recordA and it's exported values.
And you won't have any job to update it. And might never have another update on that record.

This could happen if the current job is running on outdated values.

I see 2 options to avoid this.

1. destroy/create or restart the existing job

When next job should do exactly the same we want to do a retry with udpated data.
(as far as I know an ongoing job won't be killed by changing the state so it is more a do again than a restart)

Or could we imagine interrupting the ongoing job to really restart it? But we would have to add some mechanisms to interrupt nicely.

2. don't consider `STARTED` as duplicate

This would clearly mean sometime doing the same thing and create a new job. But it would close the breach.
Having 2 jobs would show more clearly that you did 2 calls, even if the first was useless.

TDu: I think @yvaucher is correct and a started job should not be considered as a duplicate.

guewen: > destroy/create or restart the existing job

No please :dagger: 

> don't consider STARTED as duplicate

Yes

queue_job/job.py
TDu: @guewen Do you mean as a post_init_hook in the Odoo module ?

guewen: No in the `init` method, you can find examples in odoo's code, example in the mail addon:

```
    @api.model_cr
    def init(self):
        self._cr.execute('SELECT indexname FROM pg_indexes WHERE indexname = %s', ('mail_channel_partner_seen_message_id_idx',))
        if not self._cr.fetchone():
            self._cr.execute('CREATE INDEX mail_channel_partner_seen_message_id_idx ON mail_channel_partner (channel_id,partner_id,seen_message_id)')
```

TDu: I see, thanks for your answer.

TDu:
The last fixup commit answers the previous remarks.

queue_job/models/queue_job.py
```diff
@@ -83,6 +83,21 @@ class QueueJob(models.Model):
...
+    def init(self):
+        self._cr.execute(
+            'SELECT indexname FROM pg_indexes WHERE indexname = %s ',
+            ('queue_job_identity_key_state_partial_index',)
+        )
+        if not self._cr.fetchone():
+            self._cr.execute(
+                "CREATE INDEX queue_job_identity_key_state_partial_index "
+                "ON queue_job (identity_key) WHERE state in ('pending', "
+                "'enqueued');"
```
guewen: I only realize now that it would probably be way more efficient to have `WHERE state in ('pending', 'enqueued') AND identity_key IS NOT NULL` in the index, so the index will not keep updating the index for jobs without identity_key (don't slowing down jobs that don't use this feature is a target).

guewen:
I pushed 2 new commits:

* The first (fixup) adds the `identity_key IS NOT NULL` clause in the partial index
* The seconds extends the identity key feature with ability to use identity functions (taking the job as parameter). There is a provided identity function (`identity_exact`) that should cover most of the cases: if the method call and arguments are the same for the same record(s), skip it. The possibility to provide a hash string still exists but is probably an edge case, I guess that nearly always you want the hash key to be dependent of the job method, record and arguments, and it makes no sense to compute the hash beforehand with these properties rather than computing the hash from the job properties.

guewen (approved):

yvaucher (approved):
Reapproving with last changes

PR #54, opened 2018-02-08
[11.0] Use a requests session to limit opened sessions

guewen:
See description on https://github.com/OCA/connector/pull/270

Closes #44

PR #69, opened 2018-05-22
Fix auto registration of job methods and channels

### LINKED ISSUE #64: [10.0] queue_job: default_channel argument seems to not work
I have this job with `default_channel` argument:

https://github.com/OCA/l10n-spain/blob/3baa28b44211fc34516ebe87b9068204231bff47/l10n_es_aeat_sii/models/account_invoice.py#L1389

and this is how it's called:

https://github.com/OCA/l10n-spain/blob/3baa28b44211fc34516ebe87b9068204231bff47/l10n_es_aeat_sii/models/account_invoice.py#L964

But the channel of the jobs is `root`, not the one I'm putting. Do you know why?

guewen:
The automatic registration of job methods and their defaut channel
has been a bit chaotic. The initial version for the new Odoo API could
crashing as soon as a method was decorated by @property. There is
such a property in the code code, the method '_cache'.

The problem of the crash was that, the introspection basically uses a
'getattr' on every attribute of the instance. The '_cache' method
could then be called on an empty recordset, which is not supported by
'_cache'.

A first correction (49d8f37) was to naively skip the '_cache' method
from the introspection.

In any case, it is wrong to access the property of an instance only to
instrospect its members.  That's why the correction 4ebb2452 changed the
inspection from the instance to the class.

Properties are no longer accessed, however this correction was not
correct for Python 3. When members of the class are introspected, they
are neither bound neither unbound methods. they are mere functions.

The change here is to lookup for functions. _register_job must now takes
the model as input argument, because there is no way to get the name of
the model from the function.

Closes #64
Follows #50

### DIFF
--- a/queue_job/models/base.py
+++ b/queue_job/models/base.py
@@ -18,11 +18,11 @@ class Base(models.AbstractModel):
         super(Base, self)._register_hook()
         job_methods = [
             method for __, method
-            in inspect.getmembers(self.__class__, predicate=inspect.ismethod)
+            in inspect.getmembers(self.__class__, predicate=inspect.isfunction)
             if getattr(method, 'delayable', None)
         ]
         for job_method in job_methods:
-            self.env['queue.job.function']._register_job(job_method)
+            self.env['queue.job.function']._register_job(self, job_method)
 
     @api.multi
     def with_delay(self, priority=None, eta=None,
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -18,8 +18,8 @@ from ..fields import JobSerialized
 _logger = logging.getLogger(__name__)
 
 
-def channel_func_name(method):
-    return '<%s>.%s' % (method.__self__._name, method.__name__)
+def channel_func_name(model, method):
+    return '<%s>.%s' % (model._name, method.__name__)
 
 
 class QueueJob(models.Model):
@@ -104,7 +104,7 @@ class QueueJob(models.Model):
         for record in self:
             model = self.env[record.model_name]
             method = getattr(model, record.method_name)
-            channel_method_name = channel_func_name(method)
+            channel_method_name = channel_func_name(model, method)
             func_model = self.env['queue.job.function']
             function = func_model.search([('name', '=', channel_method_name)])
             record.channel_method_name = channel_method_name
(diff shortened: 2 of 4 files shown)


guewen:
Note: the behavior is different between python2 and python3

PY3:
```
Python 3.5.2 (default, Nov 23 2017, 16:37:01)
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import inspect
>>> class A():
...     def test(self):
...         pass
...
>>> A.test
<function A.test at 0x7f6e4b6a29d8>
>>> A().test
<bound method A.test of <__main__.A object at 0x7f6e4bffd630>>
>>> inspect.getmembers(A, predicate=inspect.ismethod)
[]
>>> inspect.getmembers(A(), predicate=inspect.ismethod)
[('test', <bound method A.test of <__main__.A object at 0x7f988c7fc550>>)]
>>> inspect.getmembers(A, predicate=inspect.isfunction)
[('test', <function A.test at 0x7f988bea19d8>)]
>>> inspect.getmembers(A(), predicate=inspect.isfunction)
[]
```
PY2
```
Python 2.7.12 (default, Dec  4 2017, 14:50:18)
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import inspect
>>> class A(object):
...     def test(self):
...         pass
...
>>> A.test
<unbound method A.test>
>>> A().test
<bound method A.test of <__main__.A object at 0x7f2883429750>>
>>> inspect.getmembers(A, predicate=inspect.ismethod)
[('test', <unbound method A.test>)]
>>> inspect.getmembers(A(), predicate=inspect.ismethod)
[('test', <bound method A.test of <__main__.A object at 0x7f22066f9250>>)]
>>> inspect.getmembers(A, predicate=inspect.isfunction)
[]
>>> inspect.getmembers(A(), predicate=inspect.isfunction)
[]
```

So the fix for 10.0 will be different.

simahawk (approved):
LGTM

simahawk:
FTR runbot is only yellow: usual RST parsing warnings.
BTW I wonder if we could get rid of those: seems missing pygments pkg is the cause...

lmignon (approved):
@guewen Do you plan to change the version number  later on the master branch?

guewen:
@lmignon Yes, I'd like to merge this one and #72 then bump the version of 11.0. (#54 could be nice as well)

p-tombez (approved):
LGTM

lmignon:
@guewen everything is merged :smirk:

guewen:
@lmignon thanks! Bumped the addon with a changelog :)

lmignon:
Thank you @guewen !

PR #71, opened 2018-05-25
README as fragments

sbidoul:
Since the README had a minor glitch that blocked the pypi upload I take the opportunity to convert it to fragments.

More important (to be reviewed):
* guewen as [maintainer](http://odoo-community.org/page/maintainer-role)
* [development_status](http://odoo-community.org/page/development-status) = Mature

PR #72, opened 2018-05-25
Correct or disable pylint warnings

guewen:
The remaining warning is fixed in #71 and a warning is fixed in the checker at https://github.com/OCA/pylint-odoo/pull/201


simahawk (approved):
LGTM

sbidoul (approved):
That's indeed a prerequisite for moving to Mature :)

queue_job/tests/test_json_field.py
```diff
@@ -1,11 +1,13 @@
...
 # license agpl-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 
 from datetime import datetime, date
 import json
 
 from odoo.tests import common
+
+# pylint: disable=odoo-addons-relative-import
+# we are testing, we want to test as we were an external consumer of the API
 from odoo.addons.queue_job.fields import JobEncoder, JobDecoder
```
guewen, line 11: That's maybe not the place to discuss that, but am I the only with the opinion that tests should run as if they were from the outside, so should use absolute imports rather than relative?

pedrobaeza: @guewen you should put instead this import:
```python
from ..fields import JobEncoder, JobDecoder
```
That's why pylint complains

sbidoul: @guewen has a point

pedrobaeza: Is there any actual case where the relative import would fail?

lmignon (approved):

queue_job/tests/test_json_field.py
lmignon: I find the point of view of @guewen really relevant.

pedrobaeza: Yeah, but with no case, I prefer to be homogeneous across all code, being test or not.

guewen: > you should put instead this import:
> 
> from ..fields import JobEncoder, JobDecoder
> 
> That's why pylint complains

I know, my point is that in tests, I don't want to do that (I do in in addons code for sure).

@pedrobaeza It's not the question of failure, both should work in any case.
It's more a question of what a test is and what the boundaries of the addons are.

If we say that an addon is a library, the tests are definitely not part of this library, they just happen to be in the same folder but they don't belong to the Python package.

pedrobaeza: Well, I don't find all that conceptual separation needed. As said, I prefer homogeneity, but as always, maintainers decide and this can be raised for a general discussion for removing that check on tests.

guewen:
Rebased from 11.0

PR #76, opened 2018-05-30
[draft] 11.0 UI notifications for jobs

guewen:
This is a draft of the API for job notifications. (https://twitter.com/guewenb/status/1000818392591462400)

There is code duplicated with [web_notify](https://github.com/OCA/web/tree/11.0/web_notify) which can be traded against a new dependency. 

https://github.com/OCA/web/tree/11.0/web_notifyhttps://github.com/OCA/web/tree/11.0/web_notify

The following events are triggered:

* A new job is enqueued
* Job started
* Job finished
* Job failed

You can also send a simple notification with: `notify` or `notify_warn`:

```python
            from odoo.addons.queue_job.job import job, notify

            @api.multi
            @job(on_done=notify('Partner exported'))
            def export_partner(self):
                # ...
```

For more advanced messages, you can also create methods on the `queue.job` model and refer to them with a string. This is also better for inheritance. From these hook methods, you can call `QueueJob.notify()` or  `QueueJob.notify_warn()`.

    Example usage:

```python

        class QueueJob(models.Model):
            _inherit = 'queue.job'

            @api.multi
            def on_export_partner_done(self):
                self.ensure_one()
                model = self.model_name
                partner = self.env[model].browse(self.record_ids)
                message = _('Partner %s exported') % partner.name
                self.notify(message=message)

            _inherit = 'res.partner'

            @api.multi
            @job(on_done='on_export_partner_done')
            def export_partner(self):
                # ...
```

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -255,7 +255,7 @@ class Job(object):
             new_job.kwargs,
             new_job.uuid
         )
-        self.trigger_state_change(CREATED)
+        new_job.trigger_event_change(CREATED)
         return new_job
 
     @staticmethod
@@ -435,14 +435,16 @@ class Job(object):
             sudo_job_model = self.env[self.job_model_name].sudo()
             self._db_record = sudo_job_model.create(vals)
 
-    def trigger_state_change(self, state):
+    def trigger_event_change(self, event):
         events = {
             CREATED: self.func.on_create,
             STARTED: self.func.on_start,
             DONE: self.func.on_done,
             FAILED: self.func.on_failure,
         }
-        action = events[state]
+        action = events[event]
+        if not action:
+            return
         if isinstance(action, str):
             try:
                 getattr(self.db_record(), action)()
@@ -450,8 +452,13 @@ class Job(object):
                 _logger.error(
                     '%s could not be called when the job became "%s"'
                     ' because this method does not exist on queue.job',
-                    action, state
+                    action, event
                 )
+        else:
+            if isinstance(action, NotifyWarnMessage):
+                self.db_record().notify_warn(**action)
+            else:
+                self.db_record().notify(**action)
 
     def db_record(self):
         if not self._db_record:
@@ -514,7 +521,7 @@ class Job(object):
     def set_started(self):
         self.state = STARTED
         self.date_started = datetime.now()
-        self.trigger_state_change(STARTED)
+        self.trigger_event_change(STARTED)
 
     def set_done(self, result=None):
         self.state = DONE
@@ -522,13 +529,13 @@ class Job(object):
         self.date_done = datetime.now()
         if result is not None:
             self.result = result
-        self.trigger_state_change(DONE)
+        self.trigger_event_change(DONE)
 
     def set_failed(self, exc_info=None):
         self.state = FAILED
         if exc_info is not None:
             self.exc_info = exc_info
-        self.trigger_state_change(FAILED)
+        self.trigger_event_change(FAILED)
 
     def __repr__(self):
         return '<Job %s, priority:%d>' % (self.uuid, self.priority)
@@ -816,3 +823,40 @@ def related_action(action=None, **kwargs):
         func.kwargs = kwargs
         return func
     return decorate
+
+
+class NotifyMessage(dict):
+    """Used as a normal dict
+
+    Used to differentiate normal notifications and warn notifications
+    """
+
+
+class NotifyWarnMessage(dict):
+    """Used as a normal dict
+
+    Used to differentiate normal notifications and warn notifications
+    """
+
+
+def _prepare_notify_content(message, title=None, sticky=None, **kwargs):
+    message = {'message': message}
+    if title is not None:
+        message['title'] = title
+    if sticky is not None:
+        message['sticky'] = sticky
+    if kwargs:
+        message.update(**kwargs)
+    return message
+
+
+def notify(message, title=None, sticky=None, **kwargs):
+    return NotifyMessage(**_prepare_notify_content(
+        message, title=title, sticky=sticky, **kwargs
+    ))
+
+
+def notify_warn(message, title=None, sticky=None, **kwargs):
+    return NotifyWarnMessage(**_prepare_notify_content(
+        message, title=title, sticky=sticky, **kwargs
+    ))
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -552,7 +552,8 @@ def _is_model_method(func):
             isinstance(func.__self__.__class__, odoo.models.MetaModel))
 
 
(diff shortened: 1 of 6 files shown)


queue_job/job.py
```diff
@@ -620,14 +667,67 @@ def retryable_example():
...
+
+            from odoo.addons.queue_job.job import job, notify
+
+            @api.multi
+            @job(on_done=notify(message='Partner exported'))
+            def export_partner(self):
+                # ...
+
+    For more advanced messages, you can also create methods on the
+    ``queue.job`` model and refer to them with a string. From these
```
guewen, line 693: note to myself: better inheritance than the first way

guewen:
Looking for feedbacks on the notify API

sbidoul:
@guewen I had not had time to look in  details yet. 

My first impression is that on_done notification is as easily (or more easily) implemented with an explicit call to notify (using web_notify) in the job function itself. The callback looks a bit weird to me where it can be implemented inline in the job function.

Same for job start. Same for on enqueued. 

on_error could benefit from some support of the queue framework. 

(it's just a first impression, certainly not a definitive opinion)

guewen:
Thanks for your feedback! I want to find the right design and there is a lot of place for discussion :) I work on this on my spare time, no time constraint, no stress.

When you say inline, is it what you mean?

```python
class SaleOrder(models.Model):
    _inherit = 'sale.order'

    @job
    @api.multi
    def render_report(self):
        # 1. call to notify (job started)
        self._render()
        # 2. call to notify (job done)
```

This is an option I considered, and it would look straightforward, but at this point we miss the "author" of the job. `self` is the current sale order, and we don't know anymore what the job was, hence who triggered the job. Well we could retrieve the job with some value in the context but it wouldn't look cleaner IMO.

That's how I came to the conclusion (not definitive though) that the notify should be handled outside of the execution of the job function, from the executor of the job. 

I evaluated the idea that the `on_done`, `on_started`, ... triggers could serve other purposes by the way even if that could be YAGNI.

The `notify` method in `queue.job` is meant to call directly the method provided by `web_notify`, it should become:

```python
@api.multi
def notify(self, message, title=None, sticky=False):
    self.user_id.notify_info(message, title=title, sticky=sticky)
```

sbidoul:
Still, having to inherit queue_job to implement the notification call feels a bit awkward, compared to the inline call. In the use cases I met, we did not need access to the job itself when notifying. And I kind of remember a PR somewhere that added the job UUID in the context so it should be accessible in cases we need it.

lmignon:
@guewen I've the same feeling than @sbidoul. The notification can be easily done into the job function  if needed. On reading the comments, the missing parts is a method to get the running job on base model. As mentioned by @sbidoul everything is in place to easily implement such method. https://github.com/OCA/queue/blob/11.0/queue_job/job.py#L437

```python
class BaseModel(...):
    _inherit = 'base'

    def _get_running_job(self):
        job_uuid = self.env.context.get('job_uuid')
        if not job_uuid:
            return None
        return Job.load(self.env, job_uuid)
```

I don't know if this method should return a 'Job'  or  a 'queue.job' instance.

guewen:
Then maybe we should find an easier way to get the job (record) from the inline job function by adding a generic function in the `base` model:

```python
class BaseModel(models.Model):
    _inherit = 'base'

    def outer_job(self):
        # we could pass the id of the job in the context btw
        return self.env['queue.job'].search([('uuid', '=', self.env.context.get('job_uuid'))])


[...]

    @job
    @api.multi
    def render_report(self):
        self.outer_job().notify('hello')
        # work
        self.outer_job().notify('goodbye')
```

This is simple, flexible and would make other things easier (like if we want to see the number of retries of the job: `self.outer_job().retries`... et voilà). I really like this possibility and maybe we should do it in any way.

Now I'm thinking about an issue with this approach: as the notification is sent after the commit of the transaction, it really handles only when a job is done. In my code above, `hello` and `goodbye` would be received at the same time by the user. The `on start` notification is probably much less interesting, but it can still be good to have.

Regarding the `on create` notification, sure, it can be handled "manually" by the code delaying the job:

```python
def confirm(self):
    self.env.user.notify('report will start in background')
    self.with_delay().render_report()
```
But then you have to duplicate that everywhere you are calling the job.

guewen:
@lmignon our messages crossed with the same proposition, good sign :smile:

sbidoul:
> notification is sent after the commit of the transaction

Ah I had forgotten that. But then web_notify could gain a nice option to notify in a separated short lived transaction.

lmignon:
@guewen  :smile: 
@sbidoul I was just writing the same proposal... as stated by @guewen : good sign

sbidoul:
Also, is it correct that jobs run in the context of the user that created them by default? In which case it's the main reason to access the job when notifying that disappears.

guewen:
@sbidoul Oh yes good point, `self.env.user` is the same thing than `self.outer_job().user_id` so `self.env.user.notify(...)` is enough.

guewen:
Any proposition for failures and create? 

Other options to consider:
* have boolean `notify_create` and `notify_failure` in the `job` decorator, message being a generic one such as "{job description} failed", "{job description} created"
* add parameters in `with_delay()`, which means we would have to store the messages in the job (in the language of the user). We could store them in a json field. It would not allow dynamic messages like "Job X failed after an execution time of x.y seconds after the 3 retry."

lmignon:
@guewen 

> *  have boolean notify_create and notify_failure in the job decorator, message being a generic one such as " failed", " created"

Would this `notify_failure` parameter be redundant with mail_thread? We already get the ability to be notified by mail on job failure. 

> * add parameters in with_delay(), which means we would have to store the messages in the job (in the language of the user). We could store them in a json field. It would not allow dynamic messages like "Job X failed after an execution time of x.y seconds after the 3 retry."

It seems more natural to me to request the notification of the job creation when delaying the method. In this way we could notify the creation is some cases but not in others... But to avoid to store a new info in the ```job``` instance  we can also achieve the same result with a ```notify_create``` parameter in the ```job``` decorator and a new one into the ```delay``` method to mute the notification in some specific cases. Nevertheless if we add a new parameter in the ```job``` decorator for a message, this parameter will  not be seen as translatable by the translation system. (If I'm not wrong you can't use ```_('my message')``` into a decorator).  That means that we have to manually add this message into the translation files.

In short, I don't really have a clear opinion between one or the other approach.

yelizariev:
Hello @guewen, thank you for your efforts! Job events is something really needed. 

Personally, I'm interested in API part only and cannot review UI part as I don't have usage scenarios... 

As for API, I wonder, is it possible to add ``on_done`` callback in ``with_delay`` call? It would be useful. Though the same can be implemented via several methods decorated with ``@job(on_done=...)``

Another question. Is it easy to split this PR in two, where one adds new API only and another depends on it and adds UI?

PR #83, opened 2018-07-11
[FIX] queue_job: Extract `method_name` from `func_name`

pedrobaeza:
Extract from old `func_name` field the name of the method to be put on `method_name`.

If not, pending jobs won't be properly executed, as the function name for being run is missing.

@Tecnativa


pedrobaeza:
@guewen @lmignon can you check this final one for completing the migration steps?

guewen (approved):

pedrobaeza:
Thanks!

PR #88, opened 2018-07-30
[IMP] add job helper for simplifying testing

sebastienbeau:
Add a class and a mixing to simplify the test with job
Here is an case of use : https://github.com/akretion/connector-search-engine/blob/36c59f1f773f8ad4578af940c4a3ece78383c4ff/connector_algolia/tests/test_connector_algolia.py#L14

### DIFF
--- a/queue_job/tests/common.py
+++ b/queue_job/tests/common.py
@@ -1,1 +1,37 @@
 # -*- coding: utf-8 -*-
+
+import unittest
+from odoo.addons.queue_job.job import Job
+
+
+class JobCounter(object):
+
+    def __init__(self, env):
+        super(JobCounter, self).__init__()
+        self.env = env
+        self.existing = self.search_all()
+
+    def count_all(self):
+        return len(self.search_all())
+
+    def count_created(self):
+        return len(self.search_created())
+
+    def count_existing(self):
+        return len(self.existing)
+
+    def search_created(self):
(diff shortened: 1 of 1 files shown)


queue_job/tests/common.py
```diff
@@ -1 +1,42 @@
...
+    @property
+    def count_existing(self):
+        return len(self.existing)
+
+    @property
+    def created(self):
+        return self.all - self.existing
+
+    @property
+    def all(self):
```
guewen: `all` is a  builtin ;)

queue_job/tests/common.py
```diff
@@ -1 +1,37 @@
...
+
+class JobCounter(object):
+
+    def __init__(self, env):
+        super(JobCounter, self).__init__()
+        self.env = env
+        self.existing = self.search_all()
+
+    def count_all(self):
+        return len(self.all())
```
guewen: s/self.all()/self.search_all()/

guewen (approved):
Looks good!

simahawk (approved):

guewen:
I'm going to stay offline for some time, if we have enough approvals could you take care of merging @lmignon? Thanks :)

lmignon (approved):

sebastienbeau:
Merci !

PR #92, opened 2018-08-27
[10.0][FIX]queue_job: search jobs with same identity key must be done with sudo()

Cedric-Pigeon:
When doing a search, it must be done in sudo() otherwise users which are not in the group `Job Queue Manager` will get a security error.
It can occurs when process launch a job throug a `.delay` instruction.

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -301,7 +301,7 @@ class Job(object):
 
     def job_record_with_same_identity_key(self):
         """Check if a job to be executed with the same key exists."""
-        existing = self.env['queue.job'].search(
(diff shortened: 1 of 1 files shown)


Cedric-Pigeon:
@lmignon @sebastienbeau This is required for shopinvader. If a user triggers a stock export, he will get an error if he is not in the `Job Queue Manager` group.

sebastienbeau:
:+1: you're right this should be done with sudo

sebastienbeau (approved):

rvalyi (approved):

bguillot (approved):

PR #93, opened 2018-08-28
[MIG][11.0] mail_queue_job

acsonefho:
Old module name: mail_connector_queue (from connector repo)

- [x] Move module
- [x] XML migration
- [x] Python migration
- [x] Add tests
- [x] Flake/Pylint
- [x] Setup


oca_dependencies.txt
```diff
@@ -13,3 +13,4 @@
 #
 # To provide both the URL and a branch, use:
 # sale-workflow https://github.com/OCA/sale-workflow branchname
+connector https://github.com/OCA/connector 11.0
```
sbidoul: why?

acsonefho: component and component_event modules comes from connector repo

sbidoul: ah yes, that makes a nice circular dependency loop between connector and queue. Not sure if it is an issue in practice.

Maybe it's lighter to overwrite create/write manually for this simple use case? What do you think @lmignon @guewen ?

lmignon: Indeed, It could be lighter to only overwrite 'create/write' methods into the mail.mail to avoid this dependency. On the other hand the code looks clean and is easy to read when using the component.event addon to add a hook on the create/write method. The more I use components, the more I'm convinced that we should remove the dependency on connector to make it more generic. But it's an other debate.  Since the dependency is only required for this unique use into the addon I'm in favour to remove-it.

mail_queue_job/README.rst
```diff
@@ -0,0 +1,92 @@
+====================
+Mail Connector Queue
```
guewen, line 2: `Mail Queue Job`?

guewen (approved):
Great! Thanks :)

mail_queue_job/README.rst
acsonefho: This file should be re-generated automatically after the merge?
And the name is automatically set by the name of the directory I suppose.

guewen: I was not sure from where the name comes from, it's from the manifest file: https://github.com/OCA/maintainer-tools/blob/80f894b594810200386f176e19cd79bf4d7be8c5/tools/gen_addon_readme.template#L11

So indeed this is going to be correct after the automatic update as the name in the manifest is good.

jabibi (approved):
Code looks good to me. Thanks :+1:

mail_queue_job/models/mail_mail.py
```diff
@@ -0,0 +1,90 @@
...
+    @job(default_channel='root.mail')
+    def _send_mail_jobified(self):
+        """
+        Send the email if
+        - the state is 'outgoing'
+        - the mail is not processing by another job (locked)
+        - the mail exists.
+        :return: str
+        """
+        # Make sure two jobs don't send the same email.
```
lmignon: @acsonefho It would be nice to forward port https://github.com/OCA/queue/pull/66 in 11.0 To be able to use the new concept of Idendity key on queue job. This forward port could solve the problem of multiple jobs for the same email in a more 'elegant' way.

acsonefho: @lmignon It's done, thank you for your advice :)

mail_queue_job/models/mail_mail.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @api.multi
+    def _send_email_delay(self, operation='write'):
+        """
+        Execute the _send_mail_jobified function with the priority defined
+        into the mail.mail (self) and a description depending on the
+        operation.
+        This function split given records to create one job queue per mail.
+        The identity_key (on the with_delay function) is very helpful to ensure
+        that another queue job will send the same email.
```
guewen, line 42: I'm not sure what the intention here but beware of this: it won't prevent a concurrent transaction to create a job for the same email at the same time if the concurrent transaction happens within the same timeframe. It will however prevent the creation of a job at a different moment (only before the job is done) or within the same transaction. Is it enough for this use case?

acsonefho: The intention is to avoid sending the same `mail.mail` more than once (if a job is already planned for a `mail.mail` recordset, it shouldn't have another job who'll re-send the same `mail.mail`).
But when the job is finished, a user can try to re-sent the same `mail.mail` so it could create a new job for that.
About your first case (same timeframe and concurrent transaction) it should never happens because jobs are created into `create`/`write` so it's the same transaction.
Thank you for your review @guewen

PR #97, opened 2018-08-31
[WIP] 11.0 Port jobrunner as a worker

guewen:
From @StefanRijnhart's https://github.com/OCA/connector/pull/288

This is an early untested PR.

### DIFF
--- a/queue_job/jobrunner/__init__.py
+++ b/queue_job/jobrunner/__init__.py
@@ -3,7 +3,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 
 import logging
-import os
 from threading import Thread
 import time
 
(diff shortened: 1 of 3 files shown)


simahawk (approved):
no test, LG

simahawk (changes requested):
but tests would be nice to have, at least on initialization

simahawk:
early review: missed [WIP] in title :stuck_out_tongue:

zakiuu:
Hi @guewen, just checking on any updates about this PR?

guewen:
Hi @zakiuu, I could not spend time on this unfortunately :-(

zakiuu:
you can list the remaining things you wanted to add , and I'll take care of it with you permission of corse.

guewen:
IIRC I started it by picking the changes from 07cca55 but never actually tested it.
So, that would mostly be about testing with enough jobs and leave it running long enough. I wanted to see how it reacts when a memory limit is reached on the jobrunner's worker too.

If you want to take over from there, I would highly appreciate of course :)

PR #99, opened 2018-09-25
[11.0] Forward port. changes about identity_key (on queue_job)

acsonefho:
- [x] Forward port. of #66 (and fix conflicts);
- [x] Minor changes for Flake/Pylint;
- [x] Tests pass (using test_queue_job);

### DIFF
--- a/queue_job/fields.py
+++ b/queue_job/fields.py
@@ -22,8 +22,7 @@ class JobSerialized(fields.Field):
         value = value or {}
         if isinstance(value, dict):
             return value
-        else:
-            return json.loads(value, cls=JobDecoder, env=record.env)
+        return json.loads(value, cls=JobDecoder, env=record.env)
 
 
 class JobEncoder(json.JSONEncoder):
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -3,6 +3,7 @@
 
 import inspect
 import functools
+import hashlib
 import logging
 import uuid
 import sys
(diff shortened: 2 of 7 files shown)


lmignon:
@acsonefho It would be nice to forward port all the changes in 10.0 not yet into 11.0 For #66 you must also include 3955f772cef06e8cf75a5d65e4608487b97c565c

acsonefho:
> @acsonefho It would be nice to forward port all the changes in 10.0 not yet into 11.0 For #66 you must also include [3955f77](https://github.com/OCA/queue/commit/3955f772cef06e8cf75a5d65e4608487b97c565c)

@lmignon It's maybe better to include others changes into new MR (1 per "feature"/improvement: easier to understand what is forwarded, easier to review and in case of big issue, easier to revert only 1 feature).
What do you think?
(I just include the commit that I forget into this MR)

queue_job/__manifest__.py
```diff
@@ -1,24 +1,27 @@
...
+    'data': [
+        'security/security.xml',
+        'security/ir.model.access.csv',
+        'views/queue_job_views.xml',
+        'data/queue_data.xml',
+    ],
+    'installable': True,
+    'development_status': 'Mature',
+    'maintainers': ['guewen'],
+}
```
guewen: did you really need to change the format here and in some other places? AFAIK both styles are accepted and changing them can create new conflicts when porting changes between versions

acsonefho: @guewen Thank you for your advice, it should be ok now

guewen (approved):
Thanks!

lmignon (approved):

PR #101, opened 2018-10-02
Migrate queue_job to 12.0

guewen:
First commit is the general changes for the migration.
There is one independent commit for the setuptools files.
Other commits are some particular migration changes which are worth to have in
a standalone commits or port of features from 11.0 which were merged after the branching of 12.0.

PR #102, opened 2018-10-02
Migrate base_import_async to 12.0

guewen:
* [x] needs #101

The first commits are shared with #101, relevant commits are the 2 last ones.

### DIFF
--- a/base_import_async/__manifest__.py
+++ b/base_import_async/__manifest__.py
@@ -23,4 +23,5 @@
         'static/src/xml/import.xml',
     ],
     'installable': True,
+    'development_status': 'Stable',
 }
--- a/base_import_async/__manifest__.py
+++ b/base_import_async/__manifest__.py
@@ -7,7 +7,7 @@
 {
     'name': 'Asynchronous Import',
     'summary': 'Import CSV files in the background',
-    'version': '11.0.1.0.0',
+    'version': '12.0.1.0.0',
     'author': 'Akretion, ACSONE SA/NV, Odoo Community Association (OCA)',
     'license': 'AGPL-3',
     'website': 'https://github.com/OCA/queue',
@@ -22,5 +22,5 @@
     'qweb': [
         'static/src/xml/import.xml',
     ],
-    'installable': False,
+    'installable': True,
 }
--- a/base_import_async/models/base_import_import.py
+++ b/base_import_async/models/base_import_import.py
@@ -89,9 +89,9 @@ class BaseImportImport(models.TransientModel):
         # write csv
         f = StringIO()
         writer = csv.writer(f,
-                            delimiter=str(options.get(OPT_SEPARATOR)),
+                            delimiter=str(options.get(OPT_SEPARATOR)) or ',',
                             quotechar=str(options.get(OPT_QUOTING)))
-        encoding = options.get(OPT_ENCODING, 'utf-8')
+        encoding = options.get(OPT_ENCODING) or 'utf-8'
         writer.writerow(fields)
         for row in data:
             writer.writerow(row)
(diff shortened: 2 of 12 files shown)


base_import_async/__manifest__.py
```diff
@@ -22,5 +22,5 @@
     'qweb': [
         'static/src/xml/import.xml',
     ],
-    'installable': False,
+    'installable': True,
```
lmignon, line 25: `development_status` ????

lmignon (approved):
Thank you @guewen

base_import_async/__manifest__.py
guewen: added

lmignon (commented):
Strange, the import functionality seems to not work at all on runbot....

guewen:
> Strange, the import functionality seems to not work at all on runbot....

Because of the CORS issue?

> An unknown CORS error occured. The error probably originates from a JavaScript file served from a > different origin. (Opening your browser console might give you a hint on the error.)

guewen:
On my tests on runbot, the jobs are generated but there is this CORS error that pops when the notification appears. And in the browser console I get:

```
uncaught exception: an exception occurred in a caller-provided callback function
XML Parsing Error: syntax error
Location: http://3341401-102-8f5137.runbot3.odoo-community.org/longpolling/poll
Line Number 1, Column 1:
```

There is a problem with the longpolling:
selection_395

The expected response for longpolling would have been something like (copied from odoo's runbot):
```
{"jsonrpc": "2.0", "id": 403105148, "result": [{"id": 406, "channel": ["381328-12-0-360212-all", "res.partner", 3], "message": {"type": "activity_updated", "activity_created": true}}]}
```

lmignon:
@guewen Yes the CORS issue is one problem but in my case I was not able to create the job...

lmignon:
@sbidoul @sbejaoui Do you use this addon on 12.0?

sbidoul:
@lmignon I've not tried it yet.

GSLabIt:
@guewen did u try to import products with url image? it seems to be failing, while importing "normally" it works properly.

Schermata 2019-09-18 alle 19 00 43

@lmignon 
> @guewen Yes the CORS issue is one problem but in my case I was not able to create the job...

it seems that, at least importing excel, `options` are not passed correctly; both `encoding` and `separator` are `''`, so it fails opening writer. 

Beside that i'm getting CORS problem as well ..

**edit:**: I can confirm issue problem _1-character string_ is raised by `encoding` and `separator` when importing excel file format. It will not forward `options` correctly; I fixed it having `encoding` failing to `'utf-8'` and `delimiter` to `','`.

GSLabIt:
@guewen any news on CORS issue?

guewen:
@GSLabIt No, I didn't have the opportunity to work on this.

guewen:
I found the origin of the error (which is unrelated to CORS, only appearing as such), fixed in my last commit. I think we are ready to merge!

GSLabIt:
@guewen, great news. Did u try importing image from url from excel?

have a look at https://github.com/OCA/queue/pull/102#issuecomment-532776170

guewen:
> Did u try importing image from url from excel?

Nope, forgot about this one :) (never tried this actually)

guewen:
Actually, by looking at the implementation, I think that only CSV files are supported (which sort of make sense since files are split per line). I think we should hide the options and add a comment about this in the readme. (and if someone wants to implement excel file supports, it can be added later).

GSLabIt:
@guewen what about fixing it by having `encoding` failing to `'utf-8'` and `delimiter` to `','`?

guewen:
You are right, not sure how but excel files are read correctly. I pushed a new commit with this correction.

GSLabIt:
> You are right, not sure how but excel files are read correctly. I pushed a new commit with this correction.

That's because you'll have data parsed by: 
https://github.com/OCA/queue/blob/b6de8170c85aeb69040ae3eecc097c29052b9e85/base_import_async/models/base_import_import.py#L47

So first all file is parsed, than splitted. If  have very big file, parsing could end in timeout.

guewen:
Thanks @GSLabIt  :)

GSLabIt (approved):
LGTM

PR #107, opened 2018-10-10
11.0 Extract method posting message on job failure

guewen:
Allowing to alter the behavior by inheritance.


guewen:
approved on #108

PR #108, opened 2018-10-10
12.0 Extract method posting message on job failure

guewen:
Allowing to alter the behavior by inheritance.


lmignon (approved):
Modularity improvements are always good things! Thank you @guewen  
(code review)

PR #115, opened 2018-12-14
[10.0][IMP] queue_job: Don't prefetch fields when vacuum + bumpversion

rousseldenis:
@guewen @lmignon 

Did this as sometimes, the amount of jobs is huge and there is no need fetching fields.

### DIFF
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -5,7 +5,8 @@
 import logging
 from datetime import datetime, timedelta
 
-from odoo import models, fields, api, exceptions, _
+from odoo import models, fields, api, exceptions, registry, _
+from psycopg2 import OperationalError
 
 from ..job import STATES, DONE, PENDING, Job
 from ..fields import JobSerialized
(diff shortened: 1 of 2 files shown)


rousseldenis:
@guewen I bumped version as it was not done for long. But in separate commit if you disagree.

queue_job/models/queue_job.py
```diff
@@ -226,7 +226,7 @@ def autovacuum(self):
         Called from a cron.
         """
         deadline = datetime.now() - timedelta(days=self._removal_interval)
-        jobs = self.search(
+        jobs = self.with_context(prefetch_fields=False).search(
```
guewen: I was sure that the prefetch only happen the first time you access a field (other than id)?

rousseldenis: I will try to objective that with a test script (not in the module)

rousseldenis:
@guewen I think you're right.
The problem is the timeout when your backlog is huge. I was inspired by odoo orderpoint runner (https://github.com/odoo/odoo/blob/10.0/addons/stock/models/procurement.py#L308) to make sets of jobs to unlink and commit.
I'm not sure of all the parts I've done, so if you have suggestions, I will implement them !

queue_job/models/queue_job.py
```diff
@@ -220,16 +220,39 @@ def _needaction_domain_get(self):
...
+    def autovacuum(self, jobs_set=1000, use_new_cursor=False):
         """ Delete all jobs done since more than ``_removal_interval`` days.
-
+        Uses sets of jobs to ensure that no timeout occurs
         Called from a cron.
         """
+        job_obj = self.env['queue.job']
         deadline = datetime.now() - timedelta(days=self._removal_interval)
-        jobs = self.search(
+        jobs_ids = self.with_context(prefetch_fields=True).search(
```
lmignon: @rousseldenis IMO prefetch_field is useless

queue_job/models/queue_job.py
```diff
@@ -220,16 +220,39 @@ def _needaction_domain_get(self):
...
+            jobs_ids = jobs_ids[jobs_set:]
+
+            jobs.unlink()
+
+            try:
+                if use_new_cursor:
+                    cr.commit()
+            except OperationalError:
+                if use_new_cursor:
+                    cr.rollback()
```
lmignon, line 247: @rousseldenis IMO all the logic should be enclosed into a try except finally block
```python
    if use_new_cursor:
        cr = registry(self._cr.dbname).cursor()
    try:
        self = self.with_env(self.env(cr=cr))
        jobs = job_obj.browse(jobs_ids[:jobs_set])
        jobs_ids = jobs_ids[jobs_set:]
        jobs.unlink()
        if use_new_cursor:
            cr.commit()
    except OperationalError:
         if use_new_cursor:
            cr.rollback()
            raise
    finally:
        if use_new_cursor:
            cr.close()
```

lmignon (changes requested):

rousseldenis:
@lmignon @guewen

lmignon (commented):
@rousseldenis Can you modify the changlog PLZ https://github.com/OCA/queue/blob/10.0/queue_job/README.rst#changelog

rousseldenis:
@guewen As suggested by @lmignon, should I fill in changelog as it should includes all commits descriptions since the previous version?

P.S.: The modification has been tested on a real database.

sbidoul:
@rousseldenis are you sure the benefit is worth the additional code complexity?
I understood you faced the issue in a real pathological situation (which could be resolved manually by a sql delete). When the vacuum cron runs regularly nobody faces the issue, even on the largest queue job delployments AFAIK.

rousseldenis:
@sbidoul You have to handle also mail messages. So, in fact there should be multiple sql deletes.

PR #120, opened 2019-01-22
Retry "Unable to use closed cursor"

guewen:
Sometimes during the computation of computed fields, odoo fails
with an "Unable to use closed cursor" error. A job failing due
to such random error will be retried.

An analysis of the root cause should be run, this is only a band-aid.


guewen:
Do other people get similar errors in 10.0? This might not need to be merged, if no other reports the same issue, it might be due to a problem in our local codebase.

PR #121, opened 2019-01-31
[IMP] queue: Store context in order to reuse it

etobella:
With these PR, the context is stored and it can be used in the job.
Sometimes, you don't want to lose the context of the user.

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -55,7 +55,7 @@ class DelayableRecordset(object):
 
     def __init__(self, recordset, priority=None, eta=None,
                  max_retries=None, description=None, channel=None,
-                 identity_key=None):
+                 identity_key=None, job_context=None):
         self.recordset = recordset
         self.priority = priority
         self.eta = eta
@@ -63,6 +63,7 @@ class DelayableRecordset(object):
         self.description = description
         self.channel = channel
         self.identity_key = identity_key
+        self.job_context = job_context or {}
 
     def __getattr__(self, name):
         if name in self.recordset:
@@ -87,8 +88,17 @@ class DelayableRecordset(object):
                                eta=self.eta,
                                description=self.description,
                                channel=self.channel,
-                               identity_key=self.identity_key)
+                               identity_key=self.identity_key,
+                               job_context=self.get_context(recordset_method))
         return delay
+
+    def get_context(self, method):
+        original_ctx = self.job_context
+        ctx = {}
+        for key in getattr(method, 'allow_context', []):
+            if key in original_ctx:
+                ctx[key] = original_ctx[key]
+        return ctx
 
     def __str__(self):
         return "DelayableRecordset(%s%s)" % (
@@ -241,6 +251,10 @@ class Job(object):
         be added to a channel if the existing job with the same key is not yet
         started or executed.
 
+    .. attribute::job_context
+
+        Original context of the job
+
     """
     @classmethod
     def load(cls, env, job_uuid):
@@ -259,7 +273,6 @@ class Job(object):
         args = stored.args
         kwargs = stored.kwargs
         method_name = stored.method_name
-
         model = env[stored.model_name]
 
         recordset = model.browse(stored.record_ids)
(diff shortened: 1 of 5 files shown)


guewen:
IMO it is better to add an explicit argument in the method if you want to keep something.

etobella:
That is what I have usually done and I always thought that it was a limitation.
Maybe there is a reason for this behaviour and it is a wrong approach to use this PR.

queue_job/job.py
```diff
@@ -397,6 +405,7 @@ def __init__(self, func,
 
         recordset = func.__self__
         env = recordset.env
+        self.job_context = job_context
```
simahawk: `self.job_context = job_context or {}`

pedrobaeza:
I also agree that is good to add corresponding arguments, but there can be context values that are out of your control, like `force_company`, `tz` or `lang` (that doesn't correspond with the job owner lang). Is that the case, @etobella ?

guewen:
I find that the context is sometimes/often out of control, it may even contain values which cannot be serialized. If the requirement concerns predefined fields, such as `force_company`, `tz`, `lang`, then I would agree that we should keep a stripped-down copy of the context keeping only a list of safe fields.

etobella:
It is true, that reusing context might be crazy, but it is a tool that is used sometimes and it is interesting to store it somehow. A list of context could be fine. I will work on a safe list of fields.

etobella:
@guewen The list of safe context keys has been added. It might be better now.

queue_job/models/base.py
```diff
@@ -27,6 +27,20 @@ def _register_hook(self):
         for job_method in job_methods:
             self.env['queue.job.function']._register_job(self, job_method)
 
+    @api.model
+    def get_job_context_keys(self):
```
simahawk: these methods should be private IMO

simahawk (changes requested):

queue_job/models/base.py
guewen: Suggestion: setup the context keys to pass on the @job decorator

```
@job(context_allow=['foo'])
```

With `['force_company', 'lang', 'tz']` by default and new keys added to the list.
The filtering of the context keys could then be hidden in the job implementation details, instead of cluttering the `base` model (I try hard to put as less method as possible here).

Would it work for you?

guewen: Or if the safe list really needs to be global, then it can be added to `queue.job`

etobella: great idea about the job. I will make the change

sbidoul:
Will this pass some context to the job by default, that was not passed before?

etobella:
Yes, the context was not passed to the job.

sbidoul:
@etobella then I'd say it must not be passed at all unless asked explicitly by the developer, as this could change the behavior of existing deployments.

etobella:
@sbidoul That is the last change we applied today. The context keys are defined on the job decorator in order to ensure flexibility

jarroyomorales:
@etobella Travis already fixed in #172

etobella:
@simahawk your comments have been attended

simahawk (approved):

lmignon:
With this implementation, how can we add a new key into the `allow_context`list? https://github.com/OCA/queue/issues/273#issuecomment-718842408 could be the right way to provide this kind of info.

albariera (approved):

maljac:
Any chance this PR will be merged soon (and maybe ported to v10)?

guewen:
There is an issue I'm not sure how to handle, which is that in higher versions of Odoo, the context should be stored in the Serialized records (see #281, #283). This change cannot be ported as-is to Odoo 12+. My feeling is that we should address this problem in more recent versions of Odoo, and see how to transition on migrations. Otherwise we will support something in 11.0 that will be lost when you upgrade...

guewen:
FWIW in many cases (actually I never needed the context for this reason), it is IMO better to pass an explicit argument to the job method than rely on an arbitrary context

PR #124, opened 2019-02-28
[12.0] [REF] Queue Job: allow configuration from server_environment

### LINKED ISSUE #62: Making queue_job aware of server_environment
Hi,

It would be nice to make the jobrunner configuration compatible with server_environment.

Our motivation is to push the config with code and avoid risks of deployment errors when deployment is not under our direct control.

I was wondering if something simple, albeit a bit hackish, would make sense [here](https://github.com/OCA/queue/blob/5c098fc3340f7b837cbb5da18ea4d0fa9aadc37c/queue_job/jobrunner/runner.py#L131):

```
try:
    from openerp.addons.server_environment import serv_config as config
except ImportError:
    from odoo.tools import config
```

@guewen @lmignon

ThomasBinsfeld:
Fixes #62

### DIFF
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -147,7 +147,7 @@ import requests
 
 import odoo
 from odoo.tools import config
-
+from . import queue_job_config
 from .channels import ChannelManager, PENDING, ENQUEUED, NOT_DONE
 
 SELECT_TIMEOUT = 60
@@ -170,8 +170,7 @@ session = requests.Session()
 def _channels():
     return (
         os.environ.get('ODOO_QUEUE_JOB_CHANNELS') or
-        config.misc.get("queue_job", {}).get("channels") or
-        "root:1"
+        queue_job_config.get("channels") or "root:1"
     )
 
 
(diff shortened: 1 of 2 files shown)


queue_job/jobrunner/runner.py
```diff
@@ -146,7 +146,15 @@
...
-from odoo.tools import config
+try:
+    from odoo.addons.server_environment import serv_config as config
+    if config.has_section('queue_job'):
+        queue_job_config = config['queue_job']
+    else:
+        queue_job_config = {}
+except ImportError:
+    from odoo.tools import config
+    queue_job_config = config.misc.get('queue_job', {})
```
guewen: This is the same code than the one in `__init__.py`, could you import only `queue_job_config` from a common place?

guewen (changes requested):
Besides the replication of the import logic, it looks good, thanks!

lmignon:
@ThomasBinsfeld Can you take care of the comments expressed by guewen

queue_job/jobrunner/runner.py
ThomasBinsfeld: Done

ThomasBinsfeld:
ping @lmignon

guewen (approved):

lmignon (approved):

sbidoul (approved):

PR #130, opened 2019-03-08
[12.0][FIX] retry_postpone TransactionRollback

liweijie0812:
The transaction rollback causes the execution of the retry_postpone method to also be rolled back

PR #131, opened 2019-03-08
[12.0][FIX] under pressure

sbidoul:
[Under pressure, ie when starting a job takes
more than one second, the jobrunner requeues
jobs it attempted to start](https://github.com/OCA/queue/blob/c6dbf6688db64ea1487f75e0320e2f46d9c1a4a4/queue_job/jobrunner/runner.py#L248-L249). This reveals a race condition
that was identified in a TODO that exists since I first
wrote the job runner. At the time, it
did not matter because the requeueing of started job
did not exist. Now it needs to be fixed.

To avoid interactions with the ORM cache, the approach
is to lock the job record and ensure it is in the correct state
before loading it, and changing it to started state
in _try_perform_job.

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -26,13 +26,13 @@ class RunJobController(http.Controller):
         """Try to perform the job."""
         job.set_started()
         job.store()
-        http.request.env.cr.commit()
-
+        env.cr.commit()
         _logger.debug('%s started', job)
+
         job.perform()
         job.set_done()
         job.store()
-        http.request.env.cr.commit()
+        env.cr.commit()
         _logger.debug('%s done', job)
 
     @http.route('/queue_job/session', type='http', auth="none")
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -40,19 +40,6 @@ class RunJobController(http.Controller):
 
     def _try_perform_job(self, env, job):
         """Try to perform the job."""
-
-        # if the job has been manually set to DONE or PENDING,
-        # or if something tries to run a job that is not enqueued
-        # before its execution, stop
-        if job.state != ENQUEUED:
-            _logger.warning('job %s is in state %s '
-                            'instead of enqueued in /runjob',
-                            job.uuid, job.state)
-            return
-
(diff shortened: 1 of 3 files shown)


sbidoul:
The visible effect of that race condition was the same job being run more than once *in parallel*.

sbidoul:
@guewen this may be the root cause of #130, #41 and maybe #120

queue_job/controllers/main.py
```diff
@@ -26,42 +22,17 @@
...
-
-        # if the job has been manually set to DONE or PENDING,
-        # or if something tries to run a job that is not enqueued
-        # before its execution, stop
-        if job.state != ENQUEUED:
-            _logger.warning('job %s is in state %s '
-                            'instead of enqueued in /runjob',
-                            job.uuid, job.state)
-            return
-
```
sbidoul, line 52: This is where the race condition occured: two simultaneous runs of the same job both reach this point in enqueued state.

queue_job/controllers/main.py
```diff
@@ -87,10 +58,23 @@ def retry_postpone(job, message, seconds=None):
             job.store()
             env.cr.commit()
 
-        job = self._load_job(env, job_uuid)
-        if job is None:
+        # ensure the job to run is in the correct state and lock the record
+        env.cr.execute(
+            "SELECT state FROM queue_job "
+            "WHERE uuid=%s AND state=%s "
+            "FOR UPDATE",
```
guewen, line 65: Do we want to use NOWAIT and catch concurrent errors, to avoid locking the second worker?

sbidoul: @guewen I thought about that and opted for the simpler code. My thinking is the benefit is marginal (a nicer log in rare situations) for a more complex code. The lock will never be held long (ie until the job is started immediately after).

guewen: Didn't realize it was only until it starts, better to keep it simple yes :)

guewen (approved):

ThomasBinsfeld (approved):

sbidoul:
This is working fine in production.

PR #135, opened 2019-03-13
[FIX/IMP] a couple issues observed during startup, take 2 :)

sbidoul:
Hopefully this one is correct.

@ThomasBinsfeld @guewen

### DIFF
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -133,7 +133,7 @@ Caveat
        of running Odoo is obviously not for production purposes.
 """
 
-from contextlib import closing
+from contextlib import closing, contextmanager
 import datetime
 import logging
 import os
@@ -317,6 +317,7 @@ class Database(object):
             """)
             cr.execute("LISTEN queue_job")
 
+    @contextmanager
     def select_jobs(self, where, args):
         # pylint: disable=sql-injection
         # the checker thinks we are injecting values but we are not, we are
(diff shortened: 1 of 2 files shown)


queue_job/jobrunner/runner.py
```diff
@@ -326,9 +327,9 @@ def select_jobs(self, where, args):
                  "priority, EXTRACT(EPOCH FROM eta), state "
                  "FROM queue_job WHERE %s" %
                  (where, ))
-        with closing(self.conn.cursor()) as cr:
+        with closing(self.conn.cursor("select_jobs", withhold=True)) as cr:
```
guewen, line 330: Nice, I didn't know about the `withhold` parameter

queue_job/jobrunner/runner.py
```diff
@@ -386,8 +387,9 @@ def initialize_databases(self):
                 _logger.debug('queue_job is not installed for db %s', db_name)
             else:
                 self.db_by_name[db_name] = db
-                for job_data in db.select_jobs('state in %s', (NOT_DONE,)):
-                    self.channel_manager.notify(db_name, *job_data)
+                with db.select_jobs('state in %s', (NOT_DONE,)) as cr:
+                    for job_data in cr:
```
guewen, line 391:
```suggestion
                    for job_data in cr.fetchall():
```
?

sbidoul: @guewen why? My whole goal here is to fetch piecewise from the server.

ThomasBinsfeld (approved):
Functional test

queue_job/jobrunner/runner.py
guewen: You are right :)

guewen (approved):

PR #141, opened 2019-03-28
[IMP] recordset jobs ignore non-existent records

### LINKED ISSUE #140: [RFC] should jobs ignore deleted records?
We have a situation where we create jobs, but sometimes when the job starts the concerned record has been deleted. Such jobs fail.

The use case, is asynchronous execution of costly but non urgent computation. (forecasting in the contract module more precisely).

I'm wondering if queue_job should ignore deleted records, or if that should be handled at the application level, and how?

@lmignon @guewen cc/ @sbejaoui

sbidoul:
When records underlying a recordset job have been deleted at
the time the job starts, ignore them instead of failing the job.

Closes #140 

This is WIP, untested.

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -261,9 +261,11 @@ class Job(object):
         method_name = stored.method_name
 
         model = env[stored.model_name]
-
-        recordset = model.browse(stored.record_ids)
-        method = getattr(recordset, method_name)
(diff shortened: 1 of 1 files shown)


guewen:
FTR, discussion on https://github.com/OCA/queue/issues/140

PR #142, opened 2019-04-02
[11.0] Retry postpone in a new transaction

guewen:
If the job's transaction fail with an OperationalError such as a
contraint failure, the transaction is broken and we will not be able
to postpone the job. Open a new transaction to change the job's state.

Port of:

- #41 from @2zx
- #130 from @liweijie0812


simahawk (approved):

simahawk:
FTR runbot is yellow and not red

CasVissers-360ERP (approved):
Tested. Works fine.

PR #144, opened 2019-04-18
[ADD] Added autorequeue cron job and function. Disabled by default

wpichler:
with this new cron job you can auto requeue jobs in state Started after 30 Minutes

### DIFF
--- a/queue_job/data/queue_data.xml
+++ b/queue_job/data/queue_data.xml
@@ -22,6 +22,18 @@
             <field eval="'()'" name="args"/>
         </record>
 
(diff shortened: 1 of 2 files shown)


wpichler:
does not change behaviour of current implementation - cron job is disabled by default

guewen:
Hi,

Thank you for your contribution.

A similar development has been done in the 9.0 version: https://github.com/OCA/connector/pull/325
The one in 9.0 handles the jobs stuck in `enqueued` as well. It is also more configurable, through arguments in the cron. Would you mind to port the pull request I linked as you need this feature?

Thanks

PR #145, opened 2019-04-19
New module to Export (CSV and XLS) files in background

apineux:
When you use the standard Export from Odoo, you can tick the "Asynchronous export" checkbox that will generate the export file and send it to the user by email.


base_export_async/models/delay_export.py
```diff
@@ -0,0 +1,85 @@
...
+    _description = 'Allow to delay the export'
+
+    @api.model
+    def delay_export(self, data):
+        params = json.loads(data.get('data'))
+        context = params.get('context', {})
+        uid = context.get('uid', False)
+        if not uid:
+            raise Warning(_("A problem occurs during the job creation. Please contact your administrator"))
+        user = self.env['res.users'].browse([uid])
```
sbidoul: don't we have `self.env.user`?

base_export_async/README.rst
```diff
@@ -0,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=================
+Base Export Async
+=================
+
+Standard Export can be delayed in asynchronous jobs executed in the background and then send by email to the user.
```
sbidoul, line 28: have a look at the readme fragment mechanism, so the standard OCA readme.rst is auto generated.
https://github.com/OCA/maintainer-tools/tree/master/template/module/readme

README.md
```diff
@@ -20,6 +20,7 @@ addon | version | summary
 [queue_job](queue_job/) | 12.0.1.0.0 | Job Queue
 [queue_job_cron](queue_job_cron/) | 12.0.1.0.1 | Scheduled Actions as Queue Jobs
 [test_queue_job](test_queue_job/) | 12.0.1.0.0 | Queue Job Tests
+[base_export_async](base_export_async/) | 12.0.1.0.0 | Export (CSV and XLS) files in the background
```
sbidoul, line 23: note this file is auto generated

base_export_async/models/delay_export.py
apineux: The method export is the job method, so I take the uid from the context coming from the JS when the user click on the "export data". This way I'm sure that I'm taking the right user to get the email address. I didn't check if the Jobs are executed with the same user or if it is executed with the admin.

base_export_async/README.rst
apineux: I'll take a look

README.md
apineux: I guess this is linked to your previous comment about the autogenerated readme.

sbidoul: yes and no, actually it's this specific section of README.md that is generated, so it's not mandatory to update it manualy... but it does no harm.

base_export_async/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+
+{
+    'name': 'Base Export Async',
+    'summary': """
+        Asynchronous export with job queue
+        """,
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV, Odoo Community Association (OCA)',
+    'website': 'https://acsone.eu/',
```
sbejaoui:
```suggestion
    'website': 'https://github.com/OCA/queue',
```

sbejaoui (approved):
LGTM, thanks.

apineux:
IMPROVEMENT TODO: To avoid attachment exceeding the maximum size for email (usually 10Mo or 25Mo) we're going to save the attachment in ir.attachment and provide the link in the email. This attachment will stay in Odoo for a period of time specified in ir_config_parameter (default will be 7 days) and then will be deleted.

lmignon (approved):
Thank you @apineux Great new feature! LGTM

base_export_async/models/delay_export.py
```diff
@@ -0,0 +1,106 @@
...
+            'datas_fname': name,
+            'type': 'binary',
+            'to_delete': True,
+        })
+
+        url = "{}/web/content/ir.attachment/{}/datas/{}?download=true".format(
+            self.env['ir.config_parameter'].sudo().get_param('web.base.url'),
+            attachment.id,
+            attachment.name,
+        )
```
sbidoul, line 94: Is there a need to secure access to these attachments, so only the user who requested the export can read them?

base_export_async/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_delay_export,delay.export.user,model_delay_export,,1,1,1,1
```
sbidoul: Since delay.model only has only model methods, is this ACL required?

base_export_async/models/delay_export.py
```diff
@@ -0,0 +1,106 @@
...
+from odoo.addons.web.controllers.main import CSVExport, ExcelExport
+from odoo.exceptions import Warning
+
+_logger = logging.getLogger(__name__)
+
+
+class DelayExport(models.Model):
+
+    _name = 'delay.export'
+    _description = 'Allow to delay the export'
```
sbidoul, line 21: add `_auto = False` since there are only model methods?

sbidoul (changes requested):
to avoid merging while my comments are pending

base_export_async/security/ir.model.access.csv
apineux: Yes to avoid Warning in the log (travis doesn't like that)

base_export_async/models/delay_export.py
apineux: Maybe we need to secure it yes but the email that contains the link is already secured and send to the user only. I'm open to suggestion.

apineux: We are going to link the attachment to a record of the delay.export. We are going to create a ir.rule on the delay.export so that only the user himself can read it. Therefore the attachment will be protected.

base_export_async/models/delay_export.py
apineux: Based on the security comment above, we are going to create records to be able to specify rules. The _auto = False is therefore not necessary anymore.

base_export_async/models/delay_export.py
```diff
@@ -0,0 +1,120 @@
...
+        })
+
+        url = "{}/web/content/ir.attachment/{}/datas/{}?download=true".format(
+            self.env['ir.config_parameter'].sudo().get_param('web.base.url'),
+            attachment.id,
+            attachment.name,
+        )
+
+        time_to_live = self.sudo().env. \
+            ref('base_export_async.attachment_time_to_live').value
```
sbejaoui: use `self.env["ir.config_parameter"].sudo().get_param("attachment.time.to.live", 7)`  instead.

the config_parameter can be deleted and replaced by another one with the same key.

base_export_async/data/config_parameter.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
```
sbejaoui:
```suggestion
<odoo noupdate="1">
```

base_export_async/models/delay_export.py
```diff
@@ -0,0 +1,120 @@
...
+                </span></p>
+                """).format(url, expiration_date),
+            'auto_delete': True,
+        })
+
+    @api.model
+    def cron_delete(self):
+        time_to_live = self.env. \
+            ref('base_export_async.attachment_time_to_live').value
+        date_today = fields.Date.from_string(fields.Date.today())
```
sbejaoui: with odoo 12, `fields.Date.today()` return a `datetime.date` object. the conversion is not needed here.

base_export_async/models/delay_export.py
```diff
@@ -0,0 +1,120 @@
...
+        date_to_delete = fields.Date.to_string(
+            date_today + relativedelta(days=-int(time_to_live)))
+        self.search([('create_date', '<=', date_to_delete)]).unlink()
```
sbejaoui, line 129: You can search with `datetime.date` object without conversion to string.

sbejaoui (changes requested):

sbejaoui (approved):

base_export_async/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+Arnaud Pineux (ACSONE SA/NV) authored the initial prototype.
```
guewen, line 1: It looks more than a prototype to me ;)

base_export_async/models/delay_export.py
```diff
@@ -0,0 +1,119 @@
...
+import logging
+import json
+import operator
+import base64
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models, _
+from odoo.addons.queue_job.job import job
+from odoo.addons.web.controllers.main import CSVExport, ExcelExport
+from odoo.exceptions import Warning
```
guewen:
```suggestion
from odoo.exceptions import UserError
```

base_export_async/models/delay_export.py
```diff
@@ -0,0 +1,119 @@
...
+    _name = 'delay.export'
+    _description = 'Allow to delay the export'
+
+    user_id = fields.Many2one('res.users', string='User', index=True)
+
+    @api.model
+    def delay_export(self, data):
+        params = json.loads(data.get('data'))
+        if not self.env.user.email:
+            raise Warning(_("You must set an email address to your user."))
```
guewen:
```suggestion
            raise UserError(_("You must set an email address to your user."))
```

base_export_async/models/delay_export.py
```diff
@@ -0,0 +1,119 @@
...
+    def export(self, params):
+        export_format = params.get('format')
+        raw_data = export_format != 'csv'
+
+        model_name, fields_name, ids, domain, import_compat, context = \
+            operator.itemgetter('model', 'fields', 'ids',
+                                'domain', 'import_compat', 'context')(params)
+        user = self.env['res.users'].browse([context.get('uid')])
+        if not user or not user.email:
+            raise Warning(_("The user doesn't have an email address."))
```
guewen:
```suggestion
            raise UserError(_("The user doesn't have an email address."))
```

guewen (commented):
Great! A few suggestions

gva-acsone (approved):
Functional test +1

base_export_async/data/config_parameter.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
+    <record id="attachment_time_to_live" model="ir.config_parameter">
+        <field name="key">attachment.time.to.live</field>
```
simahawk: I'd rename this as `base_export_async.attachment_TTL`

base_export_async/models/delay_export.py
```diff
@@ -0,0 +1,119 @@
...
+            columns_headers = field_names
+        else:
+            columns_headers = [val['label'].strip() for val in fields_name]
+
+        if export_format == 'csv':
+            csv = CSVExport()
+            result = csv.from_data(columns_headers, import_data)
+        else:
+            xls = ExcelExport()
+            result = xls.from_data(columns_headers, import_data)
```
simahawk: I'd split the part till here to `_get_file_content` or similar

base_export_async/models/delay_export.py
```diff
@@ -0,0 +1,119 @@
...
+
+        time_to_live = self.env['ir.config_parameter'].sudo(). \
+            get_param('attachment.time.to.live', 7)
+        date_today = fields.Date.today()
+        expiration_date = fields.Date.to_string(
+            date_today + relativedelta(days=+int(time_to_live)))
+
+        odoo_bot = self.sudo().env.ref("base.partner_root")
+        email_from = odoo_bot.email
+        self.env['mail.mail'].create({
```
simahawk, line 106: I'd add `# TODO: move to mail template`

base_export_async/static/src/js/data_export.js
```diff
@@ -0,0 +1,89 @@
...
+                        name: 'id',
+                        label: _t('External ID')
+                    });
+                }
+
+                var export_format = this.$export_format_inputs
+                    .filter(':checked').val();
+
+                framework.blockUI();
+                this._rpc({
```
simahawk, line 58: can we split/organize the code here as well? :innocent:

simahawk (commented):
really nice!

sbidoul:
@apineux can you attend the comments from @simahawk ?

base_export_async/data/config_parameter.xml
apineux: Usually we don't write xml_id with Uppercase so I will not do that. Maybe attachment_ttl

base_export_async/models/delay_export.py
apineux: We could but I'm not sure that it is necessary. The _get_file_content will not be used somewhere else and the method seems quite clear to me as it is.

base_export_async/models/delay_export.py
apineux: I don't like to add #TODO. Either we decide to add a mail template (Which is a good Idea) either we let it this way.

base_export_async/static/src/js/data_export.js
apineux: I copy the original method from Odoo Standard and I just add a small part to send the RPC to another method. Keep it this way may be easier as we understand that this method has been overwritten. If it is really a problem I can rework

base_export_async/data/config_parameter.xml
simahawk: keep it lower case, doesn't matter. as long as it contains the module name ;)

base_export_async/models/delay_export.py
simahawk: is not just fore making it reusable, is that this big method does too many things IMO :)
and in any case, if you want to delegate file content manipulation that's the option.

base_export_async/static/src/js/data_export.js
simahawk: then mention this at the top of the file and add a comment where you've added your custom code ;)

base_export_async/models/delay_export.py
simahawk: NOTE: I don't want to block you here. This can be optional if you place a nice TODO comment saying "we should split this and that... because... bla bla" ;)

sbidoul: @apineux @simahawk [ROADMAP.rst](https://github.com/OCA/maintainer-tools/tree/master/template/module/readme) is the best place for this and should settle the matter :)

apineux:
@sbidoul @simahawk The requested (or suggested) changes are done. I hope it is ready to be merge now

simahawk (approved):
LG. Remember to squash on merge.

base_export_async/models/delay_export.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # TODO : move to email template
+        odoo_bot = self.sudo().env.ref("base.partner_root")
+        email_from = odoo_bot.email
+        self.env['mail.mail'].create({
+            'email_from': email_from,
+            'reply_to': email_from,
+            'email_to': user.email,
+            'subject': _("Export {} {}").format(
+                model_name, fields.Date.to_string(fields.Date.today())),
```
sbidoul: @apineux last thing. Here could you use the model description (in the user language if it is translatable). It will be nicer for the user to see Invoice rather than account.invoice in the email subject.

sbidoul (approved):

guewen (approved):

Issue #143, opened 2019-04-08
[Feature] Add a cancel state on queue.job


benwillig:
Actually, when you don't want to process a job, you would use the "Set to done" button. The problem is that you can't easily see which job have been processed, and which job have been manually done by a user.

So the idea is to add a "Cancel" state. A wizard would also be available in the tree view to cancel more than one job at a time.

I would like your opinion @guewen before starting to develop it.

guewen:
This was an option that we decided to exclude at the beginning because it doesn't bring so much and add complexity. I still have this opinion.
The result of a job set to done contains "Manually set to done by ..." which allows to find such jobs. Also, I think that manually setting jobs to done should be an exception, in an ideal world, it should never happen. If you need to do it regularly, I'm curious to know the use case.

What we could consider however, is to ask for a reason in the "set to done" wizard, which would be stored in the result: seems even better as you keep a trace of why you needed to do that.

benwillig:
Thanks for your feedback. I totally understand. So if we create a wizard which will be called in the 'Set to done' button to ask a reason, it would be ok for you? This wizard would also be available in the tree view.

guewen:
> hanks for your feedback. I totally understand. So if we create a wizard which will be called in the 'Set to done' button to ask a reason, it would be ok for you? This wizard would also be available in the tree view

Yes, the wizard already exists, but the "reason" is hardcoded as "Manually set to done by ..." for now.
I think the only change is to add a field on the wizard and set it as the job's result.

guewen:
Done in #149

PR #149, opened 2019-05-22
[10.0] queue_job: reason to set to done

baimont:
fixes https://github.com/OCA/queue/issues/143

### DIFF
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -161,10 +161,28 @@ class QueueJob(models.Model):
             job_.store()
 
     @api.multi
-    def button_done(self):
-        result = _('Manually set to done by %s') % self.env.user.name
+    def action_done(self, reason=None):
+        result = _(
+            u"Manually set to done by {}"
+        ).format(self.env.user.name)
+        if reason:
+            result = _(
+                u"{} with reason: {}"
+            ).format(result, reason)
         self._change_job_state(DONE, result=result)
         return True
+
+    @api.multi
+    def button_done(self):
+        _logger.warning('deprecated, replaced by action_done()')
+        return self.action_done()
+
+    @api.multi
+    def button_done_ask_reason(self):
+        action = self.env.ref(
+            'queue_job.action_set_jobs_done'
+        ).read()[0]
+        return action
 
     @api.multi
     def requeue(self):
@@ -296,10 +314,12 @@ class SetJobsToDone(models.TransientModel):
     _name = 'queue.jobs.to.done'
     _description = 'Set all selected jobs to done'
 
+    reason = fields.Text(string='Reason to set to done')
+
     @api.multi
     def set_done(self):
         jobs = self.job_ids
-        jobs.button_done()
+        jobs.action_done(reason=self.reason)
         return {'type': 'ir.actions.act_window_close'}
 
 
--- a/test_queue_job/tests/test_job.py
+++ b/test_queue_job/tests/test_job.py
@@ -480,6 +480,14 @@ class TestJobModel(common.TransactionCase):
         self.assertEqual(stored.result,
                          'Manually set to done by %s' % self.env.user.name)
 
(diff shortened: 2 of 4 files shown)


benwillig (approved):

queue_job/models/queue_job.py
```diff
@@ -161,11 +161,22 @@ def _change_job_state(self, state, result=None):
             job_.store()
 
     @api.multi
-    def button_done(self):
-        result = _('Manually set to done by %s') % self.env.user.name
+    def button_done(self, reason):
```
guewen: Since it cannot be called by a button with a reason, I don't think it makes sense to keep this naming. The reason should not be mandatory, should be a named argument. Also the named argument makes the calls more clear `action_done(reason=_('foo'))`

Rougly:

```python

@api.multi
def action_done(self, reason=None):
    # conditional message, adds "with reason:" only if we have a reason
    self._change_job_state(DONE, result=result)
    return True

@api.multi
def button_done(self):
    _logger.warning('deprecated, replaced by action_done()...
    return self.action_done()

@api.multi
def button_done_ask_reason(self):
    action = self.env.ref(
        'queue_job.action_set_jobs_done'
    ).read()[0]
    return action
```

queue_job/models/queue_job.py
```diff
@@ -296,10 +307,15 @@ class SetJobsToDone(models.TransientModel):
     _name = 'queue.jobs.to.done'
     _description = 'Set all selected jobs to done'
 
+    reason = fields.Text(
+        string='Reason to set to done',
+        required=True,
```
guewen: I don't see any strong reason to have a required field.

guewen (changes requested):
Thanks for the PR!

I don't think it should be mandatory. Beside this, a suggestion for the method names to prevent confusion.

baimont:
> Thanks for the PR!
> 
> I don't think it should be mandatory. Beside this, a suggestion for the method names to prevent confusion.

Hello guewen,
Thanks for the review and your comments.
Modifications have been made. Don't hesitate to comment again.

queue_job/models/queue_job.py
```diff
@@ -161,11 +161,29 @@ def _change_job_state(self, state, result=None):
...
     @api.multi
-    def button_done(self):
-        result = _('Manually set to done by %s') % self.env.user.name
+    def action_done(self, reason=None):
+        result = _(
+            u"Manually set to done by {}"
+        ).format(self.env.user.name)
+        if reason:
+            result = _(
+                u"{} with reason : {}"
```
guewen:
```suggestion
                u"{} with reason: {}"
```

Usually no space before colon in English

queue_job/models/queue_job.py
```diff
@@ -296,10 +314,12 @@ class SetJobsToDone(models.TransientModel):
     _name = 'queue.jobs.to.done'
     _description = 'Set all selected jobs to done'
 
+    reason = fields.Text(string='Reason to set to done')
+
     @api.multi
     def set_done(self):
         jobs = self.job_ids
-        jobs.button_done()
+        jobs.action_done(self.reason)
```
guewen:
```suggestion
        jobs.action_done(reason=self.reason)
```

test_queue_job/tests/test_job.py
```diff
@@ -480,6 +480,14 @@ def test_button_done(self):
         self.assertEqual(stored.result,
                          'Manually set to done by %s' % self.env.user.name)
 
+    def test_action_done(self):
+        stored = self._create_job()
+        stored.action_done(reason="reason")
+        self.assertEqual(stored.state, DONE)
+        self.assertEqual(stored.result,
+                         'Manually set to done by %s with reason : reason'
```
guewen:
```suggestion
                         'Manually set to done by %s with reason: reason'
```

guewen (approved):
A few small suggestions, overall looks good to me, thanks!

PR #150, opened 2019-05-23
[12.0][IMP] add job helper for simplifying testing

rousseldenis:
Cherry-picked 6f007dbb19485e8d06f14dd55f534095379af22d

### DIFF
--- a/queue_job/tests/common.py
+++ b/queue_job/tests/common.py
@@ -1,0 +1,34 @@
+from odoo.addons.queue_job.job import Job
+
+
+class JobCounter:
+
+    def __init__(self, env):
+        super(JobCounter, self).__init__()
+        self.env = env
+        self.existing = self.search_all()
+
+    def count_all(self):
+        return len(self.search_all())
+
+    def count_created(self):
+        return len(self.search_created())
+
+    def count_existing(self):
+        return len(self.existing)
+
+    def search_created(self):
+        return self.search_all() - self.existing
+
+    def search_all(self):
+        return self.env['queue.job'].search([])
+
+
+class JobMixin:
+
+    def job_counter(self):
+        return JobCounter(self.env)
+
+    def perform_jobs(self, jobs):
+        for job in jobs.search_created():
+            Job.load(self.env, job.uuid).perform()


rousseldenis:
@sebastienbeau @lmignon

rousseldenis:
Any news of this ?

guewen (approved):

queue_job/tests/common.py
```diff
@@ -0,0 +1,35 @@
...
+        return len(self.existing)
+
+    def search_created(self):
+        return self.search_all() - self.existing
+
+    def search_all(self):
+        return self.env['queue.job'].search([])
+
+
+class JobMixin(unittest.TestCase):
```
simahawk: the base class here probably is irrelevant

simahawk (approved):
We should probably document this together w/ https://github.com/OCA/queue/issues/152 and the usage of `test_queue_job_no_delay` ctx key.

queue_job/tests/common.py
rousseldenis: I admit I've just cherry-picked (and it works) :smiley:

simahawk: sure it works :) I'm just saying that you have to use this mixin together w/ an odoo test class which already inherits from base unittest class.
Plus, here you need and env to work, which means you cannot really use this class standalone. 
Anyway, is just a detail...

lmignon:
@rousseldenis What's the status of this one? Do you plan to fix the last comment?

simahawk:
@rousseldenis ping :)

queue_job/tests/common.py
sbidoul:
```suggestion
class JobMixin:
```

queue_job/tests/common.py
```diff
@@ -0,0 +1,35 @@
+import unittest
+from odoo.addons.queue_job.job import Job
+
+
+class JobCounter(object):
```
sbidoul:
```suggestion
class JobCounter:
```

queue_job/tests/common.py
sbidoul: So I made 2 suggestions (no need to inherit from `object` in python 3)

rousseldenis:
@simahawk @sbidoul Done

sebastienbeau (approved):
LGTM, thanks for porting it

Issue #129, opened 2019-03-07
[Feature] Job dependencies


guewen:
Some details may be inaccurate as I didn't verify all my assumptions.

# Implementation Draft

First let's state what we would expect from dependencies

## Design assumptions

1. We can delay a job A with a dependency on another job Z: A will start only when Z is in state 'done'
2. We can delay jobs A and B with a dependency on another job Z: both A and B will start only when Z is in state 'done'
3. We can delay a job A with a dependency on other jobs Z and Y: A will start only when Z and Y are in state 'done'
4. We can delay jobs A and B with a dependency on other jobs Z and Y: A will start only when Z and Y are in state 'done'
5. When job Z finishes and A depends on it, A will get the result of Z
6. When job Z and Y finishes and A depends on them, A will get a list with the results of Z and Y
7. When job Z and Y finishes and A and B depends on them, A and B will get a list with the results of Z and Y

Assumptions 1. and 2. are probably the easiest to implement, and we could start with them, but we have to think the datamodel and API to be sure we can do 3. and 4. as well.

## Datamodel and workflow

* Add a new field on queue_job: `dependencies`. This is a JSON field with a dict expressing the dependencies, with 2 keys: `depends_on`, `reverse_depends_on`.  We don't want FKs, jobs are not supposed to be deleted and if they are, we just have to ignore the irrelevant references.
* Add a new field for the result that is passed to the dependencies, JSON serialized (we can pass recordsets, ...). As we currently store the result as a message for the end-user, not sure how we should handle it
* We add a new state `wait_dependency`, these jobs will be ignored by the jobrunner out of the box
* When a job finishes, the job runner looks if the job has a non-empty `reverse_depends_on` list. If it finds dependent jobs, for each job in `reverse_depends_on`, it has to read the `depends_on` list, check if all the jobs in this list are done and in such case, change the state of the dependent job to 'pending'.

Using the denormalization with a `reverse_depends_on` field means that we only never have to run a `SELECT ... WHERE depends_on = ?` to find the dependencies. We have a cost penalty (besides the cost to read `reverse_depends_on`) only when we actually have dependencies.

[...]

PR #151, opened 2019-06-30
[11.0][ADD] queue_job_batch

etobella:
When you split a single process in small processes managed with queue you may have problems in order to know that everything finished. For example, if you try to execute 1000 contracts, you may want to split them by customer and execute a job for each customer. But you are unable to know that everything finished.
This PR tries to solve this issue creating Queue Job Groups.
It allows to show your jobs in a grouped form in order to know better the
results.

```

  from odoo import models, fields, api
  from odoo.addons.queue_job.job import job

  class MyModel(models.Model):
     _name = 'my.model'

     @api.multi
     @job
     def my_method(self, a, k=None):
         _logger.info('executed with a: %s and k: %s', a, k)


  class MyOtherModel(models.Model):
      _name = 'my.other.model'

      @api.multi
      def button_do_stuff(self):
          group = self.env['queue.job.group'].get_new_group('Group')
          for i in range(1, 100):
              self.env['my.model'].with_context(
                  job_group=group
              ).with_delay().my_method('a', k=i)
          group.enqueue()
```

In the snippet of code above, when we call ``button_do_stuff``, 100 jobs
capturing the method and arguments will be postponed.  It will be executed as
soon as the Jobrunner has a free bucket, which can be instantaneous if no other
job is running.

Once all the jobs have finished, the grouper will be marked as finished.

You can manage your group jobs from the Systray. A new button will be shown
with your currently executing job groups and the recently finished job groups.

A testing module is installed in order to test that everything works as expected.


etobella:
ping @jbeficent @jarroyomorales

queue_job_group/static/src/xml/systray.xml
```diff
@@ -0,0 +1,62 @@
...
+               aria-expanded="false" title="Attendance Warnings" href="#">
+                <i class="fa fa-list"/>
+                <span class="o_notification_counter badge"/>
+            </a>
+            <ul class="o_mail_navbar_dropdown dropdown-menu o_job_group_navbar_dropdown" role="menu">
+                <li class="o_mail_navbar_dropdown_top">
+                    <div>
+                        <button type="button" class="btn btn-sm o_filter_button active">Job groups</button>
+                    </div>
+                    <button type="button" class="btn btn-sm o_view_all_group_jobs"> View All </button>
```
jarroyomorales: Maybe you could add this style to the view all button, I think it will fit more
```css
.o_view_all_group_jobs {
    .btn-link;
    padding: 5px;
}
```

jarroyomorales (approved):
LGTM, and very useful module :+1:

guewen:
Looks interesting, I'll take some time for reviewing asap.
Out of curiosity, did you see these specs and had you any reason to follow another way? https://github.com/OCA/queue/issues/129

etobella:
Dependencies should be used on a graph of jobs, this tries to solve when you need to know when one has finished (order is not important). It could be a special case of your implementation, but it is easier to solve this use case. I don't know if this happens in other cases, but we could use it in a lot of places (validation of multiple invoices, generation of contract invoices...)

guewen:
> Dependencies should be used on a graph of jobs, this tries to solve when you need to know when one has finished (order is not important). 

If I rephrase, the groups are used to show the group's state to users, but they are not used to delay dependent jobs,  do I understand correctly?

> It could be a special case of your implementation, but it is easier to solve this use case.

Yes probably, I think that with my implementation proposal, we could propose a view of the graph.

Anyway, my proposal is only what it is now: a proposal with no time yet to implement it, when you have an implementation for your use case. I only want to avoid 2 solutions to grow concurrently in different directions, but I recon that should not be the case here.

etobella:
> If I rephrase, the groups are used to show the group's state to users, but they are not used to delay dependent jobs, do I understand correctly?

You are right, the group is only for information.

I made a simple video showing it:
part1

sbidoul:
Nice feature!

A quick question: is the new "waiting" state on queue jobs absolutely necessary?

etobella:
Probably not, but I think that you need to wait to enqueue all the jobs. For that reason, I created this state.

Do you think that it is better to create the job without checking the state?

guewen:
As `group.enqueue()` is meant to be called in the same transaction, jobs will not be picked up by the jobrunner until the transaction is committed, so I tend to say that the state is not needed.

guewen:
> `` is meant to be called in the same transaction

Or maybe it isn't?
As #129 mentions a similar state (`wait_dependency` there), we could maybe share a single state for both use cases?

Regarding the naming, would you accept to rename Group to Batch to prevent a confusion with the Group notion of #129? (I'd like to keep the Group term in the dependencies if possible to keep the same terms than celery)

etobella:
sure! I will change it batch.
Can I change the state behavior or do you think it is better to wait?

etobella:
@guewen Changed to batch :smile:

etobella:
I think that this PR should wait for #154.

Or maybe we can post here the creation of a new state for waiting. What do you think @guewen? 
Should we include this approach to the dependencies system?

queue_job_batch/README.rst
```diff
@@ -0,0 +1,118 @@
...
+     def my_method(self, a, k=None):
+         _logger.info('executed with a: %s and k: %s', a, k)
+
+
+  class MyOtherModel(models.Model):
+      _name = 'my.other.model'
+
+      @api.multi
+      def button_do_stuff(self):
+          group = self.env['queue.job.group'].get_new_group('Group')
```
JordiBForgeFlow: replace 'group' with 'batch'.

queue_job_batch/README.rst
```diff
@@ -0,0 +1,118 @@
...
+          for i in range(1, 100):
+              self.env['my.model'].with_context(
+                  job_group=group
```
JordiBForgeFlow: job_batch =batch

JordiBForgeFlow (approved):
Tested with https://github.com/OCA/queue/pull/155 and works very well!

etobella:
@jbeficent Comments attended :smile:

sbidoul:
I'm still not convinced a new job state is needed for this, even if jobs are created in several transactions.

Would it be a problem if jobs starts while being in a draft batch?

etobella:
Well, it is not a problem for me to remove the state. @jbeficent What do you think?

etobella:
@sbidoul I removed the state. I think this solutions is a better approach for you comments

guewen (approved):
Thanks for the changes, this is great :)

queue_job_batch/models/queue_job_batch.py
```diff
@@ -0,0 +1,118 @@
...
+        return self.write({'is_read': True})
+
+    @api.model
+    def get_new_batch(self, name, **kwargs):
+        vals = kwargs.copy()
+        if 'company_id' in self.env.context:
+            company_id = self.env.context['company_id']
+        else:
+            company_model = self.env['res.company']
+            company_model = company_model.sudo(self.user_id)
```
sbidoul: self.user_id is suspicious in a model method.
Why is this special processing to get company_id necessary, and why is this get_new_batch necessary over a regular create?

queue_job_batch/models/queue_job_batch.py
```diff
@@ -0,0 +1,118 @@
...
+        compute='_compute_job_count',
+    )
+
+    @job
+    def check_state(self):
+        self.ensure_one()
+        if self.state == 'draft' and any(
+            job.state not in ['pending', 'enqueued'] for job in self.job_ids
+        ):
+            self.write({'state': 'progress'})
```
sbidoul, line 78: Here there might be a race condition in this scenario:
1. create batch and jobs in one transaction
2. create more jobs in the same batch in another transaction

If between 1 and 2 all jobs finish, the batch will go to done state before jobs created in 2 are seen.

So maybe batch.enqueue() is necessary after all.

etobella: But the batch.enqueue has no sense without the pending state... :thinking: 
Maybe we could add the state on `queue_job` module in order to be reused in other modules?

queue_job_batch/models/queue_job_batch.py
etobella: I took the same logic used on `queue.job`:
https://github.com/OCA/queue/blob/11.0/queue_job/job.py#L445
Which should be the right approach?

queue_job_batch/models/queue_job_batch.py
sbidoul: I mean you could have the following states for the batch:
1. jobs being added (`draft`)
2. all jobs created (this state is set when the developer signals that all jobs are created, I thought it was the purposed of `enqueued`, but maybe another name is better)
3. `progress`: at least one job not in (pending, enqueued)
4. `done`: all jobs done

And a draft batch remains draft even if all jobs are done.

etobella: Done! :smile:

queue_job_batch/models/queue_job_batch.py
```diff
@@ -0,0 +1,124 @@
...
+        compute='_compute_job_count',
+    )
+    failed_percentage = fields.Float(
+        compute='_compute_job_count',
+    )
+
+    def enqueue(self):
+        self.filtered(lambda r: r.state == 'draft').write({
+            'state': 'enqueued'
+        })
```
sbidoul, line 68: You probably need to call `check_state` here, in case all jobs are already done when you enqueue the batch.

queue_job_batch/models/queue_job_batch.py
sbidoul: > Which should be the right approach?

I don't know. What I'm quite sure is that `self.user_id` on a model method is not going to work.
@guewen any idea?

queue_job_batch/models/queue_job_batch.py
etobella: You are right. Change applied

queue_job_batch/models/queue_job_batch.py
etobella: Maybe we can use `self.env.uid`

queue_job_batch/README.rst
```diff
@@ -0,0 +1,117 @@
...
+  class MyOtherModel(models.Model):
+      _name = 'my.other.model'
+
+      @api.multi
+      def button_do_stuff(self):
+          batch = self.env['queue.job.batch'].get_new_batch('Group')
+          for i in range(1, 100):
+              self.env['my.model'].with_context(
+                  job_batch=batch
+              ).with_delay().my_method('a', k=i)
```
sbidoul, line 58: re-add `batch.enqueue()` here

etobella: Added, it was on the DESCRIPTION.rst, Only the oca-gen-addon-readme command was not executed

sbidoul: Ah sorry about that.

queue_job_batch/models/queue_job_batch.py
sbidoul: > Maybe we can use self.env.uid

Yes that would be better, certainly. I still don't understand the goal of all this complex logic with sudo etc. I guess a simple create with default values for `user_id` and `company_id` would do the job, but I may be missing something.

etobella: Reviewing the original code, it has to be replaced by `self.env.uid`. I took the code, but there was something before that was necessary
https://github.com/OCA/queue/blob/11.0/queue_job/job.py#L444
Change applied.

sbidoul (approved):

PR #155, opened 2019-07-08
[11.0][base_import_async] - Add dependency to  queue_job_batch

JordiBForgeFlow:
This PR adds to the 'base_import_o' module the dependency with 'queue_job_batch' so that the end user can track the progress of the load.

Peek 2019-07-08 13-16

### DIFF
--- a/base_import_async/models/base_import_import.py
+++ b/base_import_async/models/base_import_import.py
@@ -6,6 +6,7 @@
 
 import csv
 import base64
+from datetime import datetime
 from os.path import splitext
 from io import StringIO, TextIOWrapper, BytesIO
 
@@ -147,6 +148,8 @@ class BaseImportImport(models.TransientModel):
         else:
             header_offset = 0
         chunk_size = options.get(OPT_CHUNK_SIZE) or DEFAULT_CHUNK_SIZE
+        batch_name = '%s (%s)' % (file_name, datetime.today())
+        batch = self.env['queue.job.batch'].get_new_batch(batch_name)
         for row_from, row_to in self._extract_chunks(
                 model_obj, fields, data, chunk_size):
             chunk = str(priority - INIT_PRIORITY).zfill(padding)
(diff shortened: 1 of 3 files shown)


JordiBForgeFlow:
cc @etobella @sebastienbeau @sbidoul this feature would be nice in data loads.

etobella (approved):
LGTM :+1: 
code Review and tested locally :smile:

sbidoul:
Yes, that's very nice :)

base_import_async/models/base_import_import.py
```diff
@@ -161,13 +164,15 @@ def _split_file(self, model_name, translated_model_name,
...
-            delayed_job = self.with_delay(
+            delayed_job = self.with_context(
+                job_batch=batch).with_delay(
                 description=description, priority=priority)._import_one_chunk(
                     model_name=model_name,
                     attachment=attachment,
                     options=options)
             self._link_attachment_to_job(delayed_job, attachment)
             priority += 1
+        batch.enqueue()
```
guewen, line 175: This line will be to remove since https://github.com/OCA/queue/pull/151/commits/8e2dcacda8c04bd18c17a62d3351514463f40bc1

guewen (commented):
Very cool :)

base_import_async/models/base_import_import.py
guewen: no, it has to stay now :)

etobella:
I have tested it with the last changes of #151 and it works perfectly :smile:

etobella:
Just wondering if something like https://github.com/OCA/server-ux/blob/11.0/base_tier_validation/migrations/11.0.1.2.0/post-migrate.py#L8 is necessary installing queue_job_batch.
WDYT?

sbidoul:
@etobella that should not be necessary, the new dependency will be installed automatically.

etobella:
Yes, my mistake, I always had issues with this kind of changes, but I did manage it wrongly.

sbidoul (approved):

guewen (approved):

PR #158, opened 2019-07-16
pre_init_hook to re-queue stuck jobs on module init

andriusdaunoravicius:
this is a concept, but I was wandering if there are cases why is this not already included, since jobs after the start of odoo that are in states `enqueued` and `started` are always stuck jobs.

### DIFF
--- a/queue_job/__init__.py
+++ b/queue_job/__init__.py
@@ -2,3 +2,8 @@ from . import controllers
 from . import fields
 from . import models
 from . import jobrunner
+
+
+def pre_init_hook(cr):
+    cr.execute("update queue_job set state='pending' "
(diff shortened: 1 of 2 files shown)


GSLabIt:
pre_init_hook acts at installation time only IIRC.

andriusdaunoravicius:
Yeah, sorry about that, but I was wandering why this SQL is not triggered on module load when you start odoo. That would re-queue all stuck jobs. Is there a reason?

GSLabIt:
IIRC in previous version there was a load_hook but actually it has been removed

andriusdaunoravicius:
any particular reason?

GSLabIt:
dunno .. fell free to ask Odoo SA 😬

PR #163, opened 2019-08-03
[12.0][FIX] Autovacuum cron job

naglis:
`state` is not set to `code` on the _AutoVacuum Job Queue_ `ir.cron` [data record](https://github.com/OCA/queue/blob/3682aae2970ccc1c6c3e8ee3a76464b09cbe8118/queue_job/data/queue_data.xml#L12-L22), so the [default action](https://github.com/odoo/odoo/blob/ead2d19e4a9cb25e1953e8528face4090079ae1d/odoo/addons/base/models/ir_actions.py#L375) `object_write` is assigned to it on creation, as a result, the cron job does nothing when it runs (jobs older than 30 days are not being deleted).

Affected versions: 12.0, this was already fixed for 11.0 in #112

I've also added a test case and a migration to update the record for existing installations (since the record is marked as `noupdate="1"`).

If, like me, you did not notice it for quite a while and now have hundreds of thousands of old jobs (or even more), running the fixed autovacuum cron job will most likely fail due to memory error or a timeout (depends on the setup, of course). To work around that, I've copied the original _AutoVacuum Job Queue_ cron job record and used this for `code`:

```python
deadline = datetime.datetime.now() - datetime.timedelta(days=31)
model.search([('date_done', '<=', deadline)], limit=10000).unlink()
```
and setup to run this Cron job e.g. every 5 minutes to clear away the build-up.

### DIFF
--- a/queue_job/__manifest__.py
+++ b/queue_job/__manifest__.py
@@ -2,7 +2,7 @@
 
 
 {'name': 'Job Queue',
- 'version': '12.0.1.0.0',
+ 'version': '12.0.1.0.1',
  'author': 'Camptocamp,ACSONE SA/NV,Odoo Community Association (OCA)',
  'website': 'https://github.com/OCA/queue/queue_job',
  'license': 'AGPL-3',
--- a/queue_job/data/queue_data.xml
+++ b/queue_job/data/queue_data.xml
@@ -18,6 +18,7 @@
             <field name="interval_type">days</field>
             <field name="numbercall">-1</field>
             <field eval="False" name="doall"/>
+            <field name="state">code</field>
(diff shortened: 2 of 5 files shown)


guewen (approved):
Thanks for this PR of quality and the detailed description!

sebalix (approved):
Approving, but should we instead cherry-pick https://github.com/OCA/queue/pull/112/commits/a89d1d8912c2e7993ffbed5030a544a4b8c0b7ed regarding the XML data?

pedrobaeza (approved):
I think it's OK as it is

sbejaoui (approved):

PR #168, opened 2019-08-30
Rename Chinese translation files

liweijie0812:


pedrobaeza (approved):
It's more than a renaming, but I suppose it's OK at least for avoiding the problem.

liweijie0812:
@sbidoul hello，Help deal with the merge, the Chinese translation file does not match, you know.

PR #178, opened 2019-10-15
pre-commit: unexclude installable addons

sbidoul:


sbidoul:
@guewen a small gotcha. I excluded uninstallable addons when creating the pre-commit configs.

It looks like it's actually red: https://travis-ci.org/acsone/queue/jobs/598124398

guewen:
Oh yes, didn't see that.
Locally, I don't have this error. Which makes no sense, because `dateutil` is supposed to be a "known library". I fear that we have this issue: https://github.com/OCA/pylint-odoo/issues/247, as `dateutil` is listed in `.isort.cfg`.

`pylint-odoo` uses `isort` to find which libs are standard libs or third-party libs; it uses a trick to force known odoo libs as standard library, the rest will be requested in "external depedencies":

https://github.com/OCA/pylint-odoo/blob/67fa4893ce04b93e19612a577bc4563f0bfa62dc/pylint_odoo/checkers/modules_odoo.py#L356-L359

But when we have an `.isort.cfg` file, (randomly it seems,) the `known_third_party` list of the file will take precedence over the hardcoded pylint-odoo list.

guewen:
I proposed a correction on pylint-odoo: https://github.com/OCA/pylint-odoo/pull/263

PR #180, opened 2019-10-16
[13.0] Remove deprecated default company

### LINKED ISSUE #179: not a issue: _company_default_get is deprecated in 13
https://github.com/OCA/queue/blob/f89a94bdc07b8ec3eab472623aa4c17f7de2a9a6/queue_job/job.py#L498

use self.env.company

guewen:
Now the company is part of the env, not to get from the context or from
company_default_get.

Fixes #179

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -493,11 +493,7 @@ class Job(object):
         if "company_id" in env.context:
             company_id = env.context["company_id"]
         else:
-            company_model = env["res.company"]
-            company_model = company_model.with_user(self.user_id)
-            company_id = company_model._company_default_get(
-                object="queue.job", field="company_id"
-            ).id
+            company_id = env.company.id
         self.company_id = company_id
         self._eta = None
         self.eta = eta


sbidoul:
I had a conversation yesterday with @Olivier-LAURENT about a situation where we found it preferable to have no company_id on jobs (many companies and the jobs administrator monitoring jobs for all companies).

Does it make sense (to have no company_id on jobs) and what is the new to do it if company_default_get does not exist anymore (perhaps my question is not specific to this PR, I've not investigated that much yet for 13).

guewen:
We actually never had a use case where we needed a `company_id` on the jobs. In fact, it's been even confusing for us many times when we cannot see all the jobs.

I consider jobs being technical and low-level objects, and IMO users that have access to them should be able to see all of them. (I like to compare to a Celery queue, you wouldn't give different permissions on the queue's monitoring screen).

For the record, the origin of the company rules is https://github.com/OCA/connector/commit/bff6ca3983d660931cd79362d6950e1472e3d0d0

Now, the feature is there and probably used.

About what replaces `company_default_get()`, as the default behavior used to return the company of the user, `self.env.company` should be equivalent. If `company_default_get()` was overridden with custom behavior, then I don't know.

guewen:
> about a situation where we found it preferable to have no company_id on jobs (many companies and the jobs administrator monitoring jobs for all companies).

Maybe we should have a configuration option that only enable or disable the record rule on the company.

lmignon:
> > about a situation where we found it preferable to have no company_id on jobs (many companies and the jobs administrator monitoring jobs for all companies).
> 
> Maybe we should have a configuration option that only enable or disable the record rule on the company.

:+1:

sbidoul (approved):

lmignon (approved):

guewen:
Well, I have a test to fix ;)

guewen:
@sbidoul  @lmignon so a quick search in odoo addons shows that using `force_company` and various `company_id` keys in the context is still used to change the company, contrarily to my understanding that we would have a `with_company()` (there is not). So I see no reason to remove this part. I pushed a fixup commit.

lmignon:
@guewen Thank your for the investigations.

sbidoul:
It looks that this still WIP at Odoo: https://github.com/odoo/odoo/pull/36804 ?

simahawk (approved):

guewen:
Squashed

Feyensv:
@guewen
Raphael spoke about the with_company during his OXP presentation because we were still expecting to merge it post OXP in 13.0, but it was retargeted to master in the end for stability reasons.

But indeed, the force_company/company/company_id(s) context key support will be removed soon in Odoo codebase, to only consider the generic allowed_company_ids (which should be accessed through env.company and env.companies).

PR #182, opened 2019-10-30
[13.0] pre-commit: unexclude installable addons

guewen:
Tentative to fix #178 

I added a commit on top of @sbidoul's one to temporarily remove the `missing-manifest-dependency` check from pylint.

It can be removed as soon as OCA/pylint-odoo#263 is included in a pylint-odoo release.

The risk of having a new dependency in the queue addons is lower than excluding every addons from pre-commit checks.


sbidoul:
LGTM.

/ocabot merge

PR #186, opened 2019-10-31
[12.0][IMP] fix documentation

foutoucour:
Only tilde for titles.
And move to USAGE as it is more related to USAGE than ROADMAP


foutoucour:
@simahawk @sbidoul here the fix of titles. I also move the tips to the USAGE section

sbidoul (approved):
thanks!

guewen (approved):
Thanks!

/ocabot merge

PR #188, opened 2019-11-07
[10.0][ADD] queue_job: allow to define vacuum removal interval per channel

Cedric-Pigeon:
Closes https://github.com/OCA/queue/issues/187

PR #192, opened 2019-11-08
[13.0] queue_job: allow to define vacuum removal interval per channel

guewen:
Port of #188


lmignon (approved):
Thank you for the forward port

PR #193, opened 2019-11-08
[UPD] Update queue_job_subscribe.pot

yucer:
Try to generate a .pot file.

When the folder `i18n` exist and the `.pot` file is missing some errors appear.

Adding one little translation fix.


yucer:
Something confused. Which changes do trigger the update the POT files ?

Is the **MAKEPOT** of the *matrix* session of `.travis.yml` not enough ? Or it should be in the `global` section ?

sbidoul:
The root cause is that queue_job_subscribed has not been migrated to 12 (it has installable=False). 

It should be migrated first then translated on weblate.

PR #195, opened 2019-11-18
[12.0] Port jobrunner as a worker

guewen:
Continuing the work started here (#97) but for 12.0.
I tested (locally) different conditions with good results.
I'll include it in a real project to have real life conditions.


sbejaoui (approved):
Works fine, 

thank you

Issue #152, opened 2019-07-03
[Feature] Mock with_delay()


guewen:
I use a common pattern to run tests with jobs, using a context manager to mock `with_delay()`.
When I have to test code that contains a `with_delay()`, I write 2 different kinds of tests:

- Tests of the outer code with a mock of `with_delay()`. The `mock_with_delay` context manager returns 2 mocks: the first allow to check that `with_delay()` was called and with which arguments, the second to check which job method was called and with which arguments.
- Tests of the job method itself, directly called with the arguments to test.

The context manager:

```python
    @contextmanager
    def mock_with_delay(self):
        with mock.patch(
            'odoo.addons.queue_job.models.base.DelayableRecordset',
            name='DelayableRecordset',
            spec=True,
        ) as delayable_cls:
            # prepare the mocks
            delayable = mock.MagicMock(name='DelayableBinding')
            delayable_cls.return_value = delayable
            yield delayable_cls, delayable
```

Link: https://github.com/camptocamp/connector-jira/blob/b25480c5d6bd65856e53444df10870e7e1a6300c/connector_jira/tests/common.py#L300-L310

Example of "outer code" test: https://github.com/camptocamp/connector-jira/blob/0ca4261b3920d5e8c2ae4bb0fc352ea3f6e9d2cd/connector_jira/tests/test_batch_timestamp_import.py#L43-L76
Example of "job method" test: https://github.com/camptocamp/connector-jira/blob/0ca4261b3920d5e8c2ae4bb0fc352ea3f6e9d2cd/connector_jira/tests/test_import_task.py#L34-L46

This is a reminder that I have to include this in queue job tests with proper documentation :)

guewen:
Done in 12.0 and 13.0 (at least), closing.

PR #197, opened 2019-11-22
[12.0] Add export_async_schedule

### LINKED ISSUE #196: [RFC] planned exports on base_export_async
We'll write a new feature based on `base_export_async`, where you can choose a model, an export list, a domain, a list of recipient users and a frequency. A new model will hold this configuration to be able to configure as many planned exports as we want.
A cron will generate a `delay_export` job based on these.

Any comment, ideas, proposals?
Also, should we include it in `base_export_async` or add this in an module depending on it?

cc @apineux @lmignon

guewen:
* Modify `base_export_async` to support sharing an exported file with several users
* Add `export_async_schedule`

Remaining to do:

* [x] add some tests
* [x] add documentation
* [x] add migration script to move existing relation data from `delay_export.user_id` to `delay_export.user_ids`

Screencast of the feature

Closes #196


guewen:
Included #152 in a9410a8 since I needed it.

lmignon (approved):
Thank you @guewen LGTM (Code review only) Thank you also to make me discover another useful python lib 'freezegun' !

guewen:
Squashed the fixup.


> Thank you also to make me discover another useful python lib 'freezegun' !

This one makes wonder for writing tests :)

PR #198, opened 2019-11-27
[10.0][FIX] queue_job: vaccuum query must be based on channel complete name

Cedric-Pigeon:
If you use hierarchical channels, the name stored on the queue job in the complete path.
So the search for vacuuming should use it to retrieve jobs to clean

PR #200, opened 2019-11-27
[12.0][FIX] queue_job: vaccuum query must be based on channel complete name

guewen:
Port of #198


guewen:
Straightforward port, as the build passes, I merge.

/ocabot merge

PR #205, opened 2019-12-08
[12.0][ADD] report_async -- run all reports asyncronously.

kittiu:
The new menu "Report Center" is the central place to host all your reports in one place.
From here, there are 2 ways to launch the report,

1. Run Now - run report immediately as per normal.
2. Run Background - put the report execution to queue job.

Peek 2019-12-08 21-59

TODO:
- [x] Test Script


kittiu:
@guewen @pedrobaeza 
In your opinion, which repo this module should be?

kittiu:
Ready for review.

simahawk:
what about `OCA/report-engine`?

kittiu:
@simahawk At first I think it is not really the "engine". And the main feature is more with "async". If it is here we don't need to depends on other repo.
But I am also ok, if it make more sense for end user.
Please affirm and I can move. :)

simahawk:
well, is not a generic feature, is a specific feature for reports AFAIS.

kittiu:
Moved https://github.com/OCA/reporting-engine/pull/355

PR #211, opened 2020-01-02
[12.0] [FIX] Allow users from group queue_job_batch_user to write on batches

### LINKED ISSUE #202: queue_job_batch: permission issues when user is in group queue_job_batch_user
When the user is only in group queue_job_batch_user
- one must use `sudo` to call `enqueue()` on the new batch
- the user cannot mark job batch as read because `set_read()` raises a permission error

What would be the best approach:
- use `sudo` in `set_read()` and `enqueue()`
- add `write` permission on `queue.job.batch` for the `queue_job_batch_user` group (knowing there is a record rule allowing users to view their own batches only

@etobella @lreficent  what do you think?

cc/ @qgroulard

qgroulard:
Fixes #202


LoisRForgeFlow (approved):
LGTM

Thank you!

etobella (approved):

LoisRForgeFlow:
@sbidoul is it good for you? could you merge?

sbidoul (approved):

PR #215, opened 2020-03-19
[FIX] Flush pending updates immediately when storing job data

em230418:
Odoo 13 introduced [flush](https://github.com/odoo/odoo/blob/f4ae831df40e803d1c4d7f9f39751b849b79021b/odoo/models.py#L5359) method, that is required after you execute `write` method.

Without this, jobs' db records may not be updated after env.cr.commit(), that could lead when job actually is done, while job's db record is not.

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -31,6 +31,7 @@ class RunJobController(http.Controller):
         job.perform()
         job.set_done()
         job.store()
+        env["base"].flush()
         env.cr.commit()
         _logger.debug("%s done", job)
 


guewen:
Hi @em230418, thanks for your PR!

Flush is called on commits: https://github.com/odoo/odoo/blob/f4ae831df40e803d1c4d7f9f39751b849b79021b/odoo/sql_db.py#L388

Do you have more details why it is needed here?

em230418:
> Do you have more details why it is needed here?

While porting metioned saas module I got situation, when `flush_env` does not call `env['base'].flush()`, because `list(Environment.envs)` from https://github.com/odoo/odoo/blob/f4ae831df40e803d1c4d7f9f39751b849b79021b/odoo/sql_db.py#L65 is empty.

em230418:
Deployment example, where I could reproduce it: https://github.com/em230418/odoo-saas-dev-example

guewen:
Could it be related to this: https://github.com/odoo/odoo/pull/48685?

em230418:
> Could it be related to this: [odoo/odoo#48685](https://github.com/odoo/odoo/pull/48685)?

No. As I mentioned above `list(Environment.envs)` is empty, so it does not enter that line.

guewen:
> No. As I mentioned above `list(Environment.envs)` is empty, so it does not enter that line.

Yes, you said this, sorry.

Do you have steps to reproduce (thanks for the deployment example already), if I want to reproduce it, what should I do? 
Do you know or think about a way to find from which place or sequence of actions the `store()` method is called in these conditions (no envs)?

I'd like to understand the root cause of this (no envs) rather than fixing the consequences, if possible.

em230418:
```
git clone https://github.com/em230418/docker-odoo.git -b 13.0-saas_demo docker-odoo-13
cd docker-odoo-13
mkdir -p vendor/OCA
mkdir -p vendor/it-projects-llc
cd vendor/OCA
git clone https://github.com/OCA/web.git -b 13.0
git clone https://github.com/OCA/queue.git -b 13.0
cd ../..
cd vendor/it-projects-llc
git clone https://github.com/itpp-labs/access-addons.git -b 13.0
git clone https://github.com/em230418/saas-addons.git -b 13.0-saas-port
cd ../..
docker-compose build
docker-compose up odoo
# wait until db initializes, then ctrl+c
docker-compose up odoo # again
```

After last `docker-compose up odoo` database called `template_database` in being initialized.

- Then I entered http://saas.127.0.0.1.nip.io (see https://nip.io for more detailes, what it does)
- Job Queue
- This is what I see http://i.imgur.com/dNbiZWO.png
- Let's see what is saas.operator.install_modules:
https://github.com/em230418/saas-addons/blob/13.0-saas-port/saas/models/saas_operator.py#L65-L92

Before `install_modules` finishes, it adds `post_init` job. "Job Queue" from /web shows that `install_modules` has not finished yet, but `post_init` is queued so `install_modules` defacto finished.

Logs of second docker-compose up https://pastebin.com/u8GQfmqh

em230418:
> Do you have steps to reproduce (thanks for the deployment example already), if I want to reproduce it, what should I do?

See above. Using other odoo docker configuration. Also I use 13.0-saas_demo branch instead of 13.0-saas in https://github.com/em230418/docker-odoo.git (edited message above)

em230418:
> Do you know or think about a way to find from which place or sequence of actions the store() method is called in these conditions (no envs)?

No idea

guewen:
@em230418 I've looked into https://github.com/em230418/saas-addons/blob/13.0-saas-port/saas/models/saas_operator.py#L66-L92, IMO the issue is that `IrModuleModule.button_immediate_install` [completely wipes the environments](https://github.com/odoo/odoo/blob/aca17a48280e1c1641fde08076197210c926e5f1/odoo/addons/base/models/ir_module.py#L572), which explain what you observe and the inconsistent state.

This is a very special method which does very unusual things on the environments and registries. I don't think we should add a flush here only because of this single use case. Did you try to add a flush in your method `install_module`? I understand it's weird, but calling a method that install modules and does shenanigans in the registry of modules within the execution of a job is as well, as indeed, the envs will be reset between the beginning and the end of the job.

em230418:
> This is a very special method which does very unusual things on the environments and registries. I don't think we should add a flush here only because of this single use case.

I think otherwise. I think we should add flush here or somewhere else to fix a bug, when job use odoo's built-in feature like modules installing. If not - we should add text to documentation, how to deal with it.

> Did you try to add a flush in your method install_module?

I executed `self.flush()` after executing https://github.com/em230418/saas-addons/blob/dcc4b77d87b6c848f33c32e2cd061be1a8980cbe/saas/models/saas_operator.py#L89 but I don't know what to expect. To do flush method, like in this PR I need to access job object (not record), execute set_done then store with flush.

em230418:
Hi, @guewen!
Let's finally merge this or comment my point above.

guewen:
Hi @em230418, I could finally test (while watching the odoo xp keynote...) and I think the flush should be done after this line: https://github.com/OCA/queue/blob/ac5304c47b853c5a53f00336ba1f3ce2b04191db/queue_job/controllers/main.py#L32

I tested with your use case and it works.

Reasoning:

* The issue you mention can happen only after we call `perform` and the place I linked is the entry point to perform a job
* Calling `flush` on every `store` in every case is less optimal in some situations. In case we are, for instance requeuing a lot of jobs, then updating some properties on the same jobs, we'll have only 1 `UPDATE` per record without `flush`

Also, I'd love more details in the commit message.

Thanks!

em230418:
Hi again, @guewen!
I added `env.flush()` after storing, instead of after set_done, that you suggested. 'cos if I add flush after set_done it would make inconsistency after job creates database in my case.

Also, I added some details in commit.

guewen:
> I added `env.flush()` after storing, instead of after set_done, that you suggested. 'cos if I add flush after set_done it would make inconsistency after job creates database in my case.

of course, it makes sense!

Thanks!

guewen (approved):

PR #220, opened 2020-04-01
[IMP] - Queue Job Batch User have the right to create jobs

sbejaoui:
In order to make the _Queue Job Batch User_  able to use fully job batches, he need to be able to create queue jobs.


sbejaoui:
cc/  @gva-acsone

guewen:
Usually users shouldn't need the permissions to create jobs as they are created through the `Job` class that uses [sudo](https://github.com/OCA/queue/blob/947b20a973c66b28c38e89b2028ab93798b19748/queue_job/job.py#L528-L581). What is the reason you think we need to add it here?

PR #214, opened 2020-03-10
[13.0][FIX][queue_job] Update Multi-Company Access Rules To 13.0

Tatider:

PR #222, opened 2020-04-09
[13.0][FIX] queue_job: multi-company rule

astirpe:
As per migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0


queue_job/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,10 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade  # pylint: disable=W7936
+
+
+@openupgrade.migrate()
```
simahawk: Why do you need openupgrade? It should work OOTB in odoo and you could simply write into the rule.
In any case it should address 13.0.1.0.1 as patch version, no?

simahawk (changes requested):

queue_job/migrations/13.0.1.0.0/post-migration.py
astirpe: This script is not meant to patch the running code. Is meant to handle the migration of the module from v12 to v13 and for this reason it's correct to run it when processing 13.0.1.0.0. When migrating from v12 to v13 the official OCA way is by openupgrade, or am I missing something?

pedrobaeza: Yes, that's it. Any way, I recommend openupgradelib in any possible migration, being major or minor.

pedrobaeza (approved):

queue_job/migrations/13.0.1.0.0/post-migration.py
astirpe: @pedrobaeza for minor migrations I would not recommend openupgradelib since scripts will raise errors in platforms where openupgradelib is not installed, like odoo.sh. See for example https://github.com/OCA/hr/issues/754

simahawk: I see, but my point is: if you don't use openupgrade is going to work anyway. 
You just need a "migrations" folder w/ proper version number.
In this specific case, I don't see the benefit of using it.
I'm not against it, I'm just trying to understand :wink:

simahawk:
@astirpe travis build is :red_circle:

astirpe:
Strange, I'm running tests locally and they all pass. :thinking:

astirpe:
Oops I just noticed that this fix was already proposed in https://github.com/OCA/queue/pull/214

pedrobaeza:
But without the migration script, so I prefer this one.

There's a problem with existing 13.0 users, that will have outdated record rules (as they are noupdate=1), but if we force the migration script to be executed in new minor version (raising for example folder and manifest version to `13.0.1.1.0`), openupgradelib dependency will be forced, having a rabbit hole, so for me better to be quick on merge this for affecting the less number of persons.

astirpe:
I don't understand why travis fails

pedrobaeza:
I think it can be a change upstream in Odoo adding an object not serializable due to any refactor.

pedrobaeza:
@guewen should we merge this any way as explained before the urgency and later check the problem?

simahawk (approved):
if you grant it works, let's unlock it. We can rework it if needed.

simahawk:
> But without the migration script, so I prefer this one.
> 
> There's a problem with existing 13.0 users, that will have outdated record rules (as they are noupdate=1), but if we force the migration script to be executed in new minor version (raising for example folder and manifest version to `13.0.1.1.0`), openupgradelib dependency will be forced, having a rabbit hole, so for me better to be quick on merge this for affecting the less number of persons.

I still don't understand: if you declare this upgrade step for version X.X.X, no matter if you use openupgrade or not, it will run only for that version upgrade.
So, whoever wants to deploy this fix must upgrade the module.
If you come from version 12.x you will upgrade it anyway.

pedrobaeza:
@simahawk yeah, but there's the risk of someone not having openupgradelib installed in their system, breaking the update. Do you understand me?

simahawk:
@pedrobaeza I totally agree! I thought you were advocating in favor of it :smile: 
I would remove openupgradelib here, is useless IMO.

pedrobaeza:
No, that's the point. It's not useless really. The method it uses allows to save a lot of code. You can put any way the equivalent code as replacement for this specific case for amending the situation, but it's not the cleanest one. For general migration to 13.0 guideline, openupgradelib must be used.

astirpe:
@simahawk

- If you are already running module version `13.0.1.0.0` and you want to apply this patch, you have to do it manually since the rule is a `noupdate="1"`.
- If you come from V12, you implicitly are executing openupgrade. With migration scripts proposed here the rule will be correctly updated to V13 way. It's not useless.

The point is to merge this PR asap to limit the number of users affected to this multicompany issue.

simahawk:
> @simahawk
> 
> * If you are already running module version `13.0.1.0.0` and you want to apply this patch, you have to do it manually since the rule is a `noupdate="1"`.

no, you define the migration script as you did change the version to patch, but you don't use openupgradelib. Just run a query or load the record via xmlid. End of the story.

> * If you come from V12, you implicitly are executing openupgrade. With migration scripts proposed here the rule will be correctly updated to V13 way. It's not useless.

You give for granted that you use openupgrade, might not be the case.

> The point is to merge this PR asap to limit the number of users affected to this multicompany issue.

This is why I unlocked it, but keep it mind that by doing so you add a dependency that likely is not there. BTW in this case you should probably add the lib as python external dependency.

astirpe:
> no, you define the migration script as you did but you don't use openupgradelib.
> Just run a query or load the record via xmlid. End of the story

I let you authors/maintainers do that job. I'm not going to spend extra time on this.

> You give for granted that you use openupgrade, might not be the case.

Then we should avoid the use of openupgradelib in all of the OCA modules...

queue_job/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,10 @@
...
+def migrate(env, _version):
+    openupgrade.load_data(
+        env.cr, "queue_job", "migrations/13.0.1.0.0/noupdate_changes.xml"
+    )
```
simahawk:
```suggestion
from odoo import SUPERUSER_ID, api


def migrate(cr, version):
    env = api.Environment(cr, SUPERUSER_ID, {})
    rule = env.ref("queue_job.queue_job_comp_rule", raise_if_not_found=False)
    if rule:
        rule.write({
            "domain_force": "['|',('company_id','=',False),('company_id','in',company_ids)]",
        })
```

simahawk: @astirpe there you go. Rename the folder to 13.0.1.0.1 and drop the additional file

simahawk:
> Then we should avoid the use of openupgradelib in all of the OCA modules...

or the contrary: we should promote and enforce its usage everywhere.
I bet this has not been done because we usually try to stay compliant w/ standard odoo :man_shrugging:

pedrobaeza:
For major migrations, the rule is clear: if you want to migrate OCA modules, you must have `openupgradelib` installed. It's a Python library, and it was done that way for that. 

For minor migrations, `openupgradelib` is a great resource for avoiding code, so I will continue using it in my maintained modules. If someone update its modules and they have an error, just install the library and update again. I have `openupgradelib` installed by default everywhere. We can add it to `requirements.txt` if you think this is going to help.

Tatider (commented):
> But without the migration script, so I prefer this one.

Sorry, about that. I didn't saw these migration recommendation, just fix the problem I'd gone. I would have done migration script earlier if someone had said something.
Please, tell me what did I make wrong during MR creation, that no one paid attention to it? To avoid such situations in the future.
1. I added scripts into my [MR](https://github.com/OCA/queue/pull/214)
2. I made [fix for tests](https://github.com/OCA/queue/pull/214/commits/982aa3a4b2a8feec1b0f11c528742a6757413877) as well - Odoo removed their [patch](https://github.com/odoo/odoo/commit/5bb567dc70e273933086c729f8ca2e9583c950f7#diff-b61f597e5df5c60179429d7364d8a170L255
) for `JSONEncoder.default` method 

<br/>

I don't think I can request changes, but in current MR:
1. License of module is LGPL, but it says [AGPL in migration script](https://github.com/OCA/queue/pull/222/files#diff-44c8244a2f0b6d76258591d43d3c47dfR1)
2. Current version of module is [13.0.1.1.0](https://github.com/OCA/queue/blob/13.0/queue_job/__manifest__.py#L6) - it's already higher that migration script (13.0.1.0.0)
3. `queue_job/readme/HISTORY.rst` wasn't updated

astirpe:
I close this one in favor of https://github.com/OCA/queue/pull/214

PR #223, opened 2020-04-24
[ADD] queue garbage collector methods and cron

jesusVMayor:
cherry pick the @mpanarin  work from https://github.com/OCA/connector/pull/325
Works fine also in 12.0

Replaces https://github.com/OCA/queue/pull/144


guewen (approved):
Thanks @jesusVMayor !

simahawk (approved):

Issue #225, opened 2020-04-28
[Question] License queue modules v11, v10 (LGPL, AGPL)


bobslee:
Regarding the license change to **LGPL-3** in the (queue) modules **v13** and **v12**.

I noticed the license in **v11** and **v10** is still **AGPL-3**.
Would it be possible to apply **LGPL-3** please?

guewen:
Hi, as for OCA/connector, considering we did the change on upper versions, I would approve a pull request switching the license to LGPL-3.

guewen:
I close this as the license has been changed in 10.0 and 11.0. Thanks!

bobslee:
Thanks @guewen !

PR #229, opened 2020-05-24
[10.0] Use more permissive license AGPL3 → LGPL3

bobslee:
This addresses #225, to follow [11.0], [12.0] and [13.0] similar license change.


guewen (approved):
Thank you

PR #241, opened 2020-06-23
[13.0] Fix interrupt error handling on select

### LINKED ISSUE #240: queue_job - OSError: [WinError 10038] An operation was attempted on something that is not a socket
I understand this was raised previously in this issue #65, however I'm experiencing this now as a developer relying on Windows machines for Odoo debugging via Visual Studio.

Prior to 5th June 2020, queue_job would throw the error in the odoo log file, but jobs would still succeed when called regardless. Since the latest commits after 5th June it appears #208 now forces the module to run in a separate thread, even though Odoo config is set to disable workers.
For me this makes it now impossible to debug and test the queue_job module on a Windows machine as the error blocks all code for jobs executing. 
Hoping someone has a workaround or fix for this please? For now I've reverted to using the older version.

**Error from Odoo log prior to 5th June 2020**
```
ERROR ? odoo.addons.queue_job.jobrunner.runner: exception: sleeping 5s and retrying
  File "\queue_job\jobrunner\runner.py", line 494, in run self.wait_notification()
  File "\queue_job\jobrunner\runner.py", line 469, in wait_notification
    conns, _, _ = select.select(conns, [], [], timeout)
OSError: [WinError 10038] An operation was attempted on something that is not a socket
```

**Error in current version**
```
Exception in thread Thread-7:
  File "\queue_job\jobrunner\runner.py", line 524, in run self.wait_notification()
  File "\queue_job\jobrunner\runner.py", line 499, in wait_notification
    conns, _, _ = select.select(conns, [], [], timeout)
OSError: [WinError 10038] An operation was attempted on something that is not a socket 

During handling of the above exception, another exception occurred:
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python37_64\lib\threading.py", line 926, in _bootstrap_inner
    self.run()
  File "\queue_job\jobrunner\__init__.py", line 44, in run self.runner.run()
  File "\queue_job\jobrunner\runner.py", line 529, in run
    if isinstance(e, select.error) and e[0] == 4:
TypeError: 'OSError' object is not subscriptable
```

guewen:
When select has an interruption, it raises an error with an EINTR errno.
The exception's errno is not (no longer?) available with an index:

TypeError: 'OSError' object is not subscriptable

Since Python 3.5, select() now raises an InterruptedError.
Alternatively, we could check if e.errno == errno.EINTR.

Fixes #240

### DIFF
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -524,15 +524,14 @@ class QueueJobRunner(object):
                     self.wait_notification()
             except KeyboardInterrupt:
                 self.stop()
-            except Exception as e:
+            except InterruptedError:
                 # Interrupted system call, i.e. KeyboardInterrupt during select
-                if isinstance(e, select.error) and e[0] == 4:
-                    self.stop()
-                else:
-                    _logger.exception(
-                        "exception: sleeping %ds and retrying", ERROR_RECOVERY_DELAY
-                    )
-                    self.close_databases()
-                    time.sleep(ERROR_RECOVERY_DELAY)
+                self.stop()
+            except Exception:
+                _logger.exception(
+                    "exception: sleeping %ds and retrying", ERROR_RECOVERY_DELAY
+                )
+                self.close_databases()
+                time.sleep(ERROR_RECOVERY_DELAY)
         self.close_databases(remove_jobs=False)
         _logger.info("stopped")


chrisandrewmann:
Thanks @guewen, after testing that certainly fixes the problem and it will run again on Windows.
The previous issue still exists in that error still shows in the logs repeatedly, but I understand this is more of a fundamental problem with how sockets are handled in Windows and maybe more difficult to resolve? Either way, at least the module now runs. Thanks for your help, happy to approve this if needed.

`[WinError 10038] An operation was attempted on something that is not a socket`

guewen:
> The previous issue still exists in that error still shows in the logs repeatedly, but I understand this is more of a fundamental problem with how sockets are handled in Windows and maybe more difficult to resolve? Either way, at least the module now runs. 

Yes, I realize this, but I don't have Windows so I couldn't reproduce and solve the underlying cause.

> Thanks for your help, happy to approve this if needed.
> 

Yes, that'll help to get it merged :)

apollet (approved):

PR #242, opened 2020-06-24
[FIX] force binary file type

vrenaville:
* Purpose
Force attachment for async import as 'binary'
For unknown reason, possibly a "default_type" key in context, the attachment
has been created with a value "product" for the type, which fails with an
exception because this type is invalid.
Even if the root cause is not known yet, forcing the type as binary is more
defensive and always the value we expect.

### DIFF
--- a/base_import_async/models/base_import_import.py
+++ b/base_import_async/models/base_import_import.py
@@ -101,7 +101,8 @@ class BaseImportImport(models.TransientModel):
         attachment = self.env['ir.attachment'].create({
             'name': file_name,
             'datas': datas,
-            'datas_fname': file_name
+            'datas_fname': file_name,
+            'type': 'binary'
         })
         return attachment
 


guewen:
Do you have more details?

guewen:
Could you amend the commit message to put this description?

```
Force attachment for async import as 'binary'

For unknown reason, possibly a "default_type" key in context, the attachment
has been created with a value "product" for the type, which fails with an
exception because this type is invalid.

Even if the root cause is not known yet, forcing the type as binary is more
defensive and always the value we expect.
```

simahawk:
@vrenaville ping :)

sergiocorato (approved):

giarve:
Can this be cherry-picked to 14.0 branch?

Same error, same solution (tested).

Double quotes **"** are used instead of single quotes **'** in 14.0, do you want me to create a PR for this or cherry-picking it is enough?

Issue #246, opened 2020-06-18
Keep database connection alive in job runner wait loop


sbidoul:
The [main loop](https://github.com/OCA/queue/blob/a07d3735c74166b6b1e0a756107fc012cb1bc4a5/queue_job/jobrunner/runner.py#L521-L524) of the job runner waits for NOTIFY events via a `select()` call, with a 60 seconds timeout.

If there are no activity in the queue_job table it seems that tcp keepalives are not sent (in that very specific scenario), causing some intermediaries (such as haproxy) to close the connection, making the jobrunner to restart on a socket error, in turn triggering OCA/queue#245.

A solution (to be tested) is to insert a `conn.execute("SELECT 1")` around [here](https://github.com/OCA/queue/blob/a07d3735c74166b6b1e0a756107fc012cb1bc4a5/queue_job/jobrunner/runner.py#L475).

houzefa-abba:
Another take on this we didn't get around to submit here as the solution is Linux-specific and not optional: https://github.com/xcgd/queue/pull/1

PR #249, opened 2020-07-24
[12.0][IMP] - keep connexion alive in case of job inactivity

sbejaoui:
fixes: #246

### DIFF
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -335,6 +335,11 @@ class Database(object):
             cr.execute(query, args)
             yield cr
 
+    def keep_alive(self):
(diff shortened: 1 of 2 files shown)


guewen:
@sbejaoui Is my understanding correct that it will wait for notifications 60 seconds, then if nothing happened, execute the keep-alive and listen again for 60 seconds?

sbejaoui:
> @sbejaoui Is my understanding correct that it will wait for notifications 60 seconds, then if nothing happened, execute the keep-alive and listen again for 60 seconds?

yes that's the point

guewen (approved):
Thanks @sbejaoui

sbidoul (approved):

PR #252, opened 2020-08-28
Do not save session in /queue_job/runjob route

sbidoul:
Since https://github.com/odoo/odoo/commit/f65528a74e8d3f11cadd8658f5495bba21338199 routes have a `save_session` option.

We can simplify the job runner by making use of it.

This is currently lUNTESTED.

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -34,19 +34,7 @@ class RunJobController(http.Controller):
         env.cr.commit()
         _logger.debug("%s done", job)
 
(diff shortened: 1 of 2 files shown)


guewen:
Much less code, I love it :)
I may try to test it today or tomorrow.

guewen (approved):
@sbidoul tested successfully, to me you can set it as ready

simahawk (approved):
not tested but the change seems legit

lmignon (approved):
No test but if the functionality is provided by Odoo ...

Issue #169, opened 2019-09-09
How to set up queue_job in odoo sh?


elliotxin:
I followed the instruction on the ubuntu server  the queue_job working very well, after that, I try to set up the queue_job in odoo sh, unfortunately, its does's work, when  i go throw the odoo log, one error message show like this:

> Traceback (most recent call last):
File "/usr/local/lib/python3.6/dist-packages/requests/adapters.py", line 440, in send
timeout=timeout
File "/usr/local/lib/python3.6/dist-packages/urllib3/connectionpool.py", line 639, in urlopen
_stacktrace=sys.exc_info()[2])
File "/usr/local/lib/python3.6/dist-packages/urllib3/util/retry.py", line 388, in increment
raise MaxRetryError(_pool, url, error or ResponseError(cause))
urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='localhost', port=8069): Max retries exceeded with url: /queue_job/session (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f25a8441550>: Failed to establish a new connection: [Errno 111] Connection refused',))

Anyone can tell me how to fix this in odoo sh, thanks a lot.

CasVissers-360ERP:
@Cedric-Pigeon @guewen 
Any idea on this? Or is Odoo.sh not yet supported?

CasVissers-360ERP:
@elliotxin 
I got it running, add this to the Odoo config file:

[queue_job]
channels = root:4
host="https://yourenv.odoo.com"
port=443
http_auth_user="youruser"
http_auth_password="yourpassword"

Good luck!

CasVissers-360ERP:
@Cedric-Pigeon @guewen 
Think this issue can be closed. This works.

guewen:
Thanks for the feedback!

acysos:
Hello @CasVissers ,

How do you locate the odoo config file in odoo.sh?

Greetings

CasVissers-360ERP:
@acysos 
nano .config/odoo/odoo.conf

acysos:
@CasVissers thank you

NMisko:
@CasVissers Did you have to do anything else except modifying the config file? Does the amount of database workers make a difference? I can't get it working on my odoo.sh (see #244 )

amigrave:
Hi there, I'm Fabien Meghazi, core developer and responsible of the Odoo.sh platform.
[...]

PR #256, opened 2020-09-30
[13.0] Support multi-nodes with lock on jobrunner

guewen:
Starting several odoo (main) processes with "--load=web,queue_job"
was unsupported, as it would start several jobrunner, which would all
listen to postgresql notifications and try to enqueue jobs in concurrent
workers.

This is an issue in several cases:

* it causes issues on odoo.sh that uses an hybrid model for workers
and starts several jobrunners (https://github.com/OCA/queue/issues/169#issuecomment-693407232)
* it defeats any setup that would use several nodes to keep the service
available in case of failure of a node/host

The solution implemented here is using a PostgreSQL advisory lock,
at session level in a connection on the "postgres" database, which
ensure 2 job runners are not working on the same set of databases.

At loading, the job runner tries to acquire the lock. If it can, it
initializes the connection and listen for jobs. If the lock is taken
by another job runner, it waits and retry to acquire it every 30
seconds.

Example when a job runner is started and another one starts:

    INFO ? odoo.addons.queue_job.jobrunner.runner: starting
    INFO ? odoo.addons.queue_job.jobrunner.runner: already started on another node

The shared lock identifier is computed based on the set of databases
the job runner has to listen to: if a job runner is started with
``--database=queue1`` and another with ``--database=queue2``, they will
have different locks and such will be able to work in parallel.

Important: new databases need a restart of the job runner. This was
already the case, and would be a great improvement, but is out of
scope for this improvement.

### DIFF
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -18,6 +18,35 @@ How does it work?
 * It does not run jobs itself, but asks Odoo to run them through an
   anonymous ``/queue_job/runjob`` HTTP request. [1]_
 
+How does concurrent job runners work?
+-------------------------------------
+
+If several nodes (on different hosts or not) of job runners are started,
+a shared lock ensures that only one job runner works on a database at
+a time. These rules are to take in consideration:
+
+* The identifier of the shared lock is based on the database list provided,
+  so either ``--database``/``db_name`` or all the databases in PostgreSQL.
+* When 2 job runners with the exact same list of databases are started,
+  only the first one will work. The second one will wait and take over
+  if the first one is stopped.
+
+Caveats:
+
+* If 2 job runners have a database in common but a different list (e.g.
+  ``db_name=project1,project2`` and ``db_name=project2,project3``), both job
+  runners will work and listen to ``project2``, which will lead to unexpected
+  behavior.
+* The same applies when no database is specified and all the cluster's databases
+  are used. If a job runner is started on the cluster's databases, a new database
+  is created and a second job runner is started, they'll both work on a same set
+  of databases with unexpected behaviors.
+* PostgreSQL advisory locks are based on a integer, the list of database names
+  is sorted, hashed and converted to an int64, so we lose information in the
+  identifier. A low risk of collision is possible. If it happens some day, we
+  should add an option for a custom lock identifier.
+
+
 How to use it?
 --------------
 
@@ -134,6 +163,7 @@ Caveat
 """
 
 import datetime
+import hashlib
 import logging
 import os
 import select
@@ -152,6 +182,8 @@ from . import queue_job_config
 from .channels import ENQUEUED, NOT_DONE, PENDING, ChannelManager
 
 SELECT_TIMEOUT = 60
+TRY_ACQUIRE_INTERVAL = 30  # seconds
+SHARED_LOCK_KEEP_ALIVE = 60  # seconds
 ERROR_RECOVERY_DELAY = 5
 
 _logger = logging.getLogger(__name__)
@@ -251,7 +283,52 @@ def _async_http_get(scheme, host, port, user, password, db_name, job_uuid):
     thread.start()
 
 
-class Database(object):
+class SharedLockDatabase(object):
+    def __init__(self, db_name, lock_name):
+        self.db_name = db_name
+        self.lock_ident = self.name_to_int64(lock_name)
+        connection_info = _connection_info_for(db_name)
+        self.conn = psycopg2.connect(**connection_info)
+        self.conn.set_isolation_level(ISOLATION_LEVEL_AUTOCOMMIT)
+        self.acquired = False
+        self.try_acquire()
+
+    @staticmethod
+    def name_to_int64(lock_name):
+        hasher = hashlib.sha256()
+        hasher.update(lock_name.encode("utf-8"))
+        # pg_try_advisory_lock is limited to an 8-byte (64bit) signed integer
+        return int.from_bytes(hasher.digest()[:8], byteorder="big", signed=True)
+
+    def try_acquire(self):
+        self.acquired = self._acquire()
+
+    def _acquire(self):
+        with closing(self.conn.cursor()) as cr:
+            cr.execute("SET idle_in_transaction_session_timeout = 60000;")
+            # session level lock
+            cr.execute("SELECT pg_try_advisory_lock(%s);", (self.lock_ident,))
+            acquired = cr.fetchone()[0]
+        return acquired
+
+    def keep_alive(self):
+        query = "SELECT 1"
+        with closing(self.conn.cursor()) as cr:
+            cr.execute(query)
+
+    def close(self):
+        # pylint: disable=except-pass
+        # if close fail for any reason, it's either because it's already closed
+        # and we don't care, or for any reason but anyway it will be closed on
+        # del
+        try:
+            self.conn.close()
+        except Exception:
+            pass
+        self.conn = None
+
+
+class QueueDatabase(object):
     def __init__(self, db_name):
         self.db_name = db_name
         connection_info = _connection_info_for(db_name)
@@ -355,6 +432,11 @@ class QueueJobRunner(object):
         if channel_config_string is None:
             channel_config_string = _channels()
         self.channel_manager.simple_configure(channel_config_string)
+
+        self.shared_lock_db = None
+        # TODO: how to detect new databases or databases
+        #       on which queue_job is installed after server start?
+        self.list_db_names = self.get_db_names()
         self.db_by_name = {}
         self._stop = False
         self._stop_pipe = os.pipe()
@@ -404,11 +486,22 @@ class QueueJobRunner(object):
                 db.close()
             except Exception:
                 _logger.warning("error closing database %s", db_name, exc_info=True)
+
         self.db_by_name = {}
 
+        if self.shared_lock_db:
+            try:
+                self.shared_lock_db.close()
+            except Exception:
+                _logger.warning(
+                    "error closing database %s",
+                    self.shared_lock_db.db_name,
+                    exc_info=True,
+                )
+
     def initialize_databases(self):
-        for db_name in self.get_db_names():
-            db = Database(db_name)
+        for db_name in self.list_db_names:
+            db = QueueDatabase(db_name)
             if db.has_queue_job:
                 self.db_by_name[db_name] = db
                 with db.select_jobs("state in %s", (NOT_DONE,)) as cr:
@@ -484,6 +577,12 @@ class QueueJobRunner(object):
                 for conn in conns:
                     conn.poll()
 
+    def keep_alive_shared_lock(self):
+        self.shared_lock_db.keep_alive()
+
+    def _lock_ident(self):
+        return "qj:{}".format("-".join(sorted(self.list_db_names)))
+
     def stop(self):
         _logger.info("graceful stop requested")
         self._stop = True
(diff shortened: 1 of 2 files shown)


guewen:
Tested only locally for now, but I'm taking reviews.
Also, I'd love to hear feedback from someone using odoo.sh since it's a huge issue on this platform currently.

sbidoul:
Hi Guewen, thanks for tackling this topic.

So if I understand correctly if there are multiple databases we could in theory have some databases handled by one runner and others by another runner?

guewen:
> So if I understand correctly if there are multiple databases we could in theory have some databases handled by one runner and others by another runner?

Yes, that's a possible scenario. If 2 jobrunners are started at the same time for the same set of databases, they'll each race for the lock and could each acquire only some of them.

sbidoul:
If there is more than one active jobrunner (even on different databases) it could overload the system because each will have it's own `root` channel and potentially each will start as many concurrent jobs.

I was wondering if we could acquire only one advisory lock on the `postgres` database ?

StefanRijnhart (approved):
We run a multi node setup with a similar patch that applies an advisory lock to prevent duplicate jobrunners. Works well.

guewen:
> If there is more than one active jobrunner it could overload the system because each will have it's own `root` channel and potentially each will start as many concurrent jobs.

Right, root channels are global across databases. I overlooked this.



> I was wondering if we could acquire only one advisory lock on the `postgres` database ?

It means we have to keep open a new connection on `postgres` (or `template0`?) though.

sbidoul:
Or acquiring the lock on the first database in alphabetical order ? Dangerous if one database gets added, though.

sbidoul:
> postgres (or template0?)

Access to the `postgres` database is (was?) a requirement of Odoo (to be confirmed for v14), so that could be ok ?
Keeping a connection open on `template0` might be an issue as, IIRC, `createdb --template` fails if there are open connections on the template database.

guewen:
> Access to the `postgres` database is (was?) a requirement of Odoo (to be confirmed for v14), so that could be ok ?

Totally, the bus wouldn't work without access to postgres: 
https://github.com/odoo/odoo/blob/73a671050c97f6f65f2a1b9916e330f0a7b30dc6/addons/bus/models/bus.py#L63
 (same in 14.0).

For the record, I thought about using a single connection to `postgres`, with all the notifications going there with the database name in the notification payload, but on the side of the worker that delays a job we must use `NOTIFY` in the transaction of the current database, so that's not an option.

StefanRijnhart:
Also, is a single connection to Postgres not a problem for setups with multiple Odoo instances querying the same postgresql instance?

sbidoul:
> Also, is a single connection to Postgres not a problem for setups with multiple Odoo instances querying the same postgresql instance?

If we make the advisory lock name configurable that may work.

sbidoul:
That additional connection to `postgres` would be used for acquiring the advisory lock and nothing else.

guewen:
> Also, is a single connection to Postgres not a problem for setups with multiple Odoo instances querying the same postgresql instance?

I was reaching to the same point. We have a PostgreSQL cluster with several odoo instances, we have one jobrunner per database (`--database=myproject1`) and with a lock in postgres, I would be very annoyed to have a global lock :)

> 
> If we make the advisory lock name configurable that may work.

By default, when we start 2 jobrunners, one with `--database=project1` and a second one with `--database=project2`, we should not require customizing options IMO. It's too easy to miss it.
What about generating the advisory lock name from the list of databases? If I start 2 jobrunners for `--database=project1,project2`, they'll both have the same advisory lock generated, but if I start 2 with different database names, they'll have different locks. (It'll not work well if I start `--database=p1,p2` and `--database=p2,p3` but this could be documented...).

StefanRijnhart:
Yes, and you could normalize (i.e. sort) the list of databases.

guewen:
I pushed a ``--squash`` commit (:warning: let me squash before any merge, I updated the description :)) with the `postgres` shared lock approach.

lilee115:
> Tested only locally for now, but I'm taking reviews.
> Also, I'd love to hear feedback from someone using odoo.sh since it's a huge issue on this platform currently.

@guewen Hi, guewen, thanks  for your work on this issue.

I had a try with this commit on odoo.sh, unfortunately it didn't work in our project. Here is what I did:

- We have a branch, which works with the queue_job in old version locally
- I replaced the queue_job folder with this commit
- Push to github and make the branch as a staging branch on odoo.sh

- step 1: triggered a job, and the new job kept pending
- step 2: Ran `odoo-bin --load=web,queue_job` in shell on odoo.sh, then the pending job was executed
- step 3: continued to trigger some jobs, the new jobs kept pending again
- step 4: terminated the odoo-bin, and re-ran `odoo-bin --load=web,queue_job`, one job was executed
- step 5: terminated and re-ran again, one more job and only one was executed
- It seems like executing only one job every time i did a re-run.

I hope this info would help you in this case. If I did it in a wrong way, please let me know too, I would love to try it again, thx

guewen:
thanks @lilee115 , I guess I should finally try to run it on odoo.sh to investigate

guewen:
@lilee115 it seems the issue you mention was caused by the anonymous session which could not be retrieved. After merging #252 and rebasing my branch on top of it, I could test my branch on odoo.sh, with jobs properly executing.

I pushed a new commit for the required configuration on odoo.sh.

I couldn't test with more than one odoo worker though because I am on a trial project. I don't see why the lock wouldn't work though.

Cedric-Pigeon:
Hello I am facing the same problem as @lilee115  but not on odoo.sh Job runner only runs once and after that jobs remains pending.

guewen:
> Hello I am facing the same problem as @lilee115  but not on odoo.sh Job runner only runs once and after that jobs remains pending.

Do you have logs using the debug level?

Cedric-Pigeon:
Seems that problem only appears on a given server, not elsewhere. So it should be a wrong parameters setting.

nilshamerlinck:
This would be a great built-in feature for HA support!

2 remarks:
- I think we need a `SET idle_in_transaction_session_timeout = 60000;` to make sure that the pg server closes the session and frees the lock if the host is unreachable
- this implementation assumes a stable list of databases; it could be the default case, but it would be nice to have an alternative option to explicitly define a custom `lock_name`; for example `ODOO_QUEUE_JOB_JOBRUNNER_LOCK_NAME=static`

guewen:
>     * I think we need a `SET idle_in_transaction_session_timeout = 60000;` to make sure that the pg server closes the session and frees the lock if the host is unreachable

Very good point!


> 
> * this implementation assumes a stable list of databases; it could be the default case, but it would be nice to have an alternative option to explicitly define a custom `lock_name`; for example `ODOO_QUEUE_JOB_JOBRUNNER_LOCK_NAME=static`

What's your idea behind the custom lock name? Is it to complement #267? If so, releasing and taking a new lock with the new list of database wouldn't be better? Or could a custom lock name be better for another reason?

hoangtrann:
Without this PR, would it work if we have multiple odoo instances connected to 1 db server, in which might have one or multiple database which could be accessed from odoo?

guewen:
> Without this PR, would it work if we have multiple odoo instances connected to 1 db server, in which might have one or multiple database which could be accessed from odoo?

Without this PR, you have to start the jobrunner (--load=queue_job) on only one odoo node. With this PR, you can start it on several node, only one will actually dispatch the jobs, but one of the other node will take over in case the first node fails.

guewen:
I added 

```python
cr.execute("SET idle_in_transaction_session_timeout = 60000;")
```

before taking the lock. 

This comment:

> this implementation assumes a stable list of databases; it could be the default case, but it would be nice to have an alternative option to explicitly define a custom lock_name; for example ODOO_QUEUE_JOB_JOBRUNNER_LOCK_NAME=static

Could be part of a follow-up pull request.

nilshamerlinck:
Hi @guewen 

I did some tests. We need to keep a transaction open if we want to leverage `idle_in_transaction_session_timeout`. 

With [these changes](https://github.com/nilshamerlinck/queue/commit/293f7adb9edc6569f12fc4d36e2013bdd4d85597), I confirm that these scenarios are supported:

1) multi nodes (on different hosts)

- instance1 and instance2 start
- instance1 acquires the lock, and becomes the active jobrunner
- instance2 becomes a standby jobrunner, and retries to acquire the lock every TRY_ACQUIRE_INTERVAL seconds
- somehow instance1 becomes unreachable (network issue)
- after at most 2*SHARED_LOCK_KEEP_ALIVE seconds, the pg server terminates the idle connection
- instance2 takes over as the new active jobrunner
- instance1 connectivity is back to normal, an exception is raised (`psycopg2.DatabaseError: SSL SYSCALL error: Connection timed out` or `psycopg2.OperationalError: server closed the connection unexpectedly`)
- instance1 becomes a standby jobrunner
- somehow instance2 becomes unreachable (network issue)
- instance1 takes over as the new active jobrunner

2) odoo.sh "hybrid" workers

- worker1 starts, becomes the active jobrunner
- worker2 starts, becomes a standby jobrunner
- odoo.sh stops worker1 (graceful stop, so the sharedlock connection is properly closed)
- worker2 becomes the active jobrunner

My only concern in the context of odoo.sh is that in the absence of http trafic, all http workers will be stopped and thus all jobrunners. If you have a scheduled action that creates queue jobs, they will stay pending until the next http request (that will trigger the start of an http worker). One workaround is to use an external service (e.g. pingdom or a custom prometheus) to request the instance periodically.

guewen:
Excellent! Many thanks @nilshamerlinck for the test, fix and details!



> My only concern in the context of odoo.sh is that in the absence of http trafic, all http workers will be stopped and thus all jobrunners. If you have a scheduled action that creates queue jobs, they will stay pending until the next http request (that will trigger the start of an http worker). One workaround is to use an external service (e.g. pingdom or a custom prometheus) to request the instance periodically.

Yes, I suppose you are right. Same story for jobs with an ETA in the future.

yelizariev:
Can't you use an odoo cron for that?

> My only concern in the context of odoo.sh is that in the absence of http trafic, all http workers will be stopped and thus all jobrunners. If you have a scheduled action that creates queue jobs, they will stay pending until the next http request (that will trigger the start of an http worker). One workaround is to use an external service (e.g. pingdom or a custom prometheus) to request the instance periodically.

nilshamerlinck:
Hi @yelizariev

> Can't you use an odoo cron for that?

A cron could help to keep workers warm indeed, but:
- ["Scheduled actions [...] are run on a best effort policy"](https://www.odoo.com/documentation/user/14.0/odoo_sh/advanced/frequent_technical_questions.html#scheduled-actions-do-not-run-at-the-exact-time-they-were-expected)
- having an external service to monitor the instance also brings the benefit of alerting you in case of node/platform downtime ;-)

hoangtrann:
Hi guys, is this good to run on a production instance on Odoo.sh?

amigrave:
> Hi guys, is this good to run on a production instance on Odoo.sh?

Sorry to hijack the thread again but unfortunately this solution will still cause issues on Odoo.sh.
The advisory lock breaks our server consolidation system by preventing the postgresql replication to occur.
As more and more instances using queue_job are causing issues we had to document it as [incompatible with Odoo.sh](https://www.odoo.sh/faq#install_dependencies).
We will check if an alternative locking is possible and advise accordingly (I will keep you posted here but can't provide an ETA for now).

queue_job/jobrunner/runner.py
```diff
@@ -495,16 +605,42 @@ def run(self):
...
                 self.initialize_databases()
                 _logger.info("database connections ready")
+
+                last_keep_alive = None
+
                 # inner loop does the normal processing
                 while not self._stop:
                     self.process_notifications()
                     self.run_jobs()
                     self.wait_notification()
```
thomaspaulb, line 631: I'm not 100% sure what the exact problem is that @amigrave has with locking vs. replication, but if it is the fact that the advisory lock is held for long - maybe a solution can be to always try to re-acquire the lock at the beginning of this loop, and let it go at the end? Then the lock will not be held so long. 

If, by chance, another instance of the job runner is 'first' to acquire the lock, it will just change the master job runner to that one, but there will still only be one running at a time.

queue_job/jobrunner/runner.py
```diff
@@ -18,6 +18,35 @@
...
+  runners will work and listen to ``project2``, which will lead to unexpected
+  behavior.
+* The same applies when no database is specified and all the cluster's databases
+  are used. If a job runner is started on the cluster's databases, a new database
+  is created and a second job runner is started, they'll both work on a same set
+  of databases with unexpected behaviors.
+* PostgreSQL advisory locks are based on a integer, the list of database names
+  is sorted, hashed and converted to an int64, so we lose information in the
+  identifier. A low risk of collision is possible. If it happens some day, we
+  should add an option for a custom lock identifier.
```
thomaspaulb, line 47: Could it be a solution to try to lock the `channel` records using:

```
SELECT name FROM queue_job_channel FOR UPDATE;
```
instead of going for the integer lock?

That could solve both above issues.

queue_job/jobrunner/runner.py
```diff
@@ -18,6 +18,35 @@
...
+* When 2 job runners with the exact same list of databases are started,
+  only the first one will work. The second one will wait and take over
+  if the first one is stopped.
+
+Caveats:
+
+* If 2 job runners have a database in common but a different list (e.g.
+  ``db_name=project1,project2`` and ``db_name=project2,project3``), both job
+  runners will work and listen to ``project2``, which will lead to unexpected
+  behavior.
```
thomaspaulb, line 39: Perhaps hold a lock for each of the database names separately?

simahawk:
@guewen @thomaspaulb any update on this?

Just wondering how far is this PR from being "odoo.sh ready" (if possible at all).
I don't have the full ctx yet, forgive me if I miss some bits :wink: 

Soon we'll have to provide odoo.sh compat for some connector modules which likely will lead us to remove the dependency on `queue_job` from `connector` because of this. 

CC @sebalix

guewen:
Hi @simahawk, I didn't seek for a new solution after @amigrave's last message.

If I had time, maybe I would do an external job runner as a service 😉

Maybe a fallback on cron jobs could be used as a poor's man solution.

queue_job/jobrunner/runner.py
guewen: Long locks on a table will lead to vacuum issues (and probably replication as well)

guewen:
And https://github.com/OCA/queue/pull/256#discussion_r703348074 may be worth trying.

PCatinean:
Hi @amigrave, I know this thread is quite old but there have been some new developments for an HA deployment of queue_job that is also compatible with odoo.sh.

This new approach is being discussed here https://github.com/OCA/queue/pull/607

We are now evaluating this solution vs session level advisory locks but the last thing you mentioned in this PR was that advisory locks caused an issue to the db replication.

If you could find a few minutes to give us some feedback there so we can ensure this will also be compatible with odoo.sh we would greatly appreciate it.

PR #258, opened 2020-10-01
Store worker's PID when a job is started

### LINKED ISSUE #177: [idea] store worker PID on the queue.job record when starting a job
This should open interesting avenues to detect when a job remains in `started` state while the process has actually been killed.

Thoughts?

guewen:
Add a field in the "queue_job" model to store the worker pid.
No usage for new besides "logging" for debugging (helps to find which
process to investigate in case a job is stuck).
Later, it may be useful to improve the recovery of job that stay in the
wrong state when the job runner restarts.

Closes #177

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -5,6 +5,7 @@ import functools
 import hashlib
 import inspect
 import logging
+import os
 import sys
 import uuid
 from datetime import datetime, timedelta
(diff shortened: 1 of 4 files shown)


sbidoul (approved):
Coding with one ear while listening to talks with the other ?

Thanks for implementing this !

simahawk (approved):

Issue #262, opened 2020-10-10
[11.0->12.0] queue_job module is not upgraded correctly (postgres trigger is missing)


legalsylvain:
Hi all, 

I faced an issue, with a database that comes from 11.0, with ``queue_job`` installed, when I migrated the database to 12.0.

After a (loooong !) investigation, the database is not correctly set, because [this script](https://github.com/OCA/queue/blob/12.0/queue_job/hooks/post_init_hook.py#L13) is not executed. 
- The script doesn't exist in V11.
- The script is only executed on the installation of the module (not during the update / upgrade process)

To check if the database is OK, you can run the following script : 
`` select * from information_schema.triggers;``

if there are no result, ``queue_job`` will not work correctly.

I think that moving the script from a ``post_init_hook`` to a ``migrations`` folder could fix the bug. 

What do you think ?

Thanks.

CC : @guewen

pedrobaeza:
It should be in both places. Create a migration script that imports the method (`from odoo.addons.queue_job.hooks import post_init_hook` or similar) and call it in the migration script.

I haven't experimented this problem in my 12.0 migrated databases though. I will check to see.

sbidoul:
@legalsylvain the trigger was moved from the job runner startup to a post init hook quite recently (see https://github.com/OCA/queue/issues/245).

But the trigger did exist in v11, in fact it has never changed since the job runner exists.

Could it be the trigger was lost in a db dump/restore ?

legalsylvain:
Hi @sbidoul and @pedrobaeza. Thanks for taking a time to answer my issue.

> Could it be the trigger was lost in a db dump/restore ?

No. (it would be worrying !)

> But the trigger did exist in v11, in fact it has never changed since the job runner exists.

OK ! I made a summary, but the whole story is that the db is a 8.0, openupgraded from 8.0 to 12.0, using each step. In the V11 step, a module ``pos_picking_delayed`` has a new dependency to ``queue_job``, so ``queue_job`` is installed.
But as it is only an upgrade with ``--stop-after-init``, and ``--workers=0`` and without ``server_wide_modules=queue_job,web``, maybe the ``_initialize()`` function (that was hosting before the creation of the postgres function) is not executed, that could explain my problem, and explain also why it doesn't occure in other regular cases. (like @pedrobaeza instances).

PR #264, opened 2020-10-18
[13.0][IMP] queue_job: check that queue_job_notify trigger exists in db

nilshamerlinck:
Could save time in cases such as https://github.com/OCA/queue/issues/262 :)

### DIFF
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -278,12 +278,28 @@ class Database(object):
                 "SELECT 1 FROM pg_tables WHERE tablename=%s", ("ir_module_module",)
             )
             if not cr.fetchone():
+                _logger.debug("%s doesn't seem to be an odoo db", self.db_name)
                 return False
             cr.execute(
                 "SELECT 1 FROM ir_module_module WHERE name=%s AND state=%s",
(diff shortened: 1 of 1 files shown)


lmignon (approved):

queue_job/jobrunner/runner.py
```diff
@@ -278,12 +278,28 @@ def _has_queue_job(self):
...
+                return False
+            cr.execute(
+                """SELECT COUNT(1)
+                FROM information_schema.triggers
+                WHERE event_object_table = %s
+                AND trigger_name = %s""",
+                ("queue_job", "queue_job_notify"),
+            )
+            if cr.fetchone()[0] != 3:  # INSERT, DELETE, UPDATE
+                _logger.debug(
```
guewen: Since having the module installed without the trigger is never normal, should we log with warn or error level?

lmignon: Good catch.... IMO we should log with error level.

nilshamerlinck: makes sense! fixed

guewen (approved):
Thanks!

sbidoul (approved):
LGTM. Good idea, thanks!

PR #265, opened 2020-10-21
[13.0][ADD] queue_job: innocuous test job for debugging/monitoring purposes

nilshamerlinck:
The goal of this PR is to introduce an easy and innocuous way to confirm that jobs are running fine in a production environment, by triggering the creation of a test job through an http route.

Of course, only accessible to admin users to avoid any abuse :)

### DIFF
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -8,7 +8,7 @@ from odoo import _, api, exceptions, fields, models
 from odoo.osv import expression
 
 from ..fields import JobSerialized
-from ..job import DONE, PENDING, STATES, Job
+from ..job import DONE, PENDING, STATES, Job, job
 
 _logger = logging.getLogger(__name__)
 
@@ -304,6 +304,10 @@ class QueueJob(models.Model):
                 }
             )
         return action
+
+    @job
+    def _test_job(self):
+        _logger.info("Running test job.")
 
 
 class RequeueJob(models.TransientModel):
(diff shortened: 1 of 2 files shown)


guewen (approved):
Hey that's great! I wanted to add a kind of way to create test jobs easily, and didn't think about using a simple controller endpoint. :heart:

queue_job/models/queue_job.py
```diff
@@ -305,6 +305,10 @@ def related_action_open_record(self):
             )
         return action
 
+    @job
+    def _test_job(self):
+        _logger.debug("Running test job.")
```
guewen: Maybe worth to use `info`? Since we create this job on purpose, it's worth to have more visibility here (even if we'll see if the job is done or not eventually).

queue_job/jobrunner/runner.py
```diff
@@ -432,6 +432,7 @@ def process_notifications(self):
                     break
                 notification = db.conn.notifies.pop()
                 uuid = notification.payload
+                _logger.debug("received notification on db %s: %s", db.db_name, uuid)
```
sbidoul: Is that not too noisy?

sbidoul (commented):
:+1:

queue_job/models/queue_job.py
nilshamerlinck: makes sense, fixed

queue_job/jobrunner/runner.py
nilshamerlinck: it is indeed very noisy, and can even be confusing and lead to think that there is an issue while there is none, removed

simahawk (approved):

PR #267, opened 2020-10-25
[13.0][ADD] queue_job: db listener

nilshamerlinck:
The goal of this PR is to introduce an opt-in mechanism to enable the jobrunner to take into account changes happening after server launch.

For now, it is focused on databases:
- addition (installation of the module on a new or existing database)
- removal (uninstallation of the module from a database).

Beyond the post_init/uninstall hooks, an http route is added to enable admins to manually trigger addition/removal events for a given database.

### DIFF
--- a/queue_job/__init__.py
+++ b/queue_job/__init__.py
@@ -3,3 +3,4 @@ from . import fields
 from . import models
 from . import jobrunner
 from .hooks.post_init_hook import post_init_hook
+from .hooks.uninstall_hook import uninstall_hook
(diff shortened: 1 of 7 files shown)


guewen (commented):
Very interesting implementation!
Could you test it for some time already?
I like the fact that it has to be activated explicitly, although it would be worth mentioning it in readme/CONFIGURATION.rst

BTW, you may be interested in https://github.com/OCA/queue/pull/256 too :)

nilshamerlinck:
Hi,

Yes, we've been using it in production for some time already.

Is it ok if I favor documentation through `readme/` files and remove some of the redundant module docstrings?

PR #269, opened 2020-10-28
[ADD] Index to job description and date done

mileo:


guewen:
Hi, could you explain the rationale for these indices?

guewen:
My opinion. I guess you wish them for searching purpose on the UI. IMO it should not be there by default. Adding them has an  impact on the time to create a job and the DB size. They should be added on per-project basis.

mileo:
> My opinion. I guess you wish them for searching purpose on the UI. IMO it should not be there by default. Adding them has an impact on the time to create a job and the DB size. They should be added on per-project basis.

Yes you are right.

I reviewed the queries earlier today, after a good night's sleep, and realized that they really are project needs.

PR #255, opened 2020-09-27
[13.0][IMP] queue_job, add decorator job_auto_delay

kittiu:
Just add a pending TODO from sale_automatic_workflow_job

closes: #238

PR #270, opened 2020-10-28
[ON HOLD][13.0] Add job_auto_delay decorator

### LINKED ISSUE #238: [feature] Add `job_auto_delay` decorator
Reminder to move this decorator in queue_job: 

https://github.com/OCA/sale-workflow/blob/713f66478479bd8be77b4b93789bbdfc4c9121cf/sale_automatic_workflow_job/models/automatic_workflow_job.py#L11-L87

guewen:
This is a new decorator, to use instead of `@job` on methods,
that transforms a method to a method automatically delayed as job when
called.

A typical use case is when a method in a module we don't control is called
synchronously in the middle of another method, and we'd like all the calls
to this method become asynchronous.

Closes #238

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -899,3 +899,78 @@ def related_action(action=None, **kwargs):
         return func
 
     return decorate
+
+
+def job_auto_delay(func=None, default_channel="root", retry_pattern=None):
+    """Decorator to automatically delay as job method when called
+
+    The decorator applies ``odoo.addons.queue_job.job`` at the same time,
+    so the decorated method is listed in job functions. The arguments
+    are the same, propagated to the ``job`` decorator.
+
+    When a method is decorated by ``job_auto_delay``, any call to the method
+    will not directly execute the method's body, but will instead enqueue a
+    job.
+
+    A typical use case is when a method in a module we don't control is called
+    synchronously in the middle of another method, and we'd like all the calls
+    to this method become asynchronous.
+
+    The options of the job usually passed to ``with_delay()`` (priority,
+    description, identity_key, ...) can be returned in a dictionary by a method
+    named after the name of the method suffixed by ``_job_options`` which takes
+    the same parameters as the initial method.
+
+    It is still possible to directly execute the method by setting a key
+    ``_job_force_sync`` to True in the environment context.
+
+    Example:
+    .. code-block:: python
+        class ProductProduct(models.Model):
+            _inherit = 'product.product'
+
+            def foo_job_options(self, arg1):
+                return {
+                  "priority": 100,
+                  "description": "Saying hello to {}".format(arg1)
+                }
+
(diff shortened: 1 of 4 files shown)


lmignon (approved):
LGTM and good idea! (Code review)

queue_job/job.py
```diff
@@ -899,3 +899,79 @@ def decorate(func):
         return func
 
     return decorate
+
+
+def job_auto_delay(func=None, default_channel="root", retry_pattern=None):
+    """Decorator to automatically delay as job method when called
```
simahawk, line 905: nitpicking: new empty line after the 1st one.

jgrandguillaume (approved):
This is so elegant :+1:

queue_job/job.py
```diff
@@ -899,3 +899,79 @@ def decorate(func):
...
+        else:
+            # replace the synchronous call by a job on itself
+            method_name = func.__name__
+            job_options_method = getattr(
+                self, "{}_job_options".format(method_name), None
+            )
+            job_options = {}
+            if job_options_method:
+                job_options.update(job_options_method(*args, **kwargs))
+            else:
```
simahawk: else not required

simahawk (approved):
LG, just small glitches.
BTW what about https://github.com/OCA/queue/pull/255 ?

guewen:
@simahawk I fixed the small glitches and force-pushed directly

etobella (approved):
Great idea @guewen this is something I always needed and had to solve using context keys. This is much more cleaner :smile:

guewen:
Do not merge yet, I'm investigating an error:

```
AttributeError: method cron_actions on stock.buffer(5577,) is not allowed to be delayed, it should be decorated with odoo.addons.queue_job.job.job
```
And I suspect it happens when other modules redefine the same method. To be continued...

guewen:
> And I suspect it happens when other modules redefine the same method.

I confirm this. I hope I can find a way to prevent this.

guewen:
> `AttributeError: method cron_actions on stock.buffer(5577,) is not allowed to be delayed, it should be decorated with odoo.addons.queue_job.job.job`

This is not an issue only with `@job_auto_delay`, we have the error with `@job`.
If I have such dependencies:

```
module_a
  |- module_b
  |- module_c
```

module_b and module_c have no dependencies on each other.

In `module_a`, we have a method `StockMove._update_foo`.
In `module_b`, we add `@job` on `StockMove._update_foo`.
In `module_c`, we have another override of `StockMove._update_foo`, without decorator.

It seems that depending on the loading order, we'll lose the properties added by `@job`.

guewen:
I proposed https://github.com/OCA/queue/pull/275 that works by patching the method instead of using a decorator.
It needs #274 that I still have to finalize.

PR #272, opened 2020-10-29
Store 'su' flag in jobs

### LINKED ISSUE #271: Keep `sudo()` flag
AFAIK, since 13.0, we don't keep the `su` flag, so when we run:

```
record.sudo().with_delay().foo()
```

The job `foo` will not be executed as `su`.
We should add an `func_env` json field where we can store this flag, the same field could contain the context (see #121).

guewen:
"record.sudo().with_delay().foo()" generates a job that will be run
as sudo, keeping the current user, following the new implementation of
sudo/with_user of 13.0.

Closes #271

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -257,7 +257,7 @@ class Job(object):
     @classmethod
     def _load_from_db_record(cls, job_db_record):
         stored = job_db_record
-        env = job_db_record.env
+        env = job_db_record.env(su=stored.func_env.get("su", False))
 
         args = stored.args
         kwargs = stored.kwargs
@@ -529,6 +529,11 @@ class Job(object):
             raise
         return self.result
 
+    def _func_env_values(self):
+        return {
+            "su": self.env.su,
+        }
+
     def store(self):
         """Store the Job"""
         vals = {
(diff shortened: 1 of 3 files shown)


guewen:
What is this travis failure on the branch? it passes on the pull request :/

sebalix (approved):

sebalix:
The build displays on top, could be related:
```
/home/travis/.travis/job_stages: line 900: syntax error near unexpected token `fi'
/home/travis/.travis/job_stages: line 900: `fi'
```
Rebase maybe?

TDu (approved):

guewen:
> The build displays on top, could be related:

@sebalix 
But this "Travis CI - Branch" job is bogus, generally we don't have both "Branch" and "Pull Request", but only "Pull Request".
IMO it should be ignored.

simahawk (approved):

guewen:
It would be cleaner to store the `su` flag in the jsonified recordset actually. I'll update my PR.

guewen:
I'll reopen another implementation.

guewen:
> It would be cleaner to store the `su` flag in the jsonified recordset actually. I'll update my PR.

New implementation in #281, which brings a nice refactoring.
cc @TDu @sebalix @simahawk

PR #274, opened 2020-10-30
[13.0] Deprecate job decorators

### LINKED ISSUE #273: `delayable` attribute lost when the `@job` decorator is not added on base method
Example of error message:

> `AttributeError: method cron_actions on stock.buffer(5577,) is not allowed to be delayed, it should be decorated with odoo.addons.queue_job.job.job`

If I have such dependencies:

```
module_a
  |- module_b
  |- module_c
```

module_b and module_c have no dependencies on each other.

In `module_a`, we have a method `StockMove._update_foo`.
In `module_b`, we add `@job` on `StockMove._update_foo`.
In `module_c`, we have another override of `StockMove._update_foo`, without decorator.

It seems that depending on the loading order, we'll lose the properties added by `@job`.

AFAIU, it happens because of the way the registry builds a new instance based on a recomposed dependency graph, we can't be sure the kept function will be the one which has been decorated and has the `delayable` attribute. We would need to propagate it.

That being said, the `delayable` attribute is there only to check that we decorated the method with `@job`. If we remove the check at https://github.com/OCA/queue/blob/d465bdd320baf4f83a232c4538868df609b19d74/queue_job/job.py#L81-L85, the enqueuing works perfectly.

We can consider removing the need for `@job` (make it optional with default values). That would imply:

* stop storing the `@job` attributes on the method (https://github.com/OCA/queue/blob/d465bdd320baf4f83a232c4538868df609b19d74/queue_job/job.py#L828-L831): `delayable` could be removed, `delay` is deprecated, remains `retry_pattern` and `default_channel` to store elsewhere, best place would be in `queue.job.function` I think as we could change them from the UI
* same for the related actions attrs... https://github.com/OCA/queue/blob/d465bdd320baf4f83a232c4538868df609b19d74/queue_job/job.py#L897-L898
* think about the security implications by not having the `@job` allow list
* `delayable` is used to find the job methods used to create `queue.job.function` (https://github.com/OCA/queue/blob/d465bdd320baf4f83a232c4538868df609b19d74/queue_job/models/base.py#L24-L35), have to think if we can do it another way (create `queue.job.function` the first time it's delayed? but then we don't have the ability to put default options...)

_Originally posted by @guewen in https://github.com/OCA/queue/pull/270#issuecomment-718772150_

guewen:
`@job` and `@related_action` are still supported but emit warnings and
will be removed at some point (code to remove has a TODO with:job-no-decorator:)

The initial reasons for this change are in #273, but it also brings nice side effects:

* the options previously set by the decorators are now stored on
  "queue.job.function" records, which can be explicitly created
  as XML records, which mean they can now be **edited by users**
* **any method can be used with "with_delay()"**, if no "queue.job.function"
  exists, default values are used (root channel, default related
  action, ...)

This one will require a major bump.

My wish is:

- 12.0: deprecate decorators but show no warning when they are used
- 13.0: deprecate decorators and show warnings when they are used
- 14.0: entirely remove them, as we are starting migrations, it will be easy to tackle in the modules

Remaining tasks:

* [x] Adapt tests
* [x] Document the configuration (a lot of documentation is in the decorators, move it elsewhere + in readme)
* [x] Figure out how to handle duplicates `queue.job.function` (not big issue though)
* [x] update other modules of the repository
* [x] As any method can be delayed, maybe we should guard create/write (on a subset of fields) on `queue.job` from RPC (OTOH only queue job managers can do it, but at the very least they should not be able to create a raw queue.job record for a method starting with `_`
* [x] More checks/constraints on config, improve views, ...
* [x] Add model / method fields on `queue.job.function` and compute the name from there

Fixes #273


guewen:
No runbot build?

base_import_async/data/queue_job_function_data.xml
```diff
@@ -0,0 +1,19 @@
...
+            name="related_action"
+            eval='{"func_name": "_related_action_attachment"}'
+        />
+    </record>
+    <record
+        id="job_function_base_import_import_import_one_chunk"
+        model="queue.job.function"
+    >
+        <field name="name"><![CDATA[<base_import.import>._import_one_chunk]]></field>
+        <field
```
lmignon, line 16: @guewen Wouldn't it be easier to have two fields: model and method_name and to compute the name from these two fields? This would allow a more natural syntax.

guewen: Yes for sure! I add this to the remaining tasks

guewen: Added in https://github.com/OCA/queue/pull/274/commits/27687b2ed9009edfd7f4baeb4dce0ebcdf025901

sebalix (approved):
LG,
I suppose the website https://odoo-connector.com/  will need an update later :) (but it seems related to https://github.com/OCA/connector/tree/13.0/connector )

guewen:
> I suppose the website https://odoo-connector.com/ will need an update later :)

It is supposed to be refreshed by travis, but IIRC triggered by new commits in OCA/connector

guewen:
> I suppose the website https://odoo-connector.com/ will need an update later :)

But of course it doesn't include the readme's content. We need a global refresh of the doc, ideally its own one.

simahawk (approved):

sebastienbeau:
:+1: Really good change ! LGTM code review

sebastienbeau (approved):

guewen:
@lmignon could you update your review please? :)

lmignon (approved):
Thank you @guewen

queue_job/job.py
```diff
@@ -810,6 +832,25 @@ def retryable_example():
...
+    if default_channel:
+        xml_fields.append('    <field name="channel_id" ref="[insert channel xmlid]"/>')
+    if retry_pattern:
+        xml_fields.append('    <field name="retry_pattern">{retry_pattern}</field>')
+
+    xml_record = (
+        '<record id="job_function_[insert model]_{method}"'
+        ' model="queue.job.function">\n' + "\n".join(xml_fields) + "\n</record>"
+    ).format(**{"method": func.__name__, "retry_pattern": retry_pattern})
+    _logger.warning(
```
rousseldenis, line 848: @guewen IMHO this is too much as it will break all runbot builds in repos that use queues. See https://runbot.odoo-community.org/runbot/build/3450323 for example.

Is it wanted ? Or maybe this can go into INFO.

guewen: You are right, I didn't realize runbot would not agree. We should move it to info, and move the XML snippet to debug because it is too verbose at the moment. I'll try to open a PR tomorrow.

PR #275, opened 2020-11-02
[13.0] Add method to patch a method to be automatically delayed

### LINKED ISSUE #238: [feature] Add `job_auto_delay` decorator

guewen:
This patch method has to be called in `_register_hook`.

When a method is patched, any call to the method will not directly
execute the method's body, but will instead enqueue a job.

When a `context_key` is set when calling `_patch_job_auto_delay`,
the patched method is automatically delayed only when this key is
`True` in the caller's context. It is advised to patch the method
with a `context_key`, because making the automatic delay *in any
case* can produce nasty and unexpected side effects (e.g. another
module calls the method and expects it to be computed before doing
something else, expecting a result, ...).

A typical use case is when a method in a module we don't control is called
synchronously in the middle of another method, and we'd like all the calls
to this method become asynchronous.

It relies on #274 that deprecates the `@job` decorator
Replaces #270 

Closes #238

### DIFF
--- a/queue_job/models/base.py
+++ b/queue_job/models/base.py
@@ -1,6 +1,7 @@
 # Copyright 2016 Camptocamp
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
 
+import functools
 import inspect
 import logging
 import os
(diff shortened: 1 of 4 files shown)


guewen:
TODO: Currently, the patched method is automatically delayed and can be forced to be synchrone with a context key. We should probably make it explicit by expecting a key in the context to use autodelay: if someone calls the method elsewhere, and expect to have the computation of the method to be done synchronously, having a delay fully automatic can introduce silent and vicious errors.

sebalix:
@guewen is this PR still a draft?

guewen:
> @guewen is this PR still a draft?

Well, it's ready but as I had to include commits from #274 , I wasn't sure how to proceed

guewen:
note to myself: worth to add a doc in readme

sebastienbeau (approved):
LGTM great improvement ! queue_job rocks !

simahawk (approved):

PR #276, opened 2020-11-06
[12.0] Fix channel name in jobs delayed for crons

guewen:
The channel must be 'root.ir_cron', not 'ir_cron'.


guewen:
Travis fails because of #279

guewen:
Rebased

hparfr (approved):

PR #280, opened 2020-11-13
Remove warning when @job decorator is used

guewen:
It makes the runbot builds fail. There is no urgency to remove the
decorator even if we should incentivize the removal because it is
no longer supported, hence the info log. Also, show only the XML snippet
when debug logs are active, because it is too verbose.

Following #274


guewen:
cc @rousseldenis

rousseldenis (approved):

simahawk (approved):

PR #281, opened 2020-11-13
[13.0] Refactor storage of job records with "sudo" support

guewen:
Replaces #272 

Previously, the "main" record, the one on which the job works, was stored in 3 fields: `model_name`, `user_id` and `record_ids` (and the support for any new field such as sudo or context would need a new field). The records of `args` and `kwargs` are serialized in `JobSerialized` fields.

This PR unifies the way they are stored: the "main" record is now stored in a `JobSerialized` field as well.
The support of the `su` flag (which means the record was `sudo()`-ed), is added in the second commit, which shows how we benefit from the refactor.

The original fields are kept:

* `model_name` is stored for the UI (used in search or groupby)
* `user_id` is stored for the same reason, but also has a `inverse` method that updates the serialized record
* `record_ids` is kept only for backward compatibility, but could be removed in 14.0

### DIFF
--- a/queue_job/fields.py
+++ b/queue_job/fields.py
@@ -27,18 +27,29 @@ class JobSerialized(fields.Field):
 
     _slots = {"_base_type": type}
 
-    _default_json_mapping = {dict: "{}", list: "[]", tuple: "[]"}
+    # these are the default values when we convert an empty value
+    _default_json_mapping = {
+        dict: "{}",
+        list: "[]",
+        tuple: "[]",
+        models.BaseModel: lambda env: json.dumps(
+            {"_type": "odoo_recordset", "model": "base", "ids": [], "uid": env.uid}
+        ),
+    }
 
     def __init__(self, string=fields.Default, base_type=fields.Default, **kwargs):
         super().__init__(string=string, _base_type=base_type, **kwargs)
 
     def _setup_attrs(self, model, name):
         super()._setup_attrs(model, name)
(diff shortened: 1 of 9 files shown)


queue_job/fields.py
```diff
@@ -27,18 +27,29 @@ class JobSerialized(fields.Field):
...
     _slots = {"_base_type": type}
 
-    _default_json_mapping = {dict: "{}", list: "[]", tuple: "[]"}
+    # these are the default values when we convert an empty value
+    _default_json_mapping = {
+        dict: "{}",
+        list: "[]",
+        tuple: "[]",
+        models.BaseModel: lambda env: json.dumps(
+            {"_type": "odoo_recordset", "model": "base", "ids": [], "uid": env.uid}
```
simahawk, line 36: shall we store `record.context` too?

simahawk (approved):
LG (just code review)

queue_job/fields.py
guewen: This clearly opens this possibility, this is definitely where the context has to be stored. I mentioned it in the commit message :)
However, there are some questions about what should be allowed or not to store in the context (cf #121), also whether we should or not always store it (can take useless space?), so it should not be part of this PR.

guewen: created #283

TDu (approved):

guewen:
/ocabot merge nobump

(version is changed in the PR already)

PR #284, opened 2020-11-19
[FIX] Fixed #261. Retryable Job Exception...

### LINKED ISSUE #261: RetryableJobError does not Rollback Transaction
Hi all,

Code in question is here:

wpichler:
...must also throw the exception up to the core, so that the rollback can happen.


guewen:
Could you check why the build is failing please? It is related to the lint.

guewen:
Fixed lint in #311, so closing this one the build is failing.

guewen:
Thanks @wpichler

PR #286, opened 2020-11-20
[14.0] mig queue job rebased

sebastienbeau:
The queue job in V14 is ready \o/

PR #288, opened 2020-11-28
14.0 mig queue job

dnplkndll:
ondelete= 'cascade' or 'set null' ?


guewen:
> ondelete= 'cascade' or 'set null' ?

Cascade :+1: 

I can't see if you added something compared to #286. I opened #288 as some changes were needed. I close this one, let me know if there is anything I missed in #288 that I should integrate.

Thanks!

PR #293, opened 2020-12-03
Fix required sudo in Job

zwgshr:
Fix the infinite loop:
odoo.exceptions.AccessError: ("Sorry, you are not allowed to access documents of type 'Queue Job' (queue.job). This operation is allowed for the groups:\n\t- Job Queue/Job Queue Manager - (Operation: read, User: 5679)", None) - - -

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -558,7 +558,7 @@ class Job(object):
 
         db_record = self.db_record()
         if db_record:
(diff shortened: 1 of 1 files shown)


zwgshr:
Already fix...

guewen:
Thanks nevertheless :)

PR #296, opened 2021-01-03
[13.0][FIX] run doctests as part of standard tests

nilshamerlinck:
Odoo v12 introduced [test tags](https://github.com/odoo/odoo/commit/b356b190338e3ee032b9e3a7f670f76468965006).

Since then doctests were not running anymore.

This PR introduces a wrapper for `doctest.DocTestCase` to get doctests to run again as part of standard tests.


guewen (approved):
Uh, good catch!

lmignon (approved):
Thank you @nilshamerlinck

PR #300, opened 2021-01-25
[14.0] queue_job: ensure order of compute complete_name

hparfr:
`record.parent_id.complete_name` can be `False`.
I guess it's because self is not oriented by `.parent_id` so the child is complete_name is computed before the parent.

With this fix, some records can be computed twice. I guess it's ok.

Discovered while installing module (shopinvader_import_product). Got errors during the load of xml views.

### DIFF
--- a/queue_job/models/queue_job_channel.py
+++ b/queue_job/models/queue_job_channel.py
@@ -35,6 +35,10 @@ class QueueJobChannel(models.Model):
             if not record.name:
                 complete_name = ""  # new record
             elif record.parent_id:
+                if not record.parent_id.complete_name:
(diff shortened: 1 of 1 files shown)


guewen:
Just to be sure, do you have this commit in your odoo src? https://github.com/odoo/odoo/commit/9fe4fe404b73e9ab18ab6b3b913f075c64bb6954#diff-404e91050d105f97f8785b94706814e4a6ead40fea25c0ecf9efefa6bea999f5

hparfr:
Hum, may be :) Now that everything is up to date, I can't reproduce it anymore.So....

guewen:
Cool!

PR #302, opened 2021-02-01
[13.0] Remove initial create notification and follower

guewen:
Everytime a job is created, a mail.message "Queue Job created" is
created. This is useless, as we already have the creation date and user,
and nobody will ever want to receive a notification for a created job
anyway.

Removing the auto-subscription of the user that created the job makes
sense as well since we automatically subscribe the queue job managers
for failures, and we don't send other notifications.

It allows to remove a lot of insertions of records (and of deletions
when autovacuuming jobs).

Rough benchmark when generating jobs for ddmrp buffers (it includes some base
time to fetch the ddmrp buffers too). 9191 jobs were created:

executed with | duration
-------------------|------------
no optimizations | 2m18s
create optimization (#305) | 1m41s
mail.message/follower optimization (current PR) | 1m11s
both optimizations | 35s

### DIFF
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -186,7 +186,10 @@ class QueueJob(models.Model):
             raise exceptions.AccessError(
                 _("Queue jobs must created by calling 'with_delay()'.")
             )
-        return super().create(vals_list)
+        return super(
+            QueueJob,
+            self.with_context(mail_create_nolog=True, mail_create_nosubscribe=True),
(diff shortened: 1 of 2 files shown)


simahawk (approved):
makes sense

guewen:
Added a small benchmark in the description

sebalix (approved):

etobella (approved):

TDu (approved):

PR #304, opened 2021-02-01
[13.0] [FIX] queue_job: create/replace queue_job_notify function at migration

ThomasBinsfeld:
This step seems necessary on migrated databases (9.0 -> 13.0).
The code is taken from the post init hook.

### DIFF
--- a/queue_job/__manifest__.py
+++ b/queue_job/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "Job Queue",
-    "version": "13.0.3.3.0",
+    "version": "13.0.3.3.1",
     "author": "Camptocamp,ACSONE SA/NV,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/queue/queue_job",
     "license": "LGPL-3",
--- a/queue_job/migrations/13.0.3.3.1/post-migration.py
+++ b/queue_job/migrations/13.0.3.3.1/post-migration.py
@@ -1,0 +1,8 @@
+# Copyright 2021 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.queue_job.hooks.post_init_hook import post_init_hook
+
+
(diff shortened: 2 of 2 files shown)


guewen:
Thanks! Are these migrations called on a migrated database?
Would it possible to directly call `odoo.addons.queue_job.hooks.post_init_hook.post_init_hook`?

ThomasBinsfeld:
> Thanks! Are these migrations called on a migrated database?

Yes they are.

> Would it possible to directly call `odoo.addons.queue_job.hooks.post_init_hook.post_init_hook`?

The `DROP TRIGGER IF EXISTS queue_job_notify ON queue_job;` and `CREATE TRIGGER queue_job_notify (...)` are not required here. But it would work just as nice to include them. Do you prefer to call the hook directly, to avoid code duplication?

guewen (approved):

pedrobaeza:
I think this should be included in 12.0, not in 13.0, as it was in that version when the post_init_hook was included.

StefanRijnhart:
Here is a port to 12: https://github.com/OCA/queue/pull/319. @ThomasBinsfeld at what point was the trigger dropped during the migration? I'm now proposing to drop the trigger explicitely here https://github.com/OCA/queue/pull/320 to improve the performance of that particular migration but in my experience it is not dropped otherwise. If it is dropped between 12 and 13 then we will still need the change in this PR and possibly even in later versions. Or maybe `post_init_hook` was not the best place to set the trigger and `_auto_init` could be considered instead?

ThomasBinsfeld:
> Here is a port to 12: #319. @ThomasBinsfeld at what point was the trigger dropped during the migration? I'm now proposing to drop the trigger explicitely here #320 to improve the performance of that particular migration but in my experience it is not dropped otherwise. If it is dropped between 12 and 13 then we will still need the change in this PR and possibly even in later versions. Or maybe `post_init_hook` was not the best place to set the trigger and `_auto_init` could be considered instead?

I think the script in v12 is enough. Closing this PR.

PR #305, opened 2021-02-02
[13.0] Optimize queue.job creation

guewen:
Several fields on queue.job are initialized using computed fields,
then never changed again. On creation of a queue.job record, we'll have
an initial INSERT + at least one following UPDATE for the computed
fields.

Replace all the stored computed fields by a raw initialization of the
values in `Job.store()` when the job is created, so we have only a single 
INSERT.

Highlights:

* as channel is no longer a compute/inverse field, override_channel is
  useless, I dropped it (actually the same value was stored in both
  channel and override_channel as the channel field was stored)
* one functional diff is that now, when changing a channel on a
  job.function, the channel is no longer synchronized on existing jobs,
  it will be applied only on new jobs: actually this is an improvement,
  because it was impossible to change the channel of a job function
  in a large queue_job table as it meant writing on all the done/started
  jobs
* searching the queue.job.function is now cached, as each job using the
  same will run a query on queue_job_function

Rough benchmark when generating jobs for ddmrp buffers (it includes some base
time to fetch the ddmrp buffers too). 9191 jobs were created:

executed with | duration
-------------------|------------
no optimizations | 2m18s
create optimization (current PR) | 1m41s
mail.message/follower optimization (#302) | 1m11s
both optimizations | 35s

### DIFF
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -43,27 +43,22 @@ class QueueJob(models.Model):
         "date_created",
         "model_name",
         "method_name",
+        "func_string",
+        "channel_method_name",
+        "job_function_id",
         "records",
         "args",
         "kwargs",
     )
 
     uuid = fields.Char(string="UUID", readonly=True, index=True, required=True)
-    user_id = fields.Many2one(
-        comodel_name="res.users",
-        string="User ID",
-        compute="_compute_user_id",
-        inverse="_inverse_user_id",
-        store=True,
-    )
+    user_id = fields.Many2one(comodel_name="res.users", string="User ID")
     company_id = fields.Many2one(
         comodel_name="res.company", string="Company", index=True
     )
     name = fields.Char(string="Description", readonly=True)
 
-    model_name = fields.Char(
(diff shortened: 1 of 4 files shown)


queue_job/models/queue_job.py
```diff
@@ -657,9 +610,13 @@ def _parse_retry_pattern(self):
             retry_pattern = {}
         return retry_pattern
 
+    @tools.ormcache("name")
+    def get_by_name(self, name):
+        return self.search([("name", "=", name)], limit=1)
```
simahawk: this is caching the full object. IMO you should cache the ID or a dict w/ full info to be used in `job_config`

guewen: thaaaanks!

guewen: @simahawk  Added the id in JobConfig (which is the only think I lacked), can you check the fixup so I squash?

simahawk (approved):

guewen:
Added a small benchmark in the description

TDu (approved):

PR #308, opened 2021-02-06
[FIX] Fixed #261. Retryable Job Exception must also throw the excepti…

### LINKED ISSUE #261: RetryableJobError does not Rollback Transaction

agent-z28:
…on and rollback

Backport from https://github.com/OCA/queue/commit/ca2af1582f8d8f0b770afe392ada94991319d2f9

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -13,11 +13,7 @@ from odoo import _, http, tools
 from odoo.service.model import PG_CONCURRENCY_ERRORS_TO_RETRY
 
 from ..job import Job, ENQUEUED
-from ..exception import (NoSuchJobError,
-                         NotReadableJobError,
-                         RetryableJobError,
-                         FailedJobError,
-                         NothingToDoJob)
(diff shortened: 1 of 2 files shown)


guewen:
Thanks! Could you fix the travis build please? It complains about the lint.

guewen:
@agenterp I modified the change on #312 and #311, could you include it in your backport please?

StefanRijnhart:
@agenterp what do you think of the original PRs? Can you review them, and if you agree with the changes, update this PR?

agent-z28:
> @agenterp what do you think of the original PRs? Can you review them, and if you agree with the changes, update this PR?

@StefanRijnhart yes i can take care now

StefanRijnhart:
Sorry for the delay. Thanks for taking the last change from the other PR. Now, about this PR: the diff looks different from the other PRs, I believe because you seem to be taking other changes as well, such as https://github.com/OCA/queue/commit/b511c3af52d8735751b1933adbf2e6771a4381f5. If that is the case, can you update your branch and cherry pick these commits so that authorship and history is preserved?

PR #309, opened 2021-02-08
queue_job: add exec time to view some stats

simahawk:


queue_job/models/queue_job.py
```diff
@@ -116,6 +121,14 @@ def _compute_record_ids(self):
         for record in self:
             record.record_ids = record.records.ids
 
+    @api.depends("date_done")
+    def _compute_exec_time(self):
+        for record in self:
+            if not record.date_started or not record.date_done:
+                record.exec_time = False
+                continue
+            record.exec_time = (record.date_done - record.date_started).total_seconds()
```
guewen: I'd prefer not to use a computed field (see https://github.com/OCA/queue/pull/305).
It should be written by https://github.com/OCA/queue/blob/563db616ff33496b75e82b98f9734c8c66637ce4/queue_job/job.py#L685-L690

simahawk: ahhh ok, you are right! Then just this would do the work:
```py
self.exec_time = (self.date_done - self.date_started).total_seconds()
```

simahawk: done

simahawk:
@guewen I guess the test is wrong [here](https://github.com/OCA/queue/blob/563db616ff33496b75e82b98f9734c8c66637ce4/test_queue_job/tests/test_job.py#L150) because `date_started` in theory should be there always if you are calling `set_done`, right? It should call `set_started` before. See https://travis-ci.com/github/OCA/queue/jobs/481537133#L1082

guewen:
>  I guess the test is wrong [here](https://github.com/OCA/queue/blob/563db616ff33496b75e82b98f9734c8c66637ce4/test_queue_job/tests/test_job.py#L150) because `date_started` in theory should be there always if you are calling `set_done`, right? It should call `set_started` before

It's not wrong, for this test, the initial state `started` had no impact on `set_done`. It's a unit test, it doesn't test the full flow.
You added a new dependency (`date_started`) in this unit, so you have to be sure the initial state of the test set it correctly.
You shouldn't call `set_started` before, because then the test would test `set_started` and `set_done`.
You only need to set `job_a.date_started = xxx` as this is the initial state you miss.

simahawk:
better?

guewen (approved):
Yes, thanks!! :heart:

queue_job/models/queue_job.py
```diff
@@ -79,6 +79,7 @@ class QueueJob(models.Model):
     date_started = fields.Datetime(string="Start Date", readonly=True)
     date_enqueued = fields.Datetime(string="Enqueue Time", readonly=True)
     date_done = fields.Datetime(readonly=True)
+    exec_time = fields.Float(help="Time required to execute this job.")
```
sebalix:
```suggestion
    exec_time = fields.Float(help="Time required to execute this job in seconds.")
```

sebalix (approved):
Small comment, approving

etobella (approved):
Code Review

queue_job/models/queue_job.py
```diff
@@ -79,6 +79,7 @@ class QueueJob(models.Model):
     date_started = fields.Datetime(string="Start Date", readonly=True)
     date_enqueued = fields.Datetime(string="Enqueue Time", readonly=True)
     date_done = fields.Datetime(readonly=True)
+    exec_time = fields.Float(help="Time required to execute this job in seconds.")
```
guewen: Do we want `, group_operator="avg"`? It's handy but at the same time I find it a bit misleading on the list view when using group by.

queue_job/views/queue_job_views.xml
```diff
@@ -107,12 +109,24 @@
...
+    <record id="view_queue_job_graph" model="ir.ui.view">
+        <field name="name">queue.job.graph</field>
+        <field name="model">queue.job</field>
+        <field name="arch" type="xml">
+            <graph string="Jobs">
+                <field name="job_function_id" type="row" />
+                <field name="exec_time" type="measure" />
+            </graph>
+        </field>
+    </record>
```
guewen, line 139: I think a pivot view gives better information:

```
<pivot string="Queue Job">
    <field name="exec_time" type="measure"/>
    <field name="job_function_id" type="row"/>
    <field name="date_done" type="col" interval="week"/>
</pivot>
```

Example:
Example of pivot view

Note: in this example, I used `group_operator="avg"` on the field, which I'm not sure we want.

guewen: The average is really cool on this view as we can see if it grows over time...

queue_job/models/queue_job.py
guewen: But better in pivot :shrug:

queue_job/views/queue_job_views.xml
simahawk: looks good. At least is a great start!

guewen:
Pushed a fixup for pivot view + group by avg by default. I added a mention in string and help for the avg to prevent confusion.

guewen:
Hi @simahawk :)

I squashed my commit and rebased, can I remove the work in progress label? I'd like to merge if the build is green.

guewen:
> I squashed my commit and rebased, can I remove the work in progress label? I'd like to merge if the build is green.

Oh no sorry, I couldn't push it because I do no longer have push writes here :laughing: 
Could you do it?

simahawk:
welcome back! :)
It's done!

queue_job/migrations/13.0.3.6.0/pre-migration.py
```diff
@@ -0,0 +1,35 @@
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
```
guewen: ah crap, we already have 13.0.3.6.0 in https://github.com/OCA/queue/blob/263fbaac33a5b559c77a1bb08b0616bb42c78d2f/queue_job/__manifest__.py#L6

@simahawk could you bump the migration directory?

simahawk: ops, missed it. Done!

guewen: Thanks :green_heart:

PR #312, opened 2021-02-10
[13.0] Fix missing rollback on retried jobs

### LINKED ISSUE #261: RetryableJobError does not Rollback Transaction

guewen:
Port from #284 / #311

When RetryableJobError was raised, any change done by the job was not
rollbacked.

Using `raise` would throw the exception up to the core and rollback, but
we would have a stack trace in the logs for each try. Calling rollback
manually (rollback also clears the env) hide the tracebacks, however,
when the last try fails, the full traceback is still shown in the logs.

Fixes #261

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -77,10 +77,10 @@ class RunJobController(http.Controller):
                 if err.pgcode not in PG_CONCURRENCY_ERRORS_TO_RETRY:
                     raise
 
-                retry_postpone(
-                    job, tools.ustr(err.pgerror, errors="replace"), seconds=PG_RETRY
-                )
                 _logger.debug("%s OperationalError, postponed", job)
+                raise RetryableJobError(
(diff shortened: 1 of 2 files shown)


guewen:
There is an issue because the job has a `date_done` date set at the retry

guewen:
> There is an issue because the job has a `date_done` date set at the retry

Fixed by the last commit, not a new thing.

guewen:
However, the `raise` means we have a full stack trace every time, filling the logs. In the case of a retryable error, we don't want the exception stack everytime, only on the last one. I'll see if we can handle the rollback locally.

guewen:
@wpichler @agenterp I modified code to manually rollback / clear the env instead of raising, to prevent cluttering the logs with stack traces. Can you check? (same on #311)

StefanRijnhart (approved):

etobella (approved):

PR #313, opened 2021-02-10
[14][queue_job] Vacuum done jobs in batch

florian-dacosta:
I ran into an issue several times (on version 8 but no reason it will be different on version 14) where the AutoVacuum Job Queue cron fails due to timeout.
As a result it starts again, fail again, and have to be unblocked manually (by deleting some jobs by script for instance)

After some digging, it seems some people had this kind of issue also in past version : 
https://github.com/OCA/queue/pull/115
and was even merged in v7 : https://github.com/OCA/connector/pull/158

So I'd like to introduce the same in v14, commit every 1000 deleted jobs. (the 1000 is arbitrary, we could change it if needed, but I don't really see why it should be configurable.

I think it is important because depending on the project, the jobs may take more or less time to delete, for instance, if some mail.messages are added, it takes more time, if attachments are added (base_async_import), it takes more time, etc...
And depending how intensively jobs are used of course.

What do you think @guewen @rousseldenis @lmignon ?

### DIFF
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -286,11 +286,19 @@ class QueueJob(models.Model):
         """
         for channel in self.env["queue.job.channel"].search([]):
             deadline = datetime.now() - timedelta(days=int(channel.removal_interval))
-            jobs = self.search(
-                [("date_done", "<=", deadline), ("channel", "=", channel.complete_name)]
-            )
-            if jobs:
-                jobs.unlink()
+            while True:
+                jobs = self.search(
+                    [
+                        ("date_done", "<=", deadline),
+                        ("channel", "=", channel.complete_name),
+                    ],
+                    limit=1000,
+                )
+                if jobs:
+                    jobs.unlink()
+                    self.env.cr.commit()
+                else:
+                    break
         return True
 
     def requeue_stuck_jobs(self, enqueued_delta=5, started_delta=0):


queue_job/models/queue_job.py
```diff
@@ -284,13 +284,24 @@ def autovacuum(self):
...
         Called from a cron.
         """
-        for channel in self.env["queue.job.channel"].search([]):
-            deadline = datetime.now() - timedelta(days=int(channel.removal_interval))
-            jobs = self.search(
-                [("date_done", "<=", deadline), ("channel", "=", channel.complete_name)]
-            )
-            if jobs:
-                jobs.unlink()
+        with api.Environment.manage():
```
lmignon: @florian-dacosta Why do you need to use a new cursor?

guewen:
On the principle yes, on the implementation, as @lmignon noted, you shouldn't need a new cursor, and as the db snapshot can change after a commit, it may be better to do a search after every batch (with a limit).
You may want to have a look at #302 too which means there is less to delete on the vacuum.

florian-dacosta:
I thought it was always better to do explicit commit in new cursor.
Mainly in cron, I thought it would release the lock on the cron, I am convinced I saw this kind of thing a long time ago.
But wrong belief I guess, I have adapted the PR, thanks @lmignon @guewen 

>You may want to have a look at #302 too which means there is less to delete on the vacuum. 

Nice addition, I had not see it, but I've also done it in this v8 project, mainly for the gain at deletion indeed.
Good news it is in base module now though!

guewen:
>  Mainly in cron, I thought it would release the lock on the cron, I am convinced I saw this kind of thing a long time ago.

The cron are already executed in a second cursor

guewen:
Thanks to you :)

guewen (approved):

lmignon (approved):
Thank you @florian-dacosta. LGTM (code review)

etobella (approved):

PR #314, opened 2021-02-15
[13.0][IMP]Detect new databases or databases on which queue_job is in…

huakkai:
[13.0][IMP]Detect new databases or databases on which queue_job is installed after server start


huakkai:
@guewen Hello, can you help me check the PR? This PR is to detect new databases or databases on which queue_job is installed after server start.

PR #319, opened 2021-03-12
[12.0][FIX] queue_job: create/replace queue_job_notify function after migration

StefanRijnhart:
Port of https://github.com/OCA/queue/pull/304

PR #320, opened 2021-03-12
[10.0][FIX] queue_job: drop trigger when rewriting table in migration

StefanRijnhart:
On a large table, the migration query takes 23 seconds after dropping the
trigger (before that, it took 8 minutes).

```
                                                         QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------
 Update on queue_job  (cost=0.00..55301.11 rows=238578 width=1409) (actual time=22486.104..22486.105 rows=0 loops=1)
   ->  Seq Scan on queue_job  (cost=0.00..55301.11 rows=238578 width=1409) (actual time=0.015..415.132 rows=237130 loops=1)
 Planning Time: 0.080 ms
 Trigger queue_job_notify: time=457262.118 calls=237130
 Execution Time: 479831.502 ms
(5 rows)
```

### DIFF
--- a/queue_job/migrations/10.0.1.0.0/post-migration.py
+++ b/queue_job/migrations/10.0.1.0.0/post-migration.py
@@ -5,11 +5,13 @@
 
 def migrate(cr, version):
     """Extract from old `func_name` field the name of the method to be put
-    on `method_name`.
+    on `method_name`. Drop notify trigger for drastically better performance.
     """
     if not version:
         return
     cr.execute(
-        """UPDATE queue_job
+        """
+        DROP TRIGGER IF EXISTS queue_job_notify ON queue_job;
+        UPDATE queue_job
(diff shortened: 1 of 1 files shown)


guewen:
Is it re-created after this? Otherwise we could just use `DISABLE TRIGGER` and enable it again right after the update?

guewen (approved):
Answering to myself : it is created by the job runner.
Good catch BTW :)

StefanRijnhart:
Indeed in 10.0 it is reenabled automatically at runtime. It has been observed to be missing after migrations to 13.0, for which we have https://github.com/OCA/queue/pull/304 and https://github.com/OCA/queue/pull/319 (which make me think we should not have moved setting the trigger to `post_init_hook` at all, but to `_auto_init` instead).

ThomasBinsfeld (approved):

PR #323, opened 2021-03-19
[FIX] queue_job: Disappear trigger fixing

cmanuelalvarez11:


guewen:
Thanks for your PR.
Did it happen after a migration to 13.0? 
It was added here already: https://github.com/OCA/queue/pull/304/files
Maybe it would be better to move it to `_auto_init` then?

StefanRijnhart:
@ThomasBinsfeld said in https://github.com/OCA/queue/pull/304 that it wasn't necessary to do it in 13 if it was done in 12. But I agree that moving the creation/application of the trigger to `_auto_init` would solve this for once and for all.

cmanuelalvarez11:
Yes, it happens after a migration to 13.0, I think moving it to _auto_init
can solve the problem, or we could create the trigger directly in the
migration

Issue #306, opened 2021-02-02
Started job continues execution on requeue


guewen:
When a running job is requeued, the worker that started the execution continues to run, the job is set back to pending and start again. Then, the 2 workers have to update the state of the job to `done` and one has a concurrency error.

In the "Jobs Garbage Collector" cron, there is the possibility to define a delay after which started jobs are requeued automatically:
https://github.com/OCA/queue/blob/56b56f4e8e830e1bd63ea96313f2e4035584f382/queue_job/models/queue_job.py#L306-L312

This is very dangerous, as the job will be requeued in lead to this issue.

guewen:
I tried a quick experiment with:

```
    def requeue(self):
        for record in self.filtered(lambda record: record.state == "started"):
            os.kill(self.worker_pid, 1)
        self._change_job_state(PENDING)
        return True
```

Which works smoothly as long as...

* you use multiprocess and not multithread
* all the odoo processes are on the same node: if the worker that requeues is not on the same host as the worker executing the job, you kill the wrong pid
* after the restart of a host, the pid of a started but not running anymore job (force-killed) could be assigned to a different process now

 The first one could be acceptable, the 2 last ones, big nope

thomaspaulb:
_When a running job is requeued, the worker that started the execution continues to run, the job is set back to pending and start again._

Is that still true after https://github.com/OCA/queue/pull/131/files ? The job is set back to pending yes, but it won't start again as long as the old one is still running and thus holding the lock.

PR #324, opened 2021-03-23
[IMP] [14.0] store job execution data on execution

PieterPaulussen:
This adds functionality to store the name of the host on which the job is executed.
Also store the DB Backend PID as reference.

Primary motivation: Be able to kill either the Odoo process or the backend process in case of issues.

### DIFF
--- a/queue_job/__manifest__.py
+++ b/queue_job/__manifest__.py
@@ -18,6 +18,7 @@
         "views/queue_job_function_views.xml",
         "wizards/queue_jobs_to_done_views.xml",
         "wizards/queue_requeue_job_views.xml",
+        "wizards/queue_terminate_job_views.xml",
         "views/queue_job_menus.xml",
         "data/queue_data.xml",
         "data/queue_job_function_data.xml",
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -8,6 +8,7 @@ import os
 import sys
 import uuid
 from datetime import datetime, timedelta
+from socket import gethostname
 
 import odoo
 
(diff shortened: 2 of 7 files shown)


guewen:
Hi, I tried something a bit similar, but I was killing the python pid on the host, which turned into a russian roulette for processes as soon as you have more than one host. Acting on the postgres process is smarter. Did you encounter any shortfall that made you close your PR?

guewen:
BTW, here is a ref where such termination would be useful too: https://github.com/OCA/queue/issues/306

PieterPaulussen:
@guewen I closed this PR because the functionality will go through internal project review first, then we will propose this feature to the OCA in a more complete, tested fashion. I'll be opening it in the next couple of days. I'll give you a heads up when it's ready 😉

PieterPaulussen:
In regard to killing the python process, I ignored the issue for the same reason as you mention. Do you think we can come up with a solution if we take into account the worker hostname (also added in this PR)

guewen:
> In regard to killing the python process, I ignored the issue for the same reason as you mention. Do you think we can come up with a solution if we take into account the worker hostname (also added in this PR)

Hmm, as we can't choose which worker will receive the request and execute the method, it means if we hit the wrong host, it would involved propagating the termination on another host.

But actually, we shouldn't need to kill the pythons process as long as we terminate the postgres backend process. The python worker will receive a db disconnected error, and recover.

At this point, we should update the job's state to failed (with a message) or a new "terminated" state (I don't know if it's worth adding it?)

guewen:
Also killing a python pid when using multithreaded workers would not work

PR #325, opened 2021-03-23
[REF] Replace deprecated assertEquals with assertEqual

PieterPaulussen:
`assertEquals` has been deprecated https://docs.python.org/3/library/unittest.html#deprecated-aliases


guewen (approved):
Thank you!

marcelsavegnago (approved):

Issue #326, opened 2021-03-25
Hide job_queue_batch Systray in heavy load prodution systems


agent-z28:

agent-z28:
@guewen please leave your feedback :)

PR #327, opened 2021-03-25
11.0 add access right for batch systray

agent-z28:
Closes https://github.com/OCA/queue/issues/326


guewen (approved):
Thanks, I think it makes sense. It could be great to have a migration script to activate the group on users that currently have the batch user group, to avoid surprises?

PR #334, opened 2021-04-14
[14.0] [IMP] Store job execution information + wizard to terminate a job + test

PieterPaulussen:
Add the possibility to terminate a running job by terminating the backend process in postgres.

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -670,7 +670,7 @@ class Job(object):
 
     def get_db_txid(self):
         """
-        Get the current if of the database transaction the job is running in. Only store
+        Get the current ID of the database transaction the job is running in. Only store
         the 32bit representation variant in the database so it can be compared directly
         to the 'backend_xid' in pg_stat_activity.
         """
@@ -683,16 +683,17 @@ class Job(object):
         return self.env.cr.fetchone()[0]
 
     def set_db_txid(self):
-        """
+        """Store transaction ID on current job record.
+
         Use a new cursor to update the queue_job record of this job with the
         transaction ID the process is running in. The new cursor is necessary because
         after each commit the transaction ID changes (cfr. stored() is executed), which
         makes the transaction ID useless otherwise.
         """
-        tx_id = self.get_db_txid()
         db_record = self.db_record()
         if not db_record:
             return
+        tx_id = self.get_db_txid()
         with odoo.sql_db.db_connect(self.env.cr.dbname).cursor() as separate_cr:
             separate_cr.execute(
                 "UPDATE queue_job SET db_txid = %(tx_id)s WHERE id = %(rec_id)s;",
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -8,6 +8,7 @@ import os
 import sys
 import uuid
 from datetime import datetime, timedelta
+from socket import gethostname
 
 import odoo
 
@@ -241,6 +242,19 @@ class Job(object):
         be added to a channel if the existing job with the same key is not yet
         started or executed.
 
+    .. attribute::worker_pid
+
+        The process ID of the worker that is executing or has executed a Job.
+
+    ..attribute::worker_hostname
+
+        The name of the host where the worker is running. (Useful in scaled
+        environments to determine the node on which the process is executed.
+
+    ..attribute::db_txid
+
+        ID of the database transaction.
+
     """
 
     @classmethod
@@ -301,6 +315,8 @@ class Job(object):
             job_.company_id = stored.company_id.id
         job_.identity_key = stored.identity_key
         job_.worker_pid = stored.worker_pid
+        job_.worker_hostname = stored.worker_hostname
+        job_.db_txid = stored.db_txid
         return job_
 
     def job_record_with_same_identity_key(self):
@@ -494,6 +510,8 @@ class Job(object):
         self.eta = eta
         self.channel = channel
         self.worker_pid = None
+        self.worker_hostname = None
+        self.db_txid = None
 
     def perform(self):
         """Execute the job.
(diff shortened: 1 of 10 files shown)


PieterPaulussen:
@guewen Referring to the previous PR: https://github.com/OCA/queue/pull/324
We've reworked the solution with a better implementation (assisted by some feedback by @StefanRijnhart)

Feel free to review!

StefanRijnhart:
@PieterPaulussen did you install the pre-commit hook for this repo in your local clone?

PieterPaulussen:
No, was looking at it now

StefanRijnhart (approved):
As per internal review! (test/code review)

guewen (approved):
Wow, excellent work! Thank you @PieterPaulussen 

For the record, now that we have this feature:

* When using the "Requeue Jobs" wizard on a selection of jobs and a job is started, we should skip it, or terminate it before requeing, or add an option to choose (but requeue it blindly as of today is bad)
* The scheduled action that requeue stuck jobs should requeue "enqueued" jobs but should "terminate" started jobs after the timeout

Nothing to be done in this PR though.

queue_job/job.py
```diff
@@ -645,11 +668,45 @@ def eta(self, value):
         else:
             self._eta = value
 
+    def get_db_txid(self):
+        """
+        Get the current if of the database transaction the job is running in. Only store
```
simahawk:
```suggestion
        Get the current ID of the database transaction the job is running in. Only store
```

simahawk: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

queue_job/job.py
```diff
@@ -645,11 +668,45 @@ def eta(self, value):
...
+            self.env.cr.execute("SELECT (txid_current() % (2^32)::bigint)::text;")
+        return self.env.cr.fetchone()[0]
+
+    def set_db_txid(self):
+        """
+        Use a new cursor to update the queue_job record of this job with the
+        transaction ID the process is running in. The new cursor is necessary because
+        after each commit the transaction ID changes (cfr. stored() is executed), which
+        makes the transaction ID useless otherwise.
+        """
```
simahawk:
```suggestion
        """Store transaction ID on current job record.
        
        Use a new cursor to update the queue_job record of this job with the
        transaction ID the process is running in. The new cursor is necessary because
        after each commit the transaction ID changes (cfr. stored() is executed), which
        makes the transaction ID useless otherwise.
        """
```

queue_job/job.py
```diff
@@ -645,11 +668,45 @@ def eta(self, value):
...
+        tx_id = self.get_db_txid()
+        db_record = self.db_record()
+        if not db_record:
+            return
+        with odoo.sql_db.db_connect(self.env.cr.dbname).cursor() as separate_cr:
```
simahawk, line 697: the call to `get_db_txid` should happen before the with statement, no?

queue_job/wizards/queue_terminate_job.py
```diff
@@ -0,0 +1,42 @@
...
+        res = []
+        context = self.env.context
+        if context.get("active_model") == "queue.job" and context.get("active_ids"):
+            res = context["active_ids"]
+
+        # Filter out jobs not in 'started' state
+        res = (
+            self.env["queue.job"]
+            .browse(res)
+            .filtered(lambda qj: qj.state == "started")
```
simahawk: filter the one w/ out tx_id as well?

test_queue_job/models/test_models.py
```diff
@@ -44,6 +45,10 @@ def testing_method(self, *args, **kwargs):
             return self.env.context
         return args, kwargs
 
+    def test_long_running_method(self):
```
simahawk: what is this test for?

queue_job/job.py
PieterPaulussen: @simahawk That's what happens, is it not? Or did I miss something?
On line 692 we get the the transaction ID and then afterwards commit it with a different cursor inside the `with` statement

queue_job/wizards/queue_terminate_job.py
PieterPaulussen: Yes good suggestion!

test_queue_job/models/test_models.py
PieterPaulussen: Terminating a running job can only be tested if you have something that lasts long enough for a job to be queued and then started. Hence this "sleep trigger on the DB to have a long running DB transaction.
The reason I added this was for practical testing. An automated test would be nice of course, but would require a threaded test so the job can be started and then terminated by a different process. I did not have the time to spend on this to write this kind of test.

queue_job/job.py
simahawk: Sorry, poor wording... I mean: the call to `get_db_txid` is useless if you don't have a db_record.

test_queue_job/models/test_models.py
simahawk: Probably you can mock it instead. In this case I don't see the benefit if not let tests take longer than they should.

PieterPaulussen: @simahawk the tests do not take longer since this method is not tested, it is there for manual configuration. It is a sample to ensure the functionality can be tested.

queue_job/job.py
PieterPaulussen: Ah yes, thanks for the clarification. I'll switch the order.

PieterPaulussen: Updated!

test_queue_job/models/test_models.py
simahawk: ah I see... method name cheated me :sweat_smile: 
Ok, then I see no reason to keep it. If you really want to keep it please add an extended comment explaining the purpose. Also, rename it to `testing_*`.

PieterPaulussen: Thanks for the improvements, I really appreciate it!
I do prefer to keep this method so I updated the method name and added the docstring.

simahawk (approved):

PR #335, opened 2021-04-14
[WIP] [FIX] [14.0] Include 'test_tags' attribute on DocTests so they are se…

PieterPaulussen:
…lected during standard Odoo run


StefanRijnhart:
Ah look @PieterPaulussen it was solved in 13 with https://github.com/OCA/queue/pull/296

PieterPaulussen:
@StefanRijnhart Thanks! Created a port of the commit to 14.0 here https://github.com/OCA/queue/pull/336

PieterPaulussen:
@StefanRijnhart After careful consideration, I think this approach is still the best one I've currently been able to devise.

PR #344, opened 2021-05-11
[12.0][FIX] missing attachment type

sergiocorato:
When importing `product.template` with `Import in the background` set, `ir.attachment` `type` field is guessed? as `product`, which is invalid.

Screenshot(10)


guewen (approved):
Hi thanks!

Here is a proposition of commit message if you'd like:  https://github.com/OCA/queue/pull/242#issuecomment-650050653 ?

sergiocorato:
@guewen done, thanks!

Issue #343, opened 2021-05-06
[14.0] Can't set job to done. ERROR: could not serialize access due to concurrent update


lgarcia11:
Job runs successfully but fails when changing status to done. I think it's related to PR #334, before it works fine. 

`2021-05-06 18:26:42,282 25963 INFO odoo-14 werkzeug: 127.0.0.1 - - [06/May/2021 18:26:42] "GET /queue_job/runjob?db=odoo-14&job_uuid=1f99252d-b0f2-4ab0-a661-d4c8a6264420 HTTP/1.1" 200 - 13 0.027 0.035
2021-05-06 18:26:44,506 25963 INFO ? odoo.addons.queue_job.jobrunner.runner: asking Odoo to run job 7ec61c9a-e8a7-4fdf-b7a3-4729f68fc690 on db odoo-14 
2021-05-06 18:26:44,568 25963 ERROR odoo-14 odoo.sql_db: bad query: UPDATE "queue_job" SET "company_id"=1,"date_done"='2021-05-06 18:26:44.562230',"date_enqueued"='2021-05-06 18:26:44',"date_started"='2021-05-06 18:26:44.542526',"db_txid"='94823',"eta"='2021-05-06 18:26:44.502335',"exc_info"=NULL,"identity_key"=NULL,"max_retries"=5,"priority"=10,"result"=NULL,"retry"=11,"state"='done',"worker_hostname"='a0162366f236',"worker_pid"=25963 WHERE id IN (22)
ERROR: could not serialize access due to concurrent update`

omalbastin:
I do have the same error in 14.0

fuentes73:
We're having the same issue, after roll back to 221871460d09fa93a3f96f8be30e2d2c6303c334 works fine

florian-dacosta:
Hello
Same here!
Also work fine if roll back right before https://github.com/OCA/queue/pull/334

@PieterPaulussen Any idea?

guewen:
The feature is reverted, it should be resolved on the latest commits.

PR #345, opened 2021-05-11
Revert "[14.0] [IMP] Store job execution information + wizard to terminate a job + test"

guewen:
Reverts OCA/queue#334

Hi @PieterPaulussen, we have a regression with this one, see #343.
Unless you have the possilibity to fix it readily, I'll revert your PR until we have a proper fix.
I suppose the problem happens as we write on the job here https://github.com/OCA/queue/pull/334/files#diff-23034efd8360efc0b59b169eb4834058fd91f6eabf06cf8a04bac0142e3cc385R697-R702 and from the controller in concurrent transactions


StefanRijnhart (approved):

StefanRijnhart:
Fasttracking
/ocabot merge minor

PR #346, opened 2021-05-14
[14.0] [ADD] Support multi-nodes with lock on jobrunner

PieterPaulussen:
Forward port of https://github.com/OCA/queue/pull/256


StefanRijnhart:
https://github.com/OCA/queue/pull/256/commits/4137c87051aa7917bb35d711cc10025b532b2837 is missing in this PR, and I believe instead there is an empty (duplicate) commit

Issue #342, opened 2021-05-06
RFC: mark jobs as cancelled


StefanRijnhart:
In Odoo 8.0, we occassionaly use the option to set jobs to inactive + failed to mark jobs as (temporarily) cancelled. We do this to manage unexpected concurrency problems or performance problems. We reactivate the jobs and reset to pending after office hours or when an underlying condition is fixed.

I've only just noticed that in later versions, this is no longer possible because the `active` flag was removed here https://github.com/OCA/connector/pull/198 because it was obsolete for its original purpose. I would either like to reintroduce the flag, or to introduce a dedicated `cancelled` state.

cc: @PieterPaulussen

PieterPaulussen:
@StefanRijnhart Don't you agree that we misused the failed-active combination in our setup? I would add a new `cancelled` state rather than re-add the original columns.

StefanRijnhart:
Well, I am neutral to the solution, but I did mention having a distinct cancelled state as one of the options.

etobella:
Well, in this case, I would use something like `draft` or `blocked`, as the job is not `cancelled` it is just waiting for some conditions

StefanRijnhart:
That's our use case, but if the job is never restarted it is actually `cancelled`. Currently we set jobs to done manually for that, but that is a bit misleading as well.

StefanRijnhart:
So I imagine
* a list action 'Cancel jobs', applied on pending/failed jobs from the selection
* the list action 'Requeue jobs' to work on cancelled jobs as well as on jobs in other states
* Button 'Set Done' on failed jobs replaced by 'Cancel job'
* Cancelled jobs vacuumed just like Done jobs are.

guewen:
Hi, 

IMO `active` is just too risky / confusing as jobs are out of visibility (BTW I'm not very happy with the default security rules on `company_id` for the same reason). I think jobs should never be hidden by default.

As for a cancelled state, the discussion popped in several times and it was unclear if it was really needed vs set to done and the simple option won.
But I reckon the description of your requirements makes sense, and could be a nice addition.

hbrunn:
so I'll go ahead and implement Stefan's bulletpoints

StefanRijnhart:
@hbrunn great! As always, it was a quick write-up that may not be entirely correct or exhaustive. Let me know if you have any doubts or new insights.

hbrunn:
https://github.com/OCA/queue/pull/347

PR #347, opened 2021-05-14
[IMP] queue_job: Add cancelled state to queue.job

hbrunn:
I'll add tests when we agree this is how it should work

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -19,6 +19,7 @@ from .exception import (NoSuchJobError,
 
 PENDING = 'pending'
 ENQUEUED = 'enqueued'
+CANCELLED = 'cancelled'
 DONE = 'done'
 STARTED = 'started'
 FAILED = 'failed'
@@ -27,6 +28,7 @@ STATES = [(PENDING, 'Pending'),
           (ENQUEUED, 'Enqueued'),
           (STARTED, 'Started'),
           (DONE, 'Done'),
+          (CANCELLED, 'Cancelled'),
           (FAILED, 'Failed')]
 
 DEFAULT_PRIORITY = 10  # used by the PriorityQueue to sort the jobs
@@ -283,6 +285,9 @@ class Job(object):
         if stored.date_done:
             job_.date_done = stored.date_done
 
+        if stored.date_cancelled:
+            job_.date_cancelled = stored.date_cancelled
+
         job_.state = stored.state
         job_.result = stored.result if stored.result else None
         job_.exc_info = stored.exc_info if stored.exc_info else None
@@ -435,6 +440,7 @@ class Job(object):
         self.date_enqueued = None
         self.date_started = None
         self.date_done = None
+        self.date_cancelled = None
 
         self.result = None
         self.exc_name = None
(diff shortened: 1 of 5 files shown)


queue_job/job.py
```diff
@@ -616,6 +626,13 @@ def set_done(self, result=None):
         if result is not None:
             self.result = result
 
+    def set_cancelled(self, result=None):
+        self.state = CANCELLED
+        self.exc_info = None
```
guewen: Do we want to empty the exc info? It could maybe be useful to have a trace of why it was cancelled. Not sure about this.

guewen:
> I'll add tests when we agree this is how it should work

It looks good to me :)

queue_job/job.py
hbrunn: good point, I think I agree. What do others think?

simahawk: seems reasonable to keep it.

StefanRijnhart: Yes, let's try to hold on to the exc_info as long as possible. On some of my projects we post the exc_info as a comment on the job to preserve it even after a successful retry. (that's just by means of illustration, not a hint to take up here).

queue_job/views/queue_job_views.xml
```diff
@@ -15,19 +15,25 @@
                         type="object"
                         groups="queue_job.group_queue_job_manager"/>
                     <button name="button_done"
-                        states="pending,enqueued,failed"
+                        states="pending,enqueued"
```
StefanRijnhart: Depending on the meaning of 'Done' (I'd say, has been executed manually outside of the queue and perhaps even outside of Odoo), you might still want to set 'failed' and even 'cancelled' jobs to done. Or do you have another interpretation?

hbrunn: I followed point 3 from https://github.com/OCA/queue/issues/342#issuecomment-833548686, I personally would show the button everywhere

StefanRijnhart: Ah, I made that suggestion with the idea to replace the 'Set Done' button entirely with a 'Set Cancelled' button, which is probably not a good idea in a stable version or not at all depending on the semantics so please show the button everywhere.

hbrunn:
thanks all, changed according to comments

queue_job/models/queue_job.py
```diff
@@ -370,6 +381,18 @@ def set_done(self):
...
 
 
+class SetJobsToCancelled(models.TransientModel):
+    _inherit = 'queue.requeue.job'
+    _name = 'queue.jobs.to.cancelled'
+    _description = 'Cancel all selected jobs'
+
+    @api.multi
+    def set_cancelled(self):
+        jobs = self.job_ids.filtered(lambda x: x.state in ('pending', 'failed'))
```
StefanRijnhart: Can you add that corner case `enqueued` here as it's listed as a valid state for `button_cancelled` below.

StefanRijnhart:
Were you still going to add a test?

hbrunn:
now with the promised tests. Anybody have permissions to fix runbot?

StefanRijnhart (approved):
Thanks!

guewen (approved):
Thanks!
About runbot, I don't.

dreispt:
@hbrunn Can you please resolve the conflicts? Thanks!

hbrunn:
@dreispt thanks for the heads up!

PR #349, opened 2021-05-18
[14.0] [IMP] overcapacity

PieterPaulussen:


queue_job/jobrunner/channels.py
```diff
@@ -474,7 +474,7 @@ def set_done(self, job):
         This removes it from the channel queue.
         """
         self.remove(job)
-        _logger.debug("job %s marked done in channel %s", job.uuid, self)
+        _logger.info("job %s marked done in channel %s", job.uuid, self)
```
guewen: Are you sure about this change, it doesn't match your commit description :)

> stop jobrunner channels' log flooding

guewen (commented):
Thanks, very interesting proposal. I understand it optimizes the usage of the workers as it will try to max out the usage of the workers if some channels have nothing to execute and other channels have too much.

I have one concern with this though: one of the issue the channels solve is to be able to constrains some jobs to be executed in a given capacity. Examples:

* Execute jobs one by one to deal with concurrency issues (you covered the `sequential` case, but we don't always want "one at a time" jobs to be sequential)
* Ensure long running jobs do not fill all the available slots

I may have misunderstood the new behavior, but I think in both cases, the jobs would be brought in the `root` channel and could be executed concurrently / fill up the slots.

Maybe we can add an option in the channels configuration to disable the overcapacity handling for some channels, or the other way around, allow some channels to be maxed out in `root` (the first one is risky for existing deployment, the second one is a bit cumbersome...).

Thoughts?

PieterPaulussen:
@guewen Thanks for the feedback.

> Execute jobs one by one to deal with concurrency issues (you covered the sequential case, but we don't always want "one at a time" jobs to be sequential)

Could you give me an example how and why you would want this behavior?
We will be proposing a sequence grouping mechanism in the near future that allows to sequentially process a number of jobs in a parallel channel. Is this what you are suggesting?

> Ensure long running jobs do not fill all the available slots

You could apply throttling in combination with job priority to solve tackle this, no? On the project @StefanRijnhart and I work on, we applied a weighted number determining the average load of the job function. That way we could tune down the number of jobs that were started if the number of heavy jobs becomes to high. Additionally, we measured the load on the database itself so we would always preserve some database CPU capacity for regular Odoo operations. I think we found the solution to be project specific so it was not proposed back to the OCA. @StefanRijnhart what's your opinion on the matter?

PS: Most of the PR's with improvements to the queue module are proposed by me, but most of the credits must go to @StefanRijnhart for designing the solutions!

queue_job/jobrunner/channels.py
PieterPaulussen: Indeed, there's a commit that should not be there

StefanRijnhart:
@PieterPaulussen it's all team effort!

I have to say that I'm a little taken back by the two regressions that we have caused so far by porting the features from our project to the 14.0 OCA modules. Our project runs on an older version but we are preparing for a migration in the distant future. But that means that this code is not actually tested on a proper setup, so that these regressions go unnoticed. And this PR looks like one of the more risky ones with functional as well as technical impact so I think we should put this one to work-in-progress until we are doing proper user acceptance of exactly this code on our project on production-like data.

Having that said, Pieter mentioned the sequence group feature and if you find that interesting you can find our internal PR of the port of this feature here: https://github.com/EssentNovaTeam/queue/pull/14/files.

I hope we will be able to come back and revisit this one but I'm not going to hold my breath to be honest.

guewen:
> Could you give me an example how and why you would want this behavior? 

Of course :)

You may have jobs which always raise concurrent update errors when executed concurrently (for instance because they all have to pick a number in a sequence).
In some cases, I don't care about the order, so I wouldn't use "sequential", which would stop the queue as soon as a a job fails, but I still would like to be sure they are execute one at a time, otherwise it is a waste of resources.

As for the long-running jobs I was mentioning, I don't think the priority helps. The priority will help to execute the fast-pacing jobs first, then if I have 10 pending jobs with an avg duration of 15 minutes, they may take 10 slots and block the queue for 15 minutes. But then I'm not sure how the throttle works with your proposal, maybe you already cover this use case.

@StefanRijnhart I can't blame you. I have some PRs that I'm not comfortable merging because they lack tests in the wild or at least feedbacks. I don't really know how to deal with that and that's a bummer.

PR #355, opened 2021-06-11
[13.0][FIX] queue_job_cron: channel_id must be storable.

LoisRForgeFlow:
Otherwise, you cannot use any channel other than default (
root.ir_cron)

The struggle was real without this fix :cry:  :
Peek 2021-06-11 16-13

After the fix life is great :smile:  :
Peek 2021-06-11 16-19

@ForgeFlow

### DIFF
--- a/queue_job_cron/models/ir_cron.py
+++ b/queue_job_cron/models/ir_cron.py
@@ -17,13 +17,16 @@ class IrCron(models.Model):
         comodel_name="queue.job.channel",
         compute="_compute_run_as_queue_job",
         readonly=False,
+        store=True,
         string="Channel",
     )
 
     @api.depends("run_as_queue_job")
     def _compute_run_as_queue_job(self):
         for cron in self:
-            if cron.run_as_queue_job and not cron.channel_id:
+            if cron.channel_id:
+                continue
+            if cron.run_as_queue_job:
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
:+1: makes sense

JordiMForgeFlow (approved):

AdriaGForgeFlow (approved):
LGTM :+1:

LoisRForgeFlow:
@guewen Do you think this is enough?, or it is better to use an extra field `manual_channel_id` and rely on that one for the compute ensuring that it won't get back to False if the compute is triggered.

etobella (approved):

guewen:
> @guewen Do you think this is enough?, or it is better to use an extra field `manual_channel_id` and rely on that one for the compute ensuring that it won't get back to False if the compute is triggered.

With the new compute methods of 13.0, we can make it work like a default value: 

```
for cron in self:
    if cron.channel_id:
        continue
    if cron.run_as_queue_job:
        cron.channel_id = ...
```

That being said, I'm not sure why this field has to be computed.

LoisRForgeFlow:
@guewen I added the continue clause, I think that makes sense. For me this is good to merge.

guewen (approved):

PR #356, opened 2021-06-22
[FIX] queue_job: Migrations raising errors

etobella:
When executing OpenUpgrade with a database that didn't have queue_job but a module change on 13.0 required it, several errors where raised (I don't know why, but OpenUpgrade executed the migrations). In order to solve them, I had to change the scripts applying this changes

### DIFF
--- a/queue_job/migrations/13.0.3.2.0/pre-migration.py
+++ b/queue_job/migrations/13.0.3.2.0/pre-migration.py
@@ -2,12 +2,14 @@
 
 import logging
 
-from odoo.tools.sql import column_exists
+from odoo.tools.sql import column_exists, table_exists
 
 _logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


queue_job/migrations/13.0.3.2.0/pre-migration.py
```diff
@@ -8,6 +8,8 @@
 
 
 def migrate(cr, version):
+    if not column_exists(cr, "queue_job", "record_ids"):
```
pedrobaeza: I don't think this column will exist ever. It seems a one2many/many2many field.

etobella: Sure, the problem is that the table does not exist, modified checking the existance of the table

pedrobaeza (approved):

guewen (approved):

PR #358, opened 2021-07-01
[IMP] no need for base_sparse_field anymore since 0f5ec1

nilshamerlinck:
- Since https://github.com/oca/queue/commit/0f5ec1 `Serialized` fields have been replaced by `JobSerialized` fields
- No need for forward ports, as this dependency has been removed already in further versions


guewen (approved):
You are right, thanks!

Issue #166, opened 2019-08-06
[RFC] Pause a channel from UI


guewen:
I have been in this situation a few times when jobs of a kind go crazy and you want to pause them to keep some place for the other jobs while you resolve the issue. Things that may be nice:

- Setting 0 in the configuration stops a channel from processing
- A button to pause a channel from the UI

PR #361, opened 2021-07-16
[14.0][REF] queue_job: enable state pause in jobs

fernandahf:
Description of the issue/feature this PR addresses:

PR enables state `paused` in jobs, it doesn't allow execute job while job is that state, even if `eta` is in past.

Current behavior before PR:

Desired behavior after PR is merged:

-- I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -18,6 +18,7 @@ ENQUEUED = "enqueued"
 DONE = "done"
 STARTED = "started"
 FAILED = "failed"
+PAUSED = "paused"
 
 STATES = [
     (PENDING, "Pending"),
@@ -25,6 +26,7 @@ STATES = [
     (STARTED, "Started"),
     (DONE, "Done"),
     (FAILED, "Failed"),
+    (PAUSED, "Paused"),
 ]
 
 DEFAULT_PRIORITY = 10  # used by the PriorityQueue to sort the jobs
@@ -677,6 +679,9 @@ class Job(object):
         self.state = FAILED
         if exc_info is not None:
             self.exc_info = exc_info
+
+    def set_paused(self):
+        self.state = PAUSED
 
     def __repr__(self):
         return "<Job %s, priority:%d>" % (self.uuid, self.priority)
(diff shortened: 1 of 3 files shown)


fernandahf:
@moylop260

Could you review this, please?

moylop260:
@guewen 

What do you think about this extra status?

moylop260:
@sbidoul @lmignon

What do you think?

guewen:
Thanks for the proposal. Wouldn't it be more practical to be able to pause a channel? (https://github.com/OCA/queue/issues/166)

From my experience, when you need to pause jobs, you pretty often want to pause all the jobs of the same job function, including those that will be created later. Maybe with a wizard to quickly change the channel on a selection of jobs.

Which means we could:

* Have a (or several) channel which is paused from startup configuration (0 capacity): one can move jobs in this channel, they will not be executed as it has no slot. We don't need a new state on the jobs for this, they stay in pending but are not processed. This should be easy to implement as it would only require to change the jobrunner to hold jobs when the capacity is 0 (+ a wizard to change the channel in batch). For users, the UX is either to change the channels of jobs you want to pause, either to change it on the job function if you want future jobs to be paused as well.
* More advanced: allow to dynamically pause any channel from the UI: much less easy as we would need a way to propagate the information to the jobrunner (maybe a special notify command), but it would allow to pause a channel without changing the channel of the job function. Maybe overkill?

Does it make sense?

moylop260:
@guewen 

Thank you for reply

> From my experience, when you need to pause jobs, you pretty often want to pause all the jobs of the same job function, including those that will be created later. Maybe with a wizard to quickly change the channel on a selection of jobs.

In this case, we need to pause only a job since that it is used to schedule a process for a future date but is needed to filter only the paused ones.

Using a wizard to change the job of channel was our second option (but now our first one option after your feedback thank you!)

So, we will do it in that way.

1) Channel called `paused` (No jobs assigned)
2) Wizard to change job of channel

Is it wizard valid for OCA project or do you prefer to use this wizard in our custom modules?

guewen:
👍
Fine with me in the core module.

PR #364, opened 2021-07-26
[IMP] queue_job: allow to change jobs to paused channel

### LINKED ISSUE #166: [RFC] Pause a channel from UI

fernandahf:
Description of the issue/feature this PR addresses:

In some cases, we need to pause only a job since that it is used to schedule a process for a future date but is needed to filter
only the paused ones.

This MR allows to change a job channel to paused channel (capacity equals zero) using a wizard.

Current behavior before PR:

No feature to pause channel, all jobs are executed according to eta and channel capacity.

Desired behavior after PR is merged:

Wizard available to change job channel to pause channel (default) no matter `eta`, wizard checks if job are in pause channel and returns them to original channel, `pause` channel is setup with capacity equals zero, if it is setup with other capacity will raise an error.

Fix https://github.com/OCA/queue/issues/166

-- I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

### DIFF
--- a/queue_job/__manifest__.py
+++ b/queue_job/__manifest__.py
@@ -18,6 +18,7 @@
         "views/queue_job_function_views.xml",
         "wizards/queue_jobs_to_done_views.xml",
         "wizards/queue_requeue_job_views.xml",
+        "wizards/queue_jobs_pause_channel_views.xml",
         "views/queue_job_menus.xml",
         "data/queue_data.xml",
         "data/queue_job_function_data.xml",
--- a/queue_job/data/queue_data.xml
+++ b/queue_job/data/queue_data.xml
@@ -33,5 +33,9 @@
         <record model="queue.job.channel" id="channel_root">
             <field name="name">root</field>
         </record>
+        <record model="queue.job.channel" id="channel_pause">
+            <field name="name">pause</field>
+            <field name="parent_id" ref="channel_root" />
+        </record>
     </data>
 </odoo>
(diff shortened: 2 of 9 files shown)


moylop260:
@guewen 

What do you think?

guewen:
Thanks for the proposal!

A question for my understanding, as I see you propose a default `root.pause` channel which is a nice idea for UX, did you check if we are able to pause another channel by setting a capacity of 0 in the jobrunner's config? 

E.g. if I put 
* `root:4,root.foo:2,root.bar:0`, is `root.bar` paused?
* `root:4,root.foo:2,root.bar:0,root.bar.baz:2` are both `root.bar` and `root.baz` paused?

fernandahf:
@guewen 

> did you check if we are able to pause another channel by setting a capacity of 0 in the jobrunner's config?

We are not able to pause another channel by setting because we are validating the channel's name to consider the capacity of 0, otherwise, the channel has a capacity of 1 by default:

https://github.com/OCA/queue/pull/364/files#diff-f9cfc30dd392ede9ca7f2a01b30fd49c73e5e872a022dd97ad473a70d3f4dc26R909-R912

cc @moylop260

fernandahf:
Hi @guewen 

Is there something pending about our side about this?

Issue #363, opened 2021-07-22
[13.0] [14.0] company_id should be preserved when a delayed method is executed


lmignon:
When a job is cretated, the current company [is stored into the job info](https://github.com/OCA/queue/blob/13.0/queue_job/job.py#L493).  Nevertheless, when a job is performed, it seems that the company is not restored into the running context...
Do I miss something or should we modify the [`func`](https://github.com/OCA/queue/blob/13.0/queue_job/job.py#L630) method to call the with_company method on the recordset before the call o the dlayed method? 
```python
    def func(self):
        recordset = self.recordset.with_context(job_uuid=self.uuid).with_company(self.company_id)
        return getattr(recordset, self.method_name)
```

lmignon:
ping @simahawk @guewen @Cedric-Pigeon

guewen:
I think the initial goal to have the company was to have a security rule based on it (https://github.com/OCA/connector/commit/bff6ca3983d660931cd79362d6950e1472e3d0d0). I'd prefer not to have this security rule by default BTW, this is the first thing I used to disable on any project (or forget to do it then shoot myself in the foot because I don't see all the jobs :D), but I digress.

Indeed, we should keep the company. It seems to be a great opportunity to store the context in the `JobSerialized` records field (https://github.com/OCA/queue/issues/283). We could start with a simple allow-list of key to keep, including the various company context keys.

Cedric-Pigeon:
I agree with @guewen. Limiting the view of jobs per companies is a nightmare and it is also the first thing I disable.
About the company used for the job itself I think it must be related to the user running the job.

lmignon:
The company used to run the job must be the one the active one at the time the job has been scheduled. Otherwise, we've weird issues with multi company users if they change the active company before a scheduled job is performed.

Cedric-Pigeon:
@lmignon yes you are right

pedrobaeza:
Since v13, the company is not managed through context, so IMO the job should recreate the environment with the main company and the rest of them displayed the same as when you perform the enqueuing.

lmignon:
@pedrobaeza The fields `company` and  `companies` on the env are both initialized from the `allowed_company_ids` values (https://github.com/odoo/odoo/blob/14.0/odoo/api.py#L555) . Therefore, if we store and restore the context,  we don't need to explicitly call  `with_company` and we are sure that the job will be executed on the right company. Therefore the @guewen 's proposal seems to be the right way to go..

[...]

Issue #283, opened 2020-11-13
[feature] Storing context in recordsets


guewen:
The context is currently not kept, neither for the "main" record on which the job works, neither on the records passed in args or kwargs.

Some hints for the implementation

* https://github.com/OCA/queue/pull/121 proposed an implementation in 11.0 with an allow list
* https://github.com/OCA/queue/pull/281 refactors the "main" record: the context should now be stored in the json-serialized recordsets (on records, args, kwargs) (we don't need new fields)
* #274 could provide a way to customize the allowed list of keys in the context? (https://github.com/OCA/queue/pull/121#issuecomment-719381455)

* Add a context key in the json serializer/deserializer https://github.com/OCA/queue/blob/13.0/queue_job/fields.py
* Pass a method to the `JobSerialized` field which does the filtering of context keys `records = fields.JobSerialized(..., allow_list="_records_context_allow_list")`

Other questions:

* should we store the context in any case?
* what should be the basis for the allow list? At least: company fields (allowed_company_ids, force_company, ... depends of the version), lang and tz?

PR #365, opened 2021-08-02
[13.0][IMP] queue_job current company

acsonefho:
I currently have some trouble to trigger `job` because they use the default company of the user instead of the active one.
I see this open discussion: https://github.com/OCA/queue/issues/363 but IMO using a serialized context is not in opposition with this solution as the key `allowed_company_ids` is not always in the context (only if the user switch company?).

I did 2 commits (IMO it's important to keep them):
1) Force companies into the context (via `allowed_company_ids` key)
2) Only load the job's company into `allowed_company_ids` because it could load too many unexpected records (during search).

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -628,7 +628,15 @@ class Job(object):
 
     @property
     def func(self):
-        recordset = self.recordset.with_context(job_uuid=self.uuid)
+        user = self.env["res.users"].browse(self.user_id)
+        company_ids = user.company_ids.ids
+        # Insert the current company in top position
+        company_ids.insert(0, self.company_id)
+        # Remove duplicates but keep order
+        company_ids = list(dict.fromkeys(company_ids))
+        recordset = self.recordset.with_context(
+            job_uuid=self.uuid, allowed_company_ids=company_ids
+        )
         return getattr(recordset, self.method_name)
 
     @property
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -628,12 +628,13 @@ class Job(object):
 
     @property
     def func(self):
-        user = self.env["res.users"].browse(self.user_id)
-        company_ids = user.company_ids.ids
-        # Insert the current company in top position
-        company_ids.insert(0, self.company_id)
-        # Remove duplicates but keep order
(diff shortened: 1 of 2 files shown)


Cedric-Pigeon (approved):

sbidoul:
Shouldn't we rather design the jobs to work correctly irrespective of the user and job company_id ?

guewen:
Hi, maybe I don't get the use case properly, but I'm not sure to agree with:

>  using a serialized context is not in opposition with this solution as the key allowed_company_ids is not always in the context 

If the job was called synchronously like a normal method, and the `allowed_company_ids` was not in the context, you would have the same issue then? So it means the issue is on the caller not setting it. I don't see why calling asynchronously would need to have a different behavior? And if the caller did set `allowed_company_ids` then if we store the context, we'll have it automatically.

acsonefho:
> Hi, maybe I don't get the use case properly, but I'm not sure to agree with:
> 
> > using a serialized context is not in opposition with this solution as the key allowed_company_ids is not always in the context
> 
> If the job was called synchronously like a normal method, and the `allowed_company_ids` was not in the context, you would have the same issue then? So it means the issue is on the caller not setting it. I don't see why calling asynchronously would need to have a different behavior? And if the caller did set `allowed_company_ids` then if we store the context, we'll have it automatically.

IMO if the `allowed_company_ids` is not into the context, so the default company of the user is used. So `job.company_id == self.env.user.company_id == self.env.company` is `True` so no issue.
But if the user switch company to trigger a job, `job.company_id == self.env.user.company_id == self.env.company` become `False`. But `job.company_id == self.env.company` still `True`. It's why it's required to keep the current company to trigger the job.

I totally agree with the option to store the `context` into the job. That fix this company issue. But for now as nobody agree to store it

guewen:
> But for now as nobody agree to store it

I do agree to store it (in the recordset as discussed in https://github.com/OCA/queue/pull/281#discussion_r522958962), I proposed it here https://github.com/OCA/queue/issues/283. I'm sure other people agree too, the variances of opinions may be more in what exactly in the context we allow to store or not, but we can start small with an allow-list with the standard keys.
It's just that nobody worked on it.

PR #351, opened 2021-05-28
[14.0][MIG] Migration of base_import_async

BT-jcamacho:
Migration of base_import_async from 13.0 to 14.0

PR #366, opened 2021-08-10
[MIG] base_import_async: Migration to 14.0

simahawk:
take over and fix #351.

Contains a small improvement on tests.


rvalyi (approved):
Thanks!

simahawk:
the builds are all green, not sure why is not update here. Since this is the same as #351 that was approved, I merge.

/ocabot merge patch

PR #373, opened 2021-09-14
[13.0] Vacuum done jobs in batch

mlaitinen:
Huge amount of jobs to delete may take a lot of time and the cron may crash, because of timeout for instance
This issue would make the cron permanently run and permanently failing... Deleting batch allow to avoid this issue
if the cron fails, it still will have delete some of the job history

This is just a cherry-pick from #313. The same problem affects `13.0` as well.


etobella (approved):

guewen (approved):
Thanks !

Issue #58, opened 2018-02-21
[Security] initialize_databases require list_db=True


loctv:
In a use case with several databases, at initial database step call "odoo.service.db.exp_list(True)" to get all explicit db names. This was only accepted with config list_db=True. But I want to prevent listing databases entirely and use db_filter option.  

```
# queue_job/runner.py
def get_db_names(self):
        if odoo.tools.config['db_name']:
            db_names = odoo.tools.config['db_name'].split(',')
        else:
            db_names = odoo.service.db.exp_list(True)
        return db_names
```

```
# odoo/service/db.py
def exp_list(document=False):
    if not odoo.tools.config['list_db']:
        raise odoo.exceptions.AccessDenied()
    return list_dbs()
```

guewen:
Hi,

You have to set the `db_name` value (or `--database` in CLI) to the list of databases you want to use (separated by commas). This list of databases will be used by the cron and by the queue job.
`db_filter` is an option for the front-end.

eLBati:
@guewen @loctv please have a look at https://github.com/OCA/queue/pull/379

NightTsarina:
Hi, this issue has been solved for v11 (#665), v12 (#379), v15 (#580), and v16 (#589); but two proposed fixes for v14 (#416 and #588) were not merged. Is there any chance this will be fixed for v14?

PR #379, opened 2021-10-26
[12.0] FIX queue_job avoiding AccessDenied when loading the module in an instance with list_db = False

eLBati:
The scenario is a multi database environment where you can't set db_name as the databases change frequently and you use dbfilter to allow users to access their DB according to domain

Also see https://github.com/OCA/queue/issues/58


sbidoul (approved):
Given that exp_list looks like this:

image

This change looks reasonable to me.

@guewen wdyt ?

guewen (approved):

PR #387, opened 2021-11-03
[14.0][IMP][FWP] queue_job current company

acsonefho:
Forward port of https://github.com/OCA/queue/pull/365
Use the current company to trigger the job (+ add related tests)

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -586,7 +586,13 @@ class Job(object):
 
     @property
     def func(self):
+        # We can fill only one company into allowed_company_ids.
+        # Because if you have many, you can have unexpected records due to ir.rule.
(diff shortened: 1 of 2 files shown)


acsonefho:
@rousseldenis

Cedric-Pigeon (approved):

rousseldenis (approved):

guewen:
Hi, same remarks as on #365. Maybe you could answer on the latter ?

sbidoul:
Hm, I'm not sure it was right to merge this. The proper solution is the queue job context propagation which we merged in #432. @guewen what do you think?

guewen:
Indeed, it shouldn't have been merged

PR #393, opened 2021-12-03
[14.0][MIG] queue_job_batch, test_queue_job_batch: Migration to 14.0

anothingguy:


leemannd (approved):

jordancoest (approved):

olaf-wagner:
Everything seems to be ok. Can anybody merge it?

jordancoest:
> Everything seems to be ok. Can anybody merge it?

We are planning to use this PR. It seems ok for us too.
Is it ok to be merged ?

simahawk:
@guewen @etobella wondering: I guess we'll able to avoid using batches once #154 reaches v14, no?

simahawk:
> @guewen @etobella wondering: I guess we'll able to avoid using batches once #154 reaches v14, no?

In any case we can merge this meanwhile, unless you see any incompatibility.

guewen:
> In any case we can merge this meanwhile, unless you see any incompatibility.

I ensured dependencies and batches are not incompatible, fine with me. 

> wondering: I guess we'll able to avoid using batches once #154 reaches v14, no?

I checked some time ago, I think both can be used a bit differently, in a batch of jobs, you can add jobs after having created the batch record, whereas with dependencies, you have to build the group with the API, then enqueue them all at once. I'm not sure if batches are used to accumulate jobs created from different requests/transactions, that would be something you can't do with dependencies.

guewen (approved):

PR #394, opened 2021-12-06
[FIX] make autovacuum work in blocks of 1000, in order to avoid memory

gfcapalbo:
backport of  https://github.com/OCA/queue/commit/085d5ba822fe86e7e3f5d17169f7198d87d8eef6
to v10.

If for any reason autovacuum fails for a while , or is deactivated,  once it starts working again may return perpetual failure, caused memory overflow, when attempting to delete sometimes millions of jobs.
Example of standard code trying to delete 5.6 million "done" jobs:

 Call of self.env[u'queue.job'].autovacuum(*()) failed in Job #19
  File "/home/ubuntu/odoo/auto/addons/base/ir/ir_cron.py", line 126, in _callback
    getattr(model, method_name)(*args)
  File "/home/ubuntu/odoo/auto/addons/mcp_connector/models/queue_job/queue_job.py", line 22, in autovacuum
    ('state', '=', 'done')],
  File "/home/ubuntu/odoo/custom/src/odoo/odoo/models.py", line 1518, in search
    res = self._search(args, offset=offset, limit=limit, order=order, count=count)
  File "/home/ubuntu/odoo/custom/src/odoo/odoo/models.py", line 4304, in _search
    return _uniquify_list([x[0] for x in res])
  File "/home/ubuntu/odoo/custom/src/odoo/odoo/models.py", line 4302, in _uniquify_list
    return [x for x in seq if x not in seen and not seen.add(x)]
MemoryError

### DIFF
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -248,12 +248,19 @@ class QueueJob(models.Model):
         for channel in self.env['queue.job.channel'].search([]):
             deadline = datetime.now() - timedelta(
                 days=int(channel.removal_interval))
-            jobs = self.search(
-                [('date_done', '<=', fields.Datetime.to_string(deadline)),
-                 ('channel', '=', channel.complete_name)],
-            )
-            if jobs:
-                jobs.unlink()
+            while True:
+                jobs = self.search(
+                    [
+                        ("date_done", "<=", fields.Datetime.to_string(deadline))
+                        ("channel", "=", channel.complete_name),
+                    ],
+                    limit=1000,
+                )
+                if jobs:
+                    jobs.unlink()
+                    self.env.cr.commit()
+                else:
+                    break
         return True
 
     @api.model


gfcapalbo:
I see it failing because of explicit cr.commit.
I will fix according to:
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#never-commit-the-transaction

guewen (approved):

guewen:
> I see it failing because of explicit cr.commit.
> I will fix according to:
> https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#never-commit-the-transaction

I'm surprised we have no lint error in 13.0: https://github.com/OCA/queue/blob/0c6d71bb5e9b41fe78ddc33da0faad8f144a6579/queue_job/models/queue_job.py#L267

As this method is meant to be called by a cron, IMO it is okay to use `commit()` here, so I would just put a `  # pylint: disable=E8102` on the line.

queue_job/models/queue_job.py
```diff
@@ -248,12 +248,19 @@ def autovacuum(self):
...
-            jobs = self.search(
-                [('date_done', '<=', fields.Datetime.to_string(deadline)),
-                 ('channel', '=', channel.complete_name)],
-            )
-            if jobs:
-                jobs.unlink()
+            while True:
+                jobs = self.search(
+                    [
+                        ("date_done", "<=", fields.Datetime.to_string(deadline))
```
thomaspaulb, line 254: You forgot a comma here.

thomaspaulb (changes requested):

PR #401, opened 2022-01-12
APX-825

KalpeshGajera:


guewen:
Hi @KalpeshGajera, I see you are working on a dependency system for jobs. You should have a look on https://github.com/OCA/queue/pull/154 which is waiting for reviews. The latter PR targets 12.0, but as it is not approved yet, I never worked on adapting it to newer versions.

guewen:
@KalpeshGajera here's the version for odoo 14.0 https://github.com/OCA/queue/pull/403

PR #407, opened 2022-02-11
[13.0][FIX] base_import_async - User's lang is ignored

acsonefho:
Keep context during import.
In some case, the lang set into the context is important to do the import.
For example when the file contain a translatable name as a key (ex: product.attribute) and currently as there is no context, the search is executed in English (default lang) instead of user's lang.

Depends on https://github.com/OCA/queue/pull/406

### DIFF
--- a/base_import_async/models/base_import_import.py
+++ b/base_import_async/models/base_import_import.py
@@ -59,7 +59,9 @@ class BaseImportImport(models.TransientModel):
         attachment = self._create_csv_attachment(
             import_fields, data, options, self.file_name
         )
-        delayed_job = self.with_delay(description=description)._split_file(
+        delayed_job = self.with_delay(
+            description=description, keep_context=True
+        )._split_file(
             model_name=self.res_model,
             translated_model_name=translated_model_name,
             attachment=attachment,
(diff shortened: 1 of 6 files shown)


guewen:
Thanks for your PR. Any reason why the context is not stored in the recordset as I proposed in https://github.com/OCA/queue/issues/283?

rousseldenis:
@acsonefho

francesco-ooops:
Hi @acsonefho , let us know if you need any help finishing this. We're planning to port this to 14 as soon as it's ready

thomaspaulb:
@guewen We want to backport this to 11.0, but could you clarify your point about "storing it on the recordset" ? It's not clear to me from the issue you point to what you mean, or why it should be done.

guewen:
@thomaspaulb here: https://github.com/OCA/queue/pull/406#issuecomment-1083213345 but check with @acsonefho who maybe already started to work on it

acsonefho:
@thomaspaulb @guewen 
I finished yesterday the improvement on #406 but it's better to wait the merge before update this PR.

francesco-ooops:
Hi @acsonefho , what are the next steps in light of #432 having been merged?

acsonefho:
> Hi @acsonefho , what are the next steps in light of #432 having been merged?

I have to update this PR according to the one merge :)

francesco-ooops:
> I have to update this PR according to the one merge :)

makes sense :)

acsonefho:
I close this PR because with the context stored in the job, we have nothing else to do in this module.

PR #409, opened 2022-02-17
[RFC] jobrunner cli

sbidoul:
~/!\ This is very rough and mostly untested. So just an idea so far.~

I wonder if running the job runner in a standalone process could be a solution to the problem of running a single job runner in a multi-node deployment.

We could then let the responsibility of running one replica of the job runner to the orchestrator (say, k8s), and we don't need ensure it's a singleton ourselves.

Use with `env ODOO_RC=odoo.cfg python -m odoo.addons.queue_job.jobrunner`.

### DIFF
--- a/queue_job/jobrunner/__main__.py
+++ b/queue_job/jobrunner/__main__.py
@@ -1,0 +1,13 @@
+import odoo
+
+from .runner import QueueJobRunner
+
+
+def main():
+    odoo.tools.config.parse_config()
+    runner = QueueJobRunner.from_environ_or_config()
+    runner.run()
+
+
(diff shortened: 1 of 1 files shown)


sbidoul:
With this, there is no need for queue_job in server_wide_modules.

sbidoul:
This seems to work well in our first tests. I think we can merge this as it is non-intrusive. I'll document it when we have more experience in production.

guewen (approved):
Very good indeed for k8s-style infrastructure. Thanks!
We've run full loop with the old odoo-job-runner start script (don't remember the exact name) :laughing:

PR #412, opened 2022-03-11
[14.0] queue_job: use parent channel if configured

simahawk:
Use case:

* you have a root channel per scope/app (eg: root.edi)
* you have several sub channels (eg: root.edi.ubl.sales,
root.edi.gs1.delivery)
* you want to configure capacity only for the main channel "root.edi"

Before this change, the channel manager falls back on root channel,
and you get flooded w/ warning log entries like "unknown channel....".

However, if you have a specific parent channel configured
it sounds a good idea to use it.

NOTE: using `parent_fallback` flag is just an attempt to not break places where this feat is not desiderable.
Also, I might fail to see why this behavior was not supported before: any remark is welcomed :)

@guewen can I have your insights please? :pray:

### DIFF
--- a/queue_job/jobrunner/channels.py
+++ b/queue_job/jobrunner/channels.py
@@ -942,7 +942,9 @@ class ChannelManager(object):
         _logger.info("Configured channel: %s", channel)
         return channel
 
-    def get_channel_by_name(self, channel_name, autocreate=False):
+    def get_channel_by_name(
+        self, channel_name, autocreate=False, parent_fallback=False
+    ):
         """Return a Channel object by its name.
 
         If it does not exist and autocreate is True, it is created
@@ -980,6 +982,9 @@ class ChannelManager(object):
         >>> c = cm.get_channel_by_name('sub')
         >>> c.fullname
         'root.sub'
+        >>> c = cm.get_channel_by_name('root.sub.not.configured', parent_fallback=True)
+        >>> c.fullname
+        'root.sub.sub.not.configured'
         """
         if not channel_name or channel_name == self._root_channel.name:
             return self._root_channel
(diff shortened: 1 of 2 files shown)


queue_job/jobrunner/channels.py
```diff
@@ -980,16 +982,36 @@ def get_channel_by_name(self, channel_name, autocreate=False):
...
-        if not autocreate:
+        if not autocreate and not parent_fallback:
             raise ChannelNotFound("Channel %s not found" % channel_name)
         parent = self._root_channel
+        if parent_fallback:
+            # Look for first direct parent w/ config.
+            # Eg: `root.edi.foo.baz` will falback on `root.edi.foo`
+            # or `root.edi` or `root` in sequence
+            parent_name = channel_name
+            while True:
```
flotho, line 1003: Hi,
Are you sure that using such condition is recommended

flotho: otherwise LGTM

simahawk: what do you mean? I break in all cases if I reach the root channel.

queue_job/jobrunner/channels.py
```diff
@@ -980,16 +982,36 @@ def get_channel_by_name(self, channel_name, autocreate=False):
...
+                parent_name = parent_name.rsplit(".", 1)[:-1][0]
```
sebalix, line 1004: And if `channel_name == "root"`?
```python
>>> parent_name = "root"
>>> parent_name.rsplit(".", 1)[:-1][0]
  File "<stdin>", line 1, in <module>
IndexError: list index out of range
```

sebalix: Ah there is a safeguard on the first line to avoid it... nevermind!

simahawk: yes :smile:

guewen (approved):
This behavior is exactly the one I would expect in all situations.
Thanks!

flotho (approved):
LGTM

simahawk:
> This behavior is exactly the one I would expect in all situations. Thanks!

Cool. I've just dropped the try/except as it's useless now.

PR #338, opened 2021-04-21
[FIX] 14.0 doctests hidden errors

PieterPaulussen:
At the moment, a doctest that fails is not registered in the output of the test results.
I have tested this hypothesis here: https://github.com/OCA/queue/pull/336 where the tests do not fail, even though there is a clear error present!

PR #413, opened 2022-03-14
Sync from OCA/oca-addons-repo-template

simahawk:


simahawk:
[This commit](https://github.com/OCA/queue/commit/ec173cf2b9be4b273bc1697b5517c0a49b25e448) from #338 is a problem 
@guewen  @fernandahf . Unless there's no other way, do not change linter settings pls :pray: 

I added an ignore here but I wonder if we shouldn't use a related field instead.

simahawk:
FTR next round of build improvement https://github.com/OCA/queue/pull/414

PR #416, opened 2022-03-25
[14.0][FIX] queue_job: support for multi-db environments

RLeeOSI:
This change improves functionality in environments with multiple databases in the postgres cluster. Some environments have hidden databases, and may use list_db = False in favor of the db_filter config parameter. In these cases, exp_list will return an AccessError. Or, some environments may have multiple databases used by multiple instances of Odoo, managed by a load balancer. It's possible that not all DBs will have the queue_job module installed. Sometimes, we will want the job runner to work on a limited set of DBs.

This change adds the option of an environment variable, ODOO_QUEUE_JOB_DB_NAME, or a config parameter in the [queue] section, jobrunner_db_name. This takes a comma separated list of DB names, just like db_name does. If the variable/parameter is not set, it defaults to db_name (the previous functionality). 

Related:
#379 - Another fix for 12.0, solves the same problem when list_db=False
#58  - This resolves the issue

### DIFF
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -35,6 +35,7 @@ How to use it?
     or ``False`` if unset.
   - ``ODOO_QUEUE_JOB_JOBRUNNER_DB_PORT=5432``, default ``db_port``
     or ``False`` if unset.
+  - ``ODOO_QUEUE_JOB_JOBRUNNER_DB_NAME=firstdb,seconddb``, default ``db_name``
 
 * Alternatively, configure the channels through the Odoo configuration
   file, like:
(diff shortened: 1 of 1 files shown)


ska-ibees:
> This change improves functionality in environments with multiple databases in the postgres cluster. Some environments have hidden databases, and may use list_db = False in favor of the db_filter config parameter. In these cases, exp_list will return an AccessError. Or, some environments may have multiple databases used by multiple instances of Odoo, managed by a load balancer. It's possible that not all DBs will have the queue_job module installed. Sometimes, we will want the job runner to work on a limited set of DBs.
> 
> This change adds the option of an environment variable, ODOO_QUEUE_JOB_DB_NAME, or a config parameter in the [queue] section, jobrunner_db_name. This takes a comma separated list of DB names, just like db_name does. If the variable/parameter is not set, it defaults to db_name (the previous functionality).
> 
> Related: #379 - Another fix for 12.0, solves the same problem when list_db=False #58 - This resolves the issue

Can't thank enough for this... have been struggling to handle multi db setup... @RLeeOSI

PR #417, opened 2022-03-25
[MOD] Autovaccum of QueueJob, limit to 1k jobs per unlink

ssaid:


queue_job/models/queue_job.py
```diff
@@ -281,10 +281,18 @@ def autovacuum(self):
                 ['|',
                  ('date_done', '<=', deadline),
                  ('date_cancelled', '<=', deadline),
-                 ('channel', '=', channel.complete_name)],
+                 ('channel', '=', channel.complete_name)], limit=1000
```
flotho, line 284: won't it be better to set it up in the channel parameter like for the removal interval?
Otherwise LGTM

flotho (approved):

queue_job/models/queue_job.py
ivantodorovich: agree

queue_job/models/queue_job.py
```diff
@@ -281,10 +281,18 @@ def autovacuum(self):
...
-            if jobs:
+            while jobs:
                 jobs.unlink()
+                self._cr.commit()
+                jobs = self.search(
+                    ['|',
+                     ('date_done', '<=', deadline),
+                     ('date_cancelled', '<=', deadline),
+                     ('channel', '=', channel.complete_name)], limit=1000
+                )
```
ivantodorovich: Store the domain before calling the first search, so you can reuse it here too

```suggestion
                jobs = self.search(domain, limit=1000)
```

queue_job/models/queue_job.py
```diff
@@ -281,10 +281,18 @@ def autovacuum(self):
...
+                _logger.info('Unlink %s' % jobs.__len__())
```
ivantodorovich:
```suggestion
```

Is this really needed?

simahawk: maybe _logger.debug?

simahawk (changes requested):
agree w/ all the changes proposed by @ivantodorovich

queue_job/models/queue_job.py
etobella: Or even better: `_logger.debug("Unlink %s jobs" % len(jobs))`

queue_job/models/queue_job.py
guewen: I'm not sure it's useful, 1000 is generally a good number for this kind of batch operations, do you see cases where we would change this number?

queue_job/models/queue_job.py
guewen: Odoo already logs unlinks: https://github.com/odoo/odoo/blob/eeca5b27ff2494b4b31038750b487a4b0e5a9d8f/odoo/models.py#L3491

queue_job/models/queue_job.py
guewen: Or use this pattern:

```
while True:
    jobs = self.search(
        ['|',
         ('date_done', '<=', deadline),
         ('date_cancelled', '<=', deadline),
         ('channel', '=', channel.complete_name)], limit=1000
    )
    if not jobs:
        break
    jobs.unlink()
    self.env.cr.commit()
```

queue_job/models/queue_job.py
guewen: Proper link for odoo 12.0: https://github.com/odoo/odoo/blob/04e37773d1bf9b885be674d85bdf0ec069926958/odoo/models.py#L3223

rousseldenis:
@ssaid Could you attend comments ?

queue_job/models/queue_job.py
ssaid: I agree with @guewen, i would keep the 1000  because I don't see benefits to set that number as a parameter

flotho:
@ssaid , too many red flag in the test, could you fix them ?

PR #421, opened 2022-04-07
[14.0][ADD] base_import_async_lang : initial commit

AshishHirapara:
Added a module to Import by language which sets the job queue language to take into account users' language when importing, in order to search for records in that language.


francesco-ooops (approved):
Tested locally, LGTM

francesco-ooops:
@guewen do you think this can be a good solution?

(btw, which PSC is in charge of queue repo?)

PR #426, opened 2022-05-02
[12.0] apply OCA repo template

sbidoul:


sbidoul:
Hm, GitHub actions don't run, likely because they are not enabled on the target branch. But runboat is green and pre-commit passes locally so this is hopefullly going to be green. So merging manually.

sbidoul:
Ow, no, I used 11.0 as version :facepalm: I'll fix manually.

guewen:
Oh thanks!

PR #430, opened 2022-05-11
[15.0][MIG] base_import_async

RosSinch:


base_import_async/__manifest__.py
```diff
@@ -11,8 +11,15 @@
...
+    "data": ["data/queue_job_function_data.xml"],
+    "assets": {
+        "web.assets_backend": [
+            "base_import_async/static/src/js/import.js",
+        ],
+        'web.assets_qweb': [
+            'base_import_async/static/src/xml/import.xml',
+        ],
+    },
     "installable": False,
```
guewen: Thank you @RosSinch :)
It is still not installable :thinking:

RosSinch: I've given up :)
I did it.
But could you please explain to me why it should be installable in v15 if it wasn't in v14?

guewen: `"installable": False,` was false in v14 because nobody did the migration work in this version, so it could not be used in v14. Perhaps odoo will still install it using the `-i` option, not sure, but this would be a kind of bug, the module would be shown as uninstallable in the UI.

As I understand you actually did the migration work so the module runs on v15, then the module should be flagged as installable, it would be a pity if the module cannot be installed if you made it installable, right? :)

RosSinch: Yes, you are absolutely right.  It should be installable and it can, because the work is done for this.

RosSinch: @guewen 
@simahawk 
Is PR ready to merge?

edison12347:
Should we expect this migration to be merged any time soon?

base_import_async/__manifest__.py
simahawk: Build is :red_circle: 
Also, please, when you get back to it, squash your commits and rewrite the 1 remaining as `[15.0][MIG] base_import_async`

simahawk:
Note that there's another PR now here https://github.com/OCA/queue/pull/434.
You might want to join efforts :wink:

Rad0van:
@RosSinch as @simahawk noted I have created PR for this as well somehow not noticing yours. I did few other things so please have a look at them.  All my test pass so there may be some inspiration. Or if you do not know ho to proceed close the PR and I will take over with mine.

Rad0van:
Now that this one was closed we could continue with the other one that's in better shape.

PR #432, opened 2022-05-12
[13.0] Add storage of context in jobs recordsets

guewen:
Alternative to #406.
Adapted from https://github.com/acsone/queue/pull/1 with tests.

This version has a breaking change in the sense that now the basic arguments such as tz, lang, etc in the context are kept, so may bring behavior changes in the existing jobs. In a way, this could be considered a bug so... to discuss (or consider this is the version we want for 16.0, and the list will be empty in earlier versions).

I would not keep all keys by default for several reasons :

* not all values can be serialized to json (random python objects, ...)
* I'm not sure we really do want to keep all the keys (e.g. `prefetch_fields`)

This implementation may be simple and naive, but I do think it is enough, as people can customize the keys allowed in the context. Also, more smart solutions could easily lead to leaky abstractions, because jobs can have contexts in their `self` recordset, but also in `args` and `kwargs` which potentially would all need different handling (hence a global handling works for each case).

### DIFF
--- a/queue_job/models/base.py
+++ b/queue_job/models/base.py
@@ -223,13 +223,9 @@ class Base(models.AbstractModel):
 
         Empty by default for backward compatibility.
         """
-        return (
-            "tz",
-            "lang",
-            "allowed_company_ids",
-            "force_company",
-            "active_test",
-        )
+        # TODO: when migrating to 16.0, active the base context keys:
+        # return ("tz", "lang", "allowed_company_ids", "force_company", "active_test")
+        return ()
 
     def _job_prepare_context_before_enqueue(self):
         """Return the context to store in the jobs
--- a/queue_job/models/base.py
+++ b/queue_job/models/base.py
@@ -216,3 +216,28 @@ class Base(models.AbstractModel):
         :return: dictionary for setting job values.
         """
         return {}
+
+    @api.model
+    def _job_prepare_context_before_enqueue_keys(self):
+        """Keys to keep in context of stored jobs
+
+        Empty by default for backward compatibility.
+        """
+        return (
+            "tz",
+            "lang",
+            "allowed_company_ids",
+            "force_company",
+            "active_test",
+        )
+
+    def _job_prepare_context_before_enqueue(self):
+        """Return the context to store in the jobs
+
+        Can be used to keep only safe keys.
+        """
+        return {
+            key: value
+            for key, value in self.env.context.items()
+            if key in self._job_prepare_context_before_enqueue_keys()
+        }
--- a/queue_job/fields.py
+++ b/queue_job/fields.py
@@ -69,6 +69,9 @@ class JobSerialized(fields.Field):
 class JobEncoder(json.JSONEncoder):
     """Encode Odoo recordsets so that we can later recompose them"""
 
+    def _get_record_context(self, obj):
+        return obj._job_prepare_context_before_enqueue()
+
     def default(self, obj):
         if isinstance(obj, models.BaseModel):
             return {
@@ -77,6 +80,7 @@ class JobEncoder(json.JSONEncoder):
                 "ids": obj.ids,
                 "uid": obj.env.uid,
                 "su": obj.env.su,
+                "context": self._get_record_context(obj),
             }
         elif isinstance(obj, datetime):
             return {"_type": "datetime_isoformat", "value": obj.isoformat()}
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -423,8 +423,6 @@ class Job(object):
         :param identity_key: A hash to uniquely identify a job, or a function
                              that returns this hash (the function takes the job
                              as argument)
-        :param env: Odoo Environment
-        :type env: :class:`odoo.api.Environment`
         """
         if args is None:
             args = ()
--- a/queue_job/readme/USAGE.rst
+++ b/queue_job/readme/USAGE.rst
@@ -121,8 +121,26 @@ Based on this configuration, we can tell that:
 **Job Context**
 
 The context of the recordset of the job, or any recordset passed in arguments of
-a job, is transferred to the job according to an allow-list. The allow-list can
+a job, is transferred to the job according to an allow-list.
+
+The default allow-list is empty for backward compatibility. The allow-list can
 be customized in ``Base._job_prepare_context_before_enqueue_keys``.
+
+Example:
+
+.. code-block:: python
+
+   class Base(models.AbstractModel):
+
+       _inherit = "base"
+
+       @api.model
+       def _job_prepare_context_before_enqueue_keys(self):
+           """Keys to keep in context of stored jobs
+
+           Empty by default for backward compatibility.
+           """
+           return ("tz", "lang", "allowed_company_ids", "force_company", "active_test")
 
 **Bypass jobs on running Odoo**
 
--- a/queue_job/readme/USAGE.rst
+++ b/queue_job/readme/USAGE.rst
@@ -118,6 +118,12 @@ Based on this configuration, we can tell that:
 * retries 10 to 15 postponed 30 seconds later
 * all subsequent retries postponed 5 minutes later
 
+**Job Context**
+
+The context of the recordset of the job, or any recordset passed in arguments of
+a job, is transferred to the job according to an allow-list. The allow-list can
+be customized in ``Base._job_prepare_context_before_enqueue_keys``.
+
 **Bypass jobs on running Odoo**
 
 When you are developing (ie: connector modules) you might want
--- a/queue_job/tests/test_json_field.py
+++ b/queue_job/tests/test_json_field.py
@@ -16,7 +16,8 @@ from odoo.addons.queue_job.fields import JobDecoder, JobEncoder
 class TestJson(common.TransactionCase):
     def test_encoder_recordset(self):
         demo_user = self.env.ref("base.user_demo")
-        partner = self.env(user=demo_user).ref("base.main_partner")
+        context = demo_user.context_get()
+        partner = self.env(user=demo_user, context=context).ref("base.main_partner")
         value = partner
         value_json = json.dumps(value, cls=JobEncoder)
         expected = {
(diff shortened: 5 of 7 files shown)


queue_job/models/base.py
```diff
@@ -216,3 +216,28 @@ def _job_store_values(self, job):
         :return: dictionary for setting job values.
         """
         return {}
+
+    def _job_prepare_context_before_enqueue_keys(self):
```
sbidoul, line 221: `@api.model` ?

guewen: Done :+1:

queue_job/models/base.py
```diff
@@ -216,3 +216,29 @@ def _job_store_values(self, job):
         :return: dictionary for setting job values.
         """
         return {}
+
+    @api.model
+    def _job_prepare_context_before_enqueue_keys(self):
+        """Keys to keep in context of stored jobs
+
+        Empty by default for backward compatibility.
```
simahawk, line 224:
```suggestion
```

queue_job/models/base.py
```diff
@@ -216,3 +216,29 @@ def _job_store_values(self, job):
...
+        """
+        return {
+            "tz",
+            "lang",
+            "uid",
+            "allowed_company_ids",
+            "force_company",
+            "active_test",
+        }
```
simahawk:
```suggestion
        return (
            "tz",
            "lang",
            "uid",
            "allowed_company_ids",
            "force_company",
            "active_test",
        )
```

queue_job/models/base.py
sbidoul, line 224: I must think more about it but my first intuition is that we should indeed keep it empty and enable this list by default in 16.0.

queue_job/models/base.py
```diff
@@ -216,3 +216,29 @@ def _job_store_values(self, job):
...
+    @api.model
+    def _job_prepare_context_before_enqueue_keys(self):
+        """Keys to keep in context of stored jobs
+
+        Empty by default for backward compatibility.
+        """
+        return {
+            "tz",
+            "lang",
+            "uid",
```
sbidoul: `uid` in context ? Is that common ?

queue_job/models/base.py
guewen: Of course, I'm starting to lose my python skills 😱

queue_job/models/base.py
guewen: I searched for context in odoo core, and found this occurrence https://github.com/odoo/odoo/blob/ac3cf2ed59c1dd5a6c91da201a0d4093212b26ae/odoo/http.py#L1082, but it's probably an edge case for requests that we don't need "internally"

queue_job/models/base.py
simahawk: Might be a wise choice :)

guewen: And it's easy to override the empty list in new projects or if we accept the risk.
Probably better to do it this way, yeah. I'll leave the code as-is to see if we can agree on the *target* for 16.0, then make the backward compatible version (which will be: empty the list, move the proposition of keys in the readme, adapt the tests).

guewen: ah yes, well, we'll see if I remove this or the return value, probably the latter

queue_job/models/base.py
guewen: Removed

florian-dacosta (approved):
LGTM thanks!

francesco-ooops:
@sbidoul @simahawk do you think we can proceed with this?

sbidoul (approved):
Yes, fine with me.

simahawk (approved):

simahawk:
/ocabot merge minor

Any plan for v14?

AshishHirapara:
> /ocabot merge minor
> 
> Any plan for v14?

@simahawk I have migrated this PR to 14.0 ==> https://github.com/OCA/queue/pull/433

francesco-ooops:
Hi @guewen @simahawk @acsonefho , so going back to the original need (from PR #407 ):

> Keep context during import.
In some case, the lang set into the context is important to do the import.
For example when the file contain a translatable name as a key (ex: product.attribute) and currently as there is no context, the search is executed in English (default lang) instead of user's lang.

How do you suggest to implement this? should we create a new module "base_import_async_lang" here on OCA to set language in the context? 
Thanks!

francesco-ooops:
@OCA/connector-maintainers

acsonefho:
> Hi @guewen @simahawk @acsonefho , so going back to the original need (from PR #407 ):
> 
> > Keep context during import.
> > In some case, the lang set into the context is important to do the import.
> > For example when the file contain a translatable name as a key (ex: product.attribute) and currently as there is no context, the search is executed in English (default lang) instead of user's lang.
> 
> How do you suggest to implement this? should we create a new module "base_import_async_lang" here on OCA to set language in the context? Thanks!

Maybe you can create a module who contains only this:
https://github.com/OCA/queue/pull/432/files#diff-c6aeb210a5f309f0b1f3ad1cc16e866ae61bab6793ac67eb2c187a28c1664fb6R41

Then ping Guewen and Simahawk to have review and feedback :)

francesco-ooops:
@acsonefho thanks a lot!

simahawk:
Instead of adding a new module, you can make `base_import_async` configurable. Eg: ir.config param by model to get the ctx keys to store by overriding `_job_prepare_context_before_enqueue_keys`. If you find a ctx key `job_ctx_store_keys__$model_name` you pick keys from there.
Hmm now that I write this I think we can add this option to queue_job straight... :thinking: 
@guewen what do you think?

hildickethan:
Hello, I've been trying to use this feature to send a context from the creation of the job to the actual job runtime itself but I feel like there's something missing.
I inherited `_job_prepare_context_before_enqueue_keys` and added my key, this works to store the context and it saves in the `records` field as expected, but I feel like [this bit](https://github.com/OCA/queue/pull/406/files#diff-46de3209aa1021218a8cb6d018b955d92422b74022b3cca97ae4fd907a23ec9bR115-R117) from the original PR https://github.com/OCA/queue/pull/406 is missing to complete the feature.

With this added it runs the jobs with the context that was passed, plus the ones added later for `job_uuid` and company, but right now it saves the context and does nothing with it.

I just want to check if this is intentionally left out (for backwards compatibility I guess?) and is planned for 16.0 or it was an oversight, because I couldn't find any discussion on this part specifically in the various related PRs

It can be worked around by searching the `job_uuid` and reading the field for the context, but it seems a bit weird to me

francesco-ooops:
@simahawk how shall we proceed?

sbidoul:
@francesco-ooops  Expanding on @acsonefho 's suggestion, I'd create a module that does only this:
https://github.com/OCA/queue/blob/c8b16ff73a85f72f8b84bbb17a774f44587be875/queue_job/models/base.py#L261
Since that will become the default in v16, it seems reasonable to have module that anticipates the default, which you can install if you are sure that all your jobs are compatible with that.

> I feel like [this bit](https://github.com/OCA/queue/pull/406/files#diff-46de3209aa1021218a8cb6d018b955d92422b74022b3cca97ae4fd907a23ec9bR115-R117) from the original PR https://github.com/OCA/queue/pull/406 is missing to complete the feature.

@hildickethan-S73 that makes sense (disclaimer: I've not tested myself). Would you mind doing a PR with the missing part?

guewen:
I totally agree with @acsonefho and @sbidoul proposal, it's like an opt-in to the behavior that will be the default in 16.0

francesco-ooops:
@sbidoul @guewen thank you both!

guewen:
@hildickethan-S73 ooch, that's indeed an oversight from my part 🤦.

This line that initializes the record(s) from the Json in JobDecoder has to set the context: 

            model = self.env(user=obj.get("uid"), su=obj.get("su"))[obj["model"]]

(Sorry, I'm on my phone and can't manage to copy-paste the exact link in the GitHub app😓)

hildickethan:
@guewen Just wanted confirmation on it being intentional or not! 
Here we go https://github.com/OCA/queue/pull/448
I just copied the code that was in the original PR, the if might not be necessary but I haven't tested without

AshishHirapara:
@guewen and @sbidoul should I also need to update the discussed changes in https://github.com/OCA/queue/pull/433? It's for v14.

GSLabIt:
> @francesco-ooops Expanding on @acsonefho 's suggestion, I'd create a module that does only this:
> 
> https://github.com/OCA/queue/blob/c8b16ff73a85f72f8b84bbb17a774f44587be875/queue_job/models/base.py#L261
> 
> 
> Since that will become the default in v16, it seems reasonable to have module that anticipates the default, which you can install if you are sure that all your jobs are compatible with that.
> > I feel like [this bit](https://github.com/OCA/queue/pull/406/files#diff-46de3209aa1021218a8cb6d018b955d92422b74022b3cca97ae4fd907a23ec9bR115-R117) from the original PR #406 is missing to complete the feature.
> 
> @hildickethan-S73 that makes sense (disclaimer: I've not tested myself). Would you mind doing a PR with the missing part?

Hi there, we developed a module to return the `context` as described, but the import (using `base_import_async`) failing because of not matching the translated country name (ie for italian language):

```
No matching record found for name 'Regno Unito' in field 'Country'
No matching record found for name 'Germania' in field 'Country'
No matching record found for name 'Francia' in field 'Country'
No matching record found for name 'Paesi Bassi' in field 'Country'
```

Are we missing something?

GSLabIt:
> > @francesco-ooops Expanding on @acsonefho 's suggestion, I'd create a module that does only this:
> > https://github.com/OCA/queue/blob/c8b16ff73a85f72f8b84bbb17a774f44587be875/queue_job/models/base.py#L261
> > 
> > Since that will become the default in v16, it seems reasonable to have module that anticipates the default, which you can install if you are sure that all your jobs are compatible with that.
> > > I feel like [this bit](https://github.com/OCA/queue/pull/406/files#diff-46de3209aa1021218a8cb6d018b955d92422b74022b3cca97ae4fd907a23ec9bR115-R117) from the original PR #406 is missing to complete the feature.
> > 
> > 
> > @hildickethan-S73 that makes sense (disclaimer: I've not tested myself). Would you mind doing a PR with the missing part?
> 
> Hi there, we developed a module to return the `context` as described, but the import (using `base_import_async`) failing because of not matching the translated country name (ie for italian language):
> 
> ```
> No matching record found for name 'Regno Unito' in field 'Country'
> No matching record found for name 'Germania' in field 'Country'
> No matching record found for name 'Francia' in field 'Country'
> No matching record found for name 'Paesi Bassi' in field 'Country'
> ```
> 
> Are we missing something?

sorry my fault ..working properly

francesco-ooops:
> @francesco-ooops Expanding on @acsonefho 's suggestion, I'd create a module that does only this:
> 
> https://github.com/OCA/queue/blob/c8b16ff73a85f72f8b84bbb17a774f44587be875/queue_job/models/base.py#L261

Hi @sbidoul , here is the module, is it ok for you? https://github.com/OCA/queue/pull/462

PR #440, opened 2022-06-15
[14.0][FIX] It seems that sometimes jobs doesn't have a job_method and the __repr__ failed

acsonefho:
That fix the issue into Shopinvader: https://github.com/shopinvader/odoo-shopinvader/runs/6897055192?check_suite_focus=true#step:8:7767 (cfr screenshot).
The issue happens into every PR in v14.0 (during merge or after a rebase).

The issue happens during the call to the `__repr__` of the `Delayable` object when the attribute `_job_method` is `None`.
https://github.com/OCA/queue/blob/14.0/queue_job/delay.py#L496
I didn't clearly identify why the `_job_method` is not filled. So maybe this PR is not a good solution.

image

### DIFF
--- a/queue_job/delay.py
+++ b/queue_job/delay.py
@@ -492,8 +492,11 @@ class Delayable:
         return [self]
 
     def __repr__(self):
+        job_method = ""
+        if self._job_method:
+            job_method = self._job_method.__name__
         return "Delayable({}.{}({}, {}))".format(
-            self.recordset, self._job_method.__name__, self._job_args, self._job_kwargs
+            self.recordset, job_method, self._job_args, self._job_kwargs
         )
 
     def __del__(self):


rousseldenis:
@acsonefho is there a mean to write a test to reproduce?

queue_job/delay.py
```diff
@@ -492,8 +492,11 @@ def _tail(self):
         return [self]
 
     def __repr__(self):
+        job_method = ""
```
rousseldenis, line 495: Shouldn't it be something more user friendly than a void string ?

rousseldenis:
@guewen What do you think of this ?

rousseldenis (approved):
Quite trivial as job_method slot is None by default.

Even if this could be enhanced, this is sufficient to avoid errors

queue_job/delay.py
acsonefho: That happens when the current object is related to nothing. So what do you propose? :)

guewen:
We see in the traceback that it happens in the `__del__` method, where we check that we didn't create a `Delayable` that we forgot to `delay()`. It `repr` the delayable not delayed. Do you have an idea why we fall in this case in the first place?

guewen (approved):

acsonefho:
> We see in the traceback that it happens in the `__del__` method, where we check that we didn't create a `Delayable` that we forgot to `delay()`. It `repr` the delayable not delayed. Do you have an idea why we fall in this case in the first place?

Thanks for the review.
But I really don't know and I can't reproduce it :/

PR #441, opened 2022-06-21
[14.0] Fix cursor already closed error on retryable errors

guewen:
When a job is postponed because of a retryable error, as the error
is not re-raised, we reach the new method

    self._enqueue_dependent_jobs(env, job)

With a closed cursor. Return early in case of postponing, as there is
no chance dependent jobs can become pending anyway.


simahawk (approved):
seems legit

yvaucher (approved):

PR #443, opened 2022-06-22
[14.0][FIX] queue_job: Add base_sparse_field dependency to prevent errors in test if addon not loaded.

victoralmau:
Add `base_sparse_field` dependency to prevent errors in test if addon not loaded. https://github.com/OCA/queue/blob/14.0/queue_job/models/queue_job.py#L12

@Tecnativa TT35938

### DIFF
--- a/queue_job/README.rst
+++ b/queue_job/README.rst
@@ -105,7 +105,7 @@ Configuration
   [options]
   (...)
   workers = 6
-  server_wide_modules = web,queue_job
+  server_wide_modules = web,queue_job,base_sparse_field
 
   (...)
   [queue_job]
--- a/queue_job/__manifest__.py
+++ b/queue_job/__manifest__.py
@@ -7,7 +7,7 @@
     "website": "https://github.com/OCA/queue",
     "license": "LGPL-3",
     "category": "Generic Modules",
-    "depends": ["mail"],
+    "depends": ["mail", "base_sparse_field"],
     "external_dependencies": {"python": ["requests"]},
     "data": [
         "security/security.xml",
--- a/queue_job/readme/CONFIGURE.rst
+++ b/queue_job/readme/CONFIGURE.rst
@@ -18,7 +18,7 @@
   [options]
   (...)
   workers = 6
-  server_wide_modules = web,queue_job
+  server_wide_modules = web,queue_job,base_sparse_field
 
   (...)
   [queue_job]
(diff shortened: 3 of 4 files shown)


pedrobaeza (approved):
Seen the dependency in CI builds

cesar-tecnativa (approved):

victoralmau:
Maybe it's ready to merge too, right?

pedrobaeza:
/ocabot merge patch

Needed in v15?

queue_job/README.rst
```diff
@@ -105,7 +105,7 @@ Configuration
   [options]
   (...)
   workers = 6
-  server_wide_modules = web,queue_job
+  server_wide_modules = web,queue_job,base_sparse_field
```
simahawk, line 108: To my knowledge, this is NOT needed when the field is imported directly instead of using the patched `odoo.fields.Serialized` (like in queue_job).

pedrobaeza: But I have shown you the capture with the error in our CI when we run the server without this:


imagen

simahawk: This problem might come from the fact that `base_sparse_field` is not available in your env (for whatever reason) or you are using an outdated version of queue_job... I don't know, can't judge from a screenshot :man_shrugging:
We never had this problem on any of our instances or dev setup or CI setup. Even when we use `--load=queue_job`.

FTR this was fixed a looong time ago here https://github.com/OCA/queue/commit/547538bd7860ced0f69d4cc257dcfc708d8870a4 and if you look at the change we removed the need for the server wide module. Hence it looks weird, at least :wink:

pedrobaeza: The code of `queue_job` is updated to latest one. Don't you think that not depending in the manifest on `base_sparse_field` is an error?

pedrobaeza: And what's the problem with having such server wide module loaded?

simahawk: I'm arguing only about the `server_wide_modules` change here. The dependency change IS correct.

pedrobaeza: But one is attached to the other: if you don't add it to the server wide modules, then it's not loaded at all. And I insist: is there any problem adding to the server wide modules that I'm not aware of?

simahawk: >  if you don't add it to the server wide modules, then it's not loaded at all

AFAIK This is not true. It was required WHEN we were using the patch.

AFAIK The server wide module IS NOT needed. If you want to recommend it for certain cases, fine, but if you want to state in the docs that is required, to me is wrong. Unless proven otherwise. This is my point.

pedrobaeza: Tested and indeed the server wide is not needed. @victoralmau please do a PR in 14.0 for removing such change in the documentation.

PR #448, opened 2022-07-18
[IMP] queue_job: Create job with saved context

hildickethan:
- Add saved context to the job
- This was in the first draft https://github.com/OCA/queue/pull/406 but was missing in the https://github.com/OCA/queue/pull/432 PR that ended up being merged due to an oversight


guewen (approved):
Thanks!

acsonefho (approved):

PR #449, opened 2022-07-29
[14.0] base_import_async: add a dedicated channel

sebalix:
So we can configure the size of the channel.

We experienced a lot of locks and concurrent update issues on a table because tens of jobs were running in parallel (`root` is set with a high value).

### DIFF
--- a/base_import_async/data/queue_job_function_data.xml
+++ b/base_import_async/data/queue_job_function_data.xml
@@ -1,7 +1,13 @@
 <odoo noupdate="1">
+    <record model="queue.job.channel" id="channel_base_import">
+        <field name="name">base_import</field>
+        <field name="parent_id" ref="queue_job.channel_root" />
+    </record>
+
     <record id="job_function_base_import_import_split_file" model="queue.job.function">
         <field name="model_id" ref="base_import.model_base_import_import" />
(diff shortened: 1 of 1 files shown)


base_import_async/data/queue_job_function_data.xml
```diff
@@ -1,7 +1,13 @@
 <odoo noupdate="1">
+    <record model="queue.job.channel" id="channel_base_import">
+        <field name="name">base_import</field>
+        <field name="parent_id" ref="queue_job.channel_root" />
+    </record>
+
     <record id="job_function_base_import_import_split_file" model="queue.job.function">
         <field name="model_id" ref="base_import.model_base_import_import" />
         <field name="method">_split_file</field>
+        <field name="channel_id" ref="channel_base_import" />
```
simahawk, line 10: to make these changes effective we need a migration step IMO

guewen: It can be changed by users, could be good to have a migration (only if no other channel than `root` was configured manually), but is not strictly required IMO

guewen (approved):

sbidoul (approved):

lmignon (approved):

PR #450, opened 2022-08-03
Update job.py

Aurelien-michon:
I had an issue where the execution of the queue job would not except a RetryableJobError, but still end up requeuing itself (because a few rows failed ? honestly not sure). This caused it to never pass in the check for max retries. Putting that check before attempting the job fixed that issue.

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -516,6 +516,15 @@ class Job(object):
         The job is executed with the user which has initiated it.
         """
         self.retry += 1
+        if self.retry > self.max_retries:
+            type_, value, traceback = sys.exc_info()
+            # change the exception type but keep the original
+            # traceback and message:
+            # http://blog.ianbicking.org/2007/09/12/re-raising-exceptions/
+            new_exc = FailedJobError(
+                "Max. retries (%d) reached: %s" % (self.max_retries, value or type_)
+            )
+            raise new_exc
         try:
             self.result = self.func(*tuple(self.args), **self.kwargs)
         except RetryableJobError as err:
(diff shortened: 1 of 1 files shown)


guewen:
Hi, thanks for your PR.
The issue with your change is that we lose the exception information that we store in the job on a failure.

It would be nice if you could get more details on the issue you got. Was your issue that the retry count exceeded the max of retries? If so, can you check if the exception is not raised with `ignore_retry=True`?

https://github.com/OCA/queue/blob/d7840260768b8819611b609b50b65082cfe72b12/queue_job/exception.py#L21-L34

Aurelien-michon:
Hi @guewen, thanks for the quick answer, 
I don't have my exact issue at hand anymore, but basically the job kept retrying and increasing the retry counter, I'm pretty sure that means that ignore retry was false since otherwise it would have been decreased back down here : 
` self.result = self.func(*tuple(self.args), **self.kwargs)
        except RetryableJobError as err:
            if err.ignore_retry:
                self.retry -= 1
                raise`

I had to kill the jobs manually or they'd just keep running with retries counter in the interface going to 6/5, 7/5, etc. 

Regarding my change, how about retrieving the last execution's exception and adding it to the max retry reached message then ? Isn't the queue_job.result field stored from the last perform attempt ?

I admit I kinda struggle fully understanding the logic of this module, with queue_jobs being db objects and jobs being based on those.

guewen:
Well then possibly it is rather than `self.retry -= 1` is never actually stored to the database.
I would prefer to fix the root case rather than find a work-around, such as moving the retry logic on the next execution of the job as this change is doing.

Aurelien-michon:
Yeah, I understand.
My problem is that I'm also using another module that is based on queue and perhaps the issue is hidden somewhere in there, I'll keep looking if I can figure out what exactly is going on and I'll update you. 
I'm not sure when I'll be able to do that though, my little hack is enough for my needs and I'm super busy atm. 
If you want to close this, it's fine for me, and I'll reopen it if it makes sense. 

Thanks a lot for your time though !

PR #451, opened 2022-08-11
[15.0][FIX] queue_job: Add base_sparse_field dependency to prevent errors in test if addon not loaded.

victoralmau:
FWP from 14.0: https://github.com/OCA/queue/pull/443

Add `base_sparse_field` dependency to prevent errors in test if addon not loaded. https://github.com/OCA/queue/blob/14.0/queue_job/models/queue_job.py#L12

Please @pedrobaeza can you review it?

@Tecnativa TT35938

### DIFF
--- a/queue_job/README.rst
+++ b/queue_job/README.rst
@@ -105,7 +105,7 @@ Configuration
   [options]
   (...)
   workers = 6
-  server_wide_modules = web,queue_job
+  server_wide_modules = web,queue_job,base_sparse_field
 
   (...)
   [queue_job]
--- a/queue_job/__manifest__.py
+++ b/queue_job/__manifest__.py
@@ -8,7 +8,7 @@
     "website": "https://github.com/OCA/queue",
     "license": "LGPL-3",
     "category": "Generic Modules",
-    "depends": ["mail"],
+    "depends": ["mail", "base_sparse_field"],
     "external_dependencies": {"python": ["requests"]},
     "data": [
         "security/security.xml",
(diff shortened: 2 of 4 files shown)


pedrobaeza (approved):

cesar-tecnativa (approved):

queue_job/README.rst
```diff
@@ -105,7 +105,7 @@ Configuration
   [options]
   (...)
   workers = 6
-  server_wide_modules = web,queue_job
+  server_wide_modules = web,queue_job,base_sparse_field
```
simahawk, line 108: IMO this is not needed when the fields is imported from the module like in v14 https://github.com/OCA/queue/blob/14.0/queue_job/models/queue_job.py#L12 and is not needed at all on v15 AFAIK since a custom field is used instead https://github.com/OCA/queue/blob/15.0/queue_job/fields.py#L14

Am I missing anything?

simahawk (changes requested):
Are you sure we need this on v15?

simahawk:
I canceled the build for the merge while waiting for an answer :wink:

queue_job/README.rst
pedrobaeza: On v14 was needed for sure. If not, we got this error:

imagen

For v15 indeed the sparse field is gone, so we can close this.

Thanks for being aware of it.

simahawk: Welcome :) 
I'll comment on the v14 PR for the other detail.

PR #464, opened 2022-10-18

baimont:
Port of the following PRs from 14.0 to 16.0:
- #390
- #387


sbidoul:
We don't want #387. See https://github.com/OCA/queue/pull/387#issuecomment-1282061718.

PR #462, opened 2022-10-13
[ADD][14.0] queue_job_context

AshishHirapara:
initial commit - Add context before queue

PR #466, opened 2022-10-18
[14.0] revert #387

sbidoul:
Revert #387 which should not have been merged, and update test to illustrate the proper way to pass with_company to jobs.

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -699,13 +699,7 @@ class Job(object):
 
     @property
     def func(self):
-        # We can fill only one company into allowed_company_ids.
-        # Because if you have many, you can have unexpected records due to ir.rule.
-        # ir.rule use allowed_company_ids to load every records in many companies.
-        # But most of the time, a job should be executed on a single company.
         recordset = self.recordset.with_context(job_uuid=self.uuid)
-        if self.company_id:
-            recordset = recordset.with_company(self.company_id)
         return getattr(recordset, self.method_name)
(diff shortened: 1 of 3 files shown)


sbidoul:
@acsonefho @Cedric-Pigeon

acsonefho (approved):
That make sense

guewen (approved):
Oh thanks!

rousseldenis (approved):

pedrobaeza:
But the problem about having several allowed companies in the context still remains, isn't it?

sbidoul:
> But the problem about having several allowed companies in the context still remains, isn't it?

Which problem exactly?

pedrobaeza:
That the method that is executed in the job doesn't have the expected main company / allowed companies.

sbidoul:
Doing `env[model].with_company(company).with_delay().method()`, together with #462 (which will become the default in 16), is the proper way to do it and will correctly propagate allowed_company_ids to the jobs.

pedrobaeza:
And why a separate module?

sbidoul:
> And why a separate module?

For backward compatibility. In 16.0 it becomes the default. See https://github.com/OCA/queue/pull/432#issuecomment-1187473985 and the many comments before and in linked issues and PRs for context :)

pedrobaeza:
Well, OK, let's have such module in mind in multi-company environments...

/ocabot merge patch

PR #470, opened 2022-10-19
[16.0][FW] [14.0][FIX] queue_job: Add base_sparse_field dependency to prevent errors in test if addon not loaded.

baimont:
Port of #443 from 14.0 to 16.0.


baimont:
this pr should not be merged as the 14.0 one has been reverted. instead, this one https://github.com/OCA/queue/pull/471 should be

PR #453, opened 2022-08-16
[14.0][FIX] queue_job: Update documentation to remove unnecessary addon (base_sparse_field) from server_wide_modules.

victoralmau:
Update documentation to remove unnecessary addon (`base_sparse_field`) from server_wide_modules.

Related to: https://github.com/OCA/queue/pull/443#discussion_r946913070

@Tecnativa

PR #471, opened 2022-10-19

baimont:
Port of the following PRs from 14.0 to 16.0:
- #443 
- #453

### DIFF
--- a/queue_job/README.rst
+++ b/queue_job/README.rst
@@ -105,7 +105,7 @@ Configuration
   [options]
   (...)
   workers = 6
-  server_wide_modules = web,queue_job,base_sparse_field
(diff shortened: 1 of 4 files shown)


pedrobaeza:
This is not changing anything right now. What is needed AFAIK is the module dependency, but not the server_wide_modules change.

baimont:
> This is not changing anything right now. What is needed AFAIK is the module dependency, but not the server_wide_modules change.

I'll adapt soon

sbidoul (approved):

rvalyi (approved):
yes it changed the README and changed it back again, but this was done with these 2 steps in v14 so I think it's okay to just cherry-pick the  2 commits the way it is done in this PR.

PR #487, opened 2022-12-01
[FIX] queue_job: delayable representation failed

etobella:
Represenation might fall in case of no job definition (as it happens on https://github.com/OCA/edi/blob/15.0/edi_oca/tests/test_record.py#L127), it tries to generate the log warning, but it fails

@simahawk

### DIFF
--- a/queue_job/delay.py
+++ b/queue_job/delay.py
@@ -493,7 +493,10 @@ class Delayable:
 
     def __repr__(self):
         return "Delayable({}.{}({}, {}))".format(
-            self.recordset, self._job_method.__name__, self._job_args, self._job_kwargs
+            self.recordset,
+            self._job_method.__name__ if self._job_method else "",
+            self._job_args,
+            self._job_kwargs,
         )
 
     def __del__(self):


etobella:
The error raised is the following:

https://github.com/OCA/edi/actions/runs/3593875771/jobs/6051404638#step:8:127

simahawk (approved):
LG, maybe worth to adapt the tests?

queue_job/delay.py
```diff
@@ -493,7 +493,10 @@ def _tail(self):
 
     def __repr__(self):
         return "Delayable({}.{}({}, {}))".format(
-            self.recordset, self._job_method.__name__, self._job_args, self._job_kwargs
+            str(self.recordset),
```
sbidoul: nit: str() is redundant here?

queue_job/delay.py
```diff
@@ -493,7 +493,10 @@ def _tail(self):
...
+            self._job_method and self._job_method.__name__ or "",
```
guewen:
```suggestion
            self._job_method.__name__ if self._job_method else "",
```

`... and ... or ...` is a pattern I prefer to avoid

guewen (approved):
Thanks! Would you mind applying my "cosmetic" suggestion?

etobella:
Thanks!!! All comments attended

PR #492, opened 2022-12-24
[14.0][MIG] base_export_async: Migration to 14.0

cuongnmtm:


base_export_async/models/delay_export.py
```diff
@@ -21,6 +19,9 @@ class DelayExport(models.Model):
     _description = "Allow to delay the export"
 
     user_id = fields.Many2one("res.users", string="User", index=True)
```
hieulucky111: no need `string`.

base_export_async/models/delay_export.py
```diff
@@ -101,40 +103,30 @@ def export(self, params):
         )
         date_today = fields.Date.today()
         expiration_date = fields.Date.to_string(
-            date_today + relativedelta(days=+int(time_to_live))
+            date_today + relativedelta(days=int(time_to_live) + 1)
```
hieulucky111: i think we should keep date_today + relativedelta(days=int(time_to_live))

base_export_async/models/delay_export.py
```diff
@@ -101,40 +103,30 @@ def export(self, params):
...
-                <p>It will be automatically deleted the {}.</p>
-                <p>&nbsp;</p>
-                <p><span style="color: #808080;">
-                This is an automated message please do not reply.
-                </span></p>
-                """
-                ).format(url, expiration_date),
-                "auto_delete": True,
-            }
+            },
```
hieulucky111, line 134: should be force_send=True?

base_export_async/security/ir_rule.xml
```diff
@@ -1,13 +1,13 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8" ?>
 <odoo>
```
hieulucky111: should be noupdate=1

hieulucky111 (changes requested):

base_export_async/models/delay_export.py
cuongnmtm: We should plus one day for the expiration date, otherwise, the record might be expired unexpectedly.

For example, I set time_to_live as 7 days, I expected the record to be available until the end of the 7th day.

base_export_async/models/delay_export.py
cuongnmtm: I am not sure why we need force_send here.

mmequignon:
Hello @cuongnmtm, thanks for this!
Looks like we are missing the git history here.
To fix that, you can either do the migration by hand, following those steps https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0
Or use the oca-port tool that you can find here https://github.com/OCA/oca-port

cuongnmtm:
@mmequignon Thanks for your great information. This `base_export_async` module is copied from `12.0` to `14.0`, it is not installable since `13.0`. I see the commit history of the module is already there, so I think I don't need to add it to this PR, right?
However, I could apply your suggestion for https://github.com/OCA/queue/pull/493.

base_export_async/data/config_parameter.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
```
mmequignon, line 1: There's a lot of code format changes in this commit.
Please, move them in a dedicated "black,  isort, prettier" commit.
Those doesn't seems to be related to copier.
Or am I mistaking?

base_export_async/models/delay_export.py
mmequignon: For this kind of change, could you please add a test that fails without it?
It would be easier to understand what you are trying to fix/improve.

base_export_async/models/delay_export.py
vincent-cowboy: [`:param bool force_send: send email immediately; otherwise use the mail queue (recommended);`](https://github.com/odoo/odoo/blob/14.0/addons/mail/models/mail_template.py#L247)

base_export_async/models/delay_export.py
cuongnmtm: I think this improvement should not be placed in this migration PR. I reverted the related changes.

base_export_async/data/config_parameter.xml
cuongnmtm: I fixed it by getting the commit history from 12.0 instead of 13.0

vincent-cowboy (approved):
LGTM, thank you for the PR

guewen (approved):

cuongnmtm:
@guewen I have just rebased 14.0 to my branch. Hopefully, it will solve the failed check(s).

guewen:
Thanks!

/ocabot merge nobump

PR #496, opened 2023-01-03
[14.0][IMP] queue_job: do not list jobs in cancel modale

hparfr:
Because, if the system is saturated you may need to quickly
cancel lot of jobs. Add a list here may be slow to generate.

Having a recap here do not add value if there is more than few items.

### DIFF
--- a/queue_job/wizards/queue_jobs_to_cancelled_views.xml
+++ b/queue_job/wizards/queue_jobs_to_cancelled_views.xml
@@ -6,9 +6,7 @@
         <field name="model">queue.jobs.to.cancelled</field>
         <field name="arch" type="xml">
             <form>
-                <group string="The selected jobs will be cancelled.">
-                    <field name="job_ids" nolabel="1" />
-                </group>
+                <span>Some jobs will be cancelled</span>
                 <footer>
                     <button
                         name="set_cancelled"
--- a/queue_job/wizards/queue_jobs_to_cancelled_views.xml
+++ b/queue_job/wizards/queue_jobs_to_cancelled_views.xml
@@ -6,7 +6,7 @@
         <field name="model">queue.jobs.to.cancelled</field>
         <field name="arch" type="xml">
             <form>
-                <span>Some jobs will be cancelled</span>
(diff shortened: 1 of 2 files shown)


sebastienbeau (approved):
LGTM (code review)
showing the list of job is not necessary and can take time for nothing

queue_job/wizards/queue_jobs_to_cancelled_views.xml
```diff
@@ -6,9 +6,7 @@
         <field name="model">queue.jobs.to.cancelled</field>
         <field name="arch" type="xml">
             <form>
-                <group string="The selected jobs will be cancelled.">
-                    <field name="job_ids" nolabel="1" />
-                </group>
+                <span>Some jobs will be cancelled</span>
```
guewen:
```suggestion
                <span>Selected jobs will be cancelled</span>
```

StefanRijnhart:
I would like to see a cutoff point here, where you show the jobs if there are less than 20 or so, and otherwise show the number of jobs to cancel to the user instead. What do you think about that?

hparfr:
@StefanRijnhart The tree view of jobs is not enough ?

StefanRijnhart:
To intervene with the statuses of jobs is always critical, I feel. So as a user I'd be happy to get as much feedback about what I'm actually doing as possible. If anything, if the user clicks the 'select all' checkbox it would be nice to show that this will actually cancel more than just the first 80 jobs on the first page of that tree view.

But it's just a suggestion.

hparfr:
>if the user clicks the 'select all' checkbox it would be nice to show that this will actually cancel more than just the first 80 jobs on the first page of that tree view.

It's already done in web module: 
image

StefanRijnhart:
I know, it's just my preference to see this information in every step in case at some point, in some migration or in some customization someone messes up their processing of the active ids.

hparfr:
So, @StefanRijnhart @guewen , how can we advance on this PR  ?

StefanRijnhart:
Would you agree to showing the number of jobs (computed as `len(job_ids)` in the wizard?

PR #504, opened 2023-02-01
[16.0][FIX] Patch Request._get_session_and_dbname to get db from URL

### LINKED ISSUE #503: [V16-queue_job] Problems with multiple databases
Currently the `queue_job` module is not running jobs when multiple databases are available.

It returns errors when executing jobs because it is not able to get a proper environment for the database as Odoo is providing the response without the associated database.

The error occurs in the file `queue/queue_job/controllers/main.py`, line 75, in `runjob`

```python
env = http.request.env(user=SUPERUSER_ID)
TypeError: 'NoneType' object is not callable
```

After this error, the job is re-queued and set to pending status.

hugosantosred:
In servers with multiple databases db is not captured from URL when processing jobs in /queue_job/runjob

With this patch db is set from the db param in URL when calling the runjob endpoint

Fixes #503

---

Maybe this monkey patch is not the most elegant or efficient solution, but this is the only way that I've managed to set the right db in order to call the petition with the `request._serve_db` method after the http.py complete refactor in v16.

### DIFF
--- a/queue_job/__init__.py
+++ b/queue_job/__init__.py
@@ -4,6 +4,7 @@ from . import models
 from . import wizards
 from . import jobrunner
 from .post_init_hook import post_init_hook
+from .post_load import post_load
 
 # shortcuts
 from .job import identity_exact
--- a/queue_job/__manifest__.py
+++ b/queue_job/__manifest__.py
@@ -31,4 +31,5 @@
     "development_status": "Mature",
     "maintainers": ["guewen"],
     "post_init_hook": "post_init_hook",
+    "post_load": "post_load",
 }
--- a/queue_job/post_load.py
+++ b/queue_job/post_load.py
@@ -1,0 +1,25 @@
+import logging
+
+from odoo import http
+
+_logger = logging.getLogger(__name__)
+
+
+def post_load():
+    _logger.info(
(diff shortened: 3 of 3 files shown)


hugosantosred:
Pre-commit is failing because of isort problem. Fixed by updating copier template in #505

StefanRijnhart:
Thanks! I agree that monkeypatching should be avoided if possible. Is the `db` argument populated in https://github.com/OCA/queue/blob/94bd236/queue_job/controllers/main.py#L74? In that case, maybe it helps to insert a line there that also assigns the `db` attribute to the request.

hugosantosred:
@StefanRijnhart At that point, the `db` argument is populated, but the problem is that the request is already served with the `_serve_nodb` method, because the `request.db` is not populated here: https://github.com/odoo/odoo/blob/16.0/odoo/http.py#L1989

In the requests that are getting served with the _serve_nodb method, there is not env available, and setting the request.db does nothing.

zuher83 (approved):
Work with me 🙂 Thanks

davejames (approved):
functional review, works well here in a multi-db scenario

guewen (approved):
Thanks for the fix!
Also, thanks for checking if avoiding a monkey patch was possible. Sadly not, but at least it is applied very narrowly.

PR #505, opened 2023-02-01
[16.0][FIX] Update copier template

hugosantosred:
An update is needed because of isort library problems with pre-commit - PyCQA/isort#2077


StefanRijnhart (approved):
Thanks!

PR #508, opened 2023-02-14
[FIX] queue_job: Commit was removed

etobella:
#507


pedrobaeza (changes requested):
This is not correct. You should never commit in any regular transaction. Why is it a problem?

guewen (approved):
Forward port of #507 
Thanks

PR #509, opened 2023-02-14
[FIX] queue_job: Commit was removed

etobella:
#507


pedrobaeza (changes requested):
This is not correct,as stated in the other PRs. Put a limit on the searched records instead.

guewen (approved):
Thanks

PR #511, opened 2023-02-22
[14.0][IMP] queue_job: new app icon

ypapouin:
To be visually more similar to other Odoo app icons.

This is a forward port of https://github.com/OCA/queue/pull/294


pedrobaeza:
From 14.0, there's a glow effect on the corners, so it's not exactly the expected one. You can use https://ivantodorovich.github.io/odoo-icon/ from your colleague to get it easily.

PR #516, opened 2023-02-28
[16.0][FWD][FIX] queue_job: fix Delayable.__repr__

simahawk:
It seems that sometimes jobs doesn't have a job_method and the __repr__ failed

FWD port of https://github.com/OCA/queue/pull/440
With better commit msg.

As the commit msg won't match I'd blacklist that PR for oca-port.

No need to backport to v15.

### DIFF
--- a/queue_job/delay.py
+++ b/queue_job/delay.py
@@ -492,8 +492,11 @@ class Delayable:
         return [self]
 
     def __repr__(self):
+        job_method = ""
+        if self._job_method:
+            job_method = self._job_method.__name__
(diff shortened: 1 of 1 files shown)


sebalix (approved):

simahawk:
I'll fwd port changes from https://github.com/OCA/queue/pull/519 before merging this, also because it seems this change was integrated in the chain dependencies mgmt.

simahawk:
it was already in v15 here https://github.com/OCA/queue/pull/487 I'll port that change w/ the others.

PR #433, opened 2022-05-24
[14.0] Add storage of context in jobs recordsets

AshishHirapara:
This is the migration of the PR https://github.com/OCA/queue/pull/432 (by @guewen) to 14.0

PR #518, opened 2023-02-28
[15.0][FW] [14.0] Add storage of context in jobs recordsets

simahawk:
Port of #433 from 14.0 to 15.0.


simahawk:
included in https://github.com/OCA/queue/pull/519

PR #521, opened 2023-02-28
[16.0] queue_job: unify no delay option

simahawk:
Current situation:

* multiple keys for no good reason
* half baked: not all of them used everywhere
* no centralization
* poor naming

With this change we'll have:

* 1 and only one key to disable via ctx: ``queue_job__no_delay``
* 1 and only one key to disable via os env: ``QUEUE_JOB__NO_DELAY``
* backward compatibility with deprecation for old keys

### DIFF
--- a/queue_job/README.rst
+++ b/queue_job/README.rst
@@ -388,13 +388,14 @@ be customized in ``Base._job_prepare_context_before_enqueue_keys``.
 When you are developing (ie: connector modules) you might want
 to bypass the queue job and run your code immediately.
 
-To do so you can set `TEST_QUEUE_JOB_NO_DELAY=1` in your enviroment.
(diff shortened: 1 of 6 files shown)


grindtildeath (commented):
Looks good overall. Is there any reason to use a double underscore in the name of ctx key and env var?
IMO it would be better to have a single underscore as in `QUEUE_JOB_NO_DELAY` to avoid potential troubles because it's not something that's easy to detect if you accidentally write only one.

simahawk:
Actually the env var has 2 underscores too. TBH this is just an approach I'm taking to identify keys by module especially if the are meant to be used in many many modules. So the `__` is just to separate the name explicitly.
I can change that of course.

guewen (approved):

theangryangel (approved):

PR #528, opened 2023-04-14
[16.0] [FIX] base_import_async: remove default values from context when creating the attachment

marielejeune:
Some default values might be present in context depending on the action we came from when clicking on 'import' button. These default values are not intended to be default values for the ir.attachment record.
In some cases they cause an error because a field with the same name exists on ir.attachment, as for e.g. the `'default_type'='opportunity'` value present in the standard crm.lead action context.

### DIFF
--- a/base_import_async/models/base_import_import.py
+++ b/base_import_async/models/base_import_import.py
@@ -88,10 +88,17 @@ class BaseImportImport(models.TransientModel):
         writer.writerow(fields)
         for row in data:
             writer.writerow(row)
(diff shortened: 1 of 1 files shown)


rousseldenis:
@OCA/core-maintainers @guewen Could you reopen this and review ?

guewen (approved):

marielejeune:
Rebased on 16.0 + fixed pre-commit.

guewen:
Thanks!

PR #537, opened 2023-05-09
[14.0] Fix trap_jobs() recordset comparison

guewen:
The trap_jobs() test helper accepts any model/recordset as long as the job function is the same, which is not enough to verify that the job has been delayed on the expected recordset or model.

Tested job functions are "bound methods" so we can (and should) compare their `__self__`.

Example of assertion that would be valid:

    partner = self.env["res.partner"].create({"name": "foo"})
    with trap_jobs() as trap:
      partner.with_delay().do_something(42)
      trap.assert_enqueued_job(
          self.env["res.partner"].do_something,
          args=(42,)
      )

Now, it would fail with an error message like:

    E           AssertionError: Job <res.partner()>.do_something(42) with properties (channel=root.test, description=Do something, eta=15, identity_key=<function identity_exact at 0x7f01bfb54cb0>, max_retries=1, priority=15) was not enqueued.
    E           Actual enqueued jobs:
    E            * <res.partner(64,)>.do_something(42) with properties (priority=15, max_retries=1, eta=15, description=Do something, channel=root.test, identity_key=<function identity_exact at 0x7f01bfb54cb0>)

### DIFF
--- a/queue_job/tests/common.py
+++ b/queue_job/tests/common.py
@@ -111,7 +111,8 @@ class JobCall(typing.NamedTuple):
         if not isinstance(other, JobCall):
             return NotImplemented
         return (
-            self.method.__func__ == other.method.__func__
+            self.method.__self__ == other.method.__self__
+            and self.method.__func__ == other.method.__func__
             and self.args == other.args
             and self.kwargs == other.kwargs
             and self.properties == other.properties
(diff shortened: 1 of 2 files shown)


mtriki (approved):

simahawk (approved):
Seems legit

hparfr (approved):

guewen:
/ocabot merge minor

I choose minor over patch because succeeding tests can now fail (these tests were not "correct" in the first place), there is no breaking change otherwise.

guewen:
Oh, it fails because of https://github.com/odoo/odoo/commit/f0cab1bf847a851b355ff90292e332ba185d2a10

PR #540, opened 2023-05-15
[13.0] [FIX] queue_job_cron: preserve cron user

ThomasBinsfeld:
From https://github.com/OCA/queue/pull/315


simahawk:
I guess you couldn't port the original commit because it's not clean, right?

StefanRijnhart:
Yes, would be nice to preserve the original author in the commit.

petrus-v (approved):
Code looks good to me

PR #542, opened 2023-05-15
[16.0] Fix trap_jobs() recordset comparison

guewen:
Port #537


guewen:
/ocabot merge minor

I choose minor over patch because succeeding tests can now fail (these tests were not "correct" in the first place), there is no breaking change otherwise.

PR #544, opened 2023-06-16
[14.0] [ENH] queue_job: identity_key enhancements

richard-willdooit:

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -300,7 +300,7 @@ class Job(object):
             .search(
                 [
                     ("identity_key", "=", self.identity_key),
-                    ("state", "in", [PENDING, ENQUEUED]),
+                    ("state", "in", [WAIT_DEPENDENCIES, PENDING, ENQUEUED, STARTED]),
                 ],
                 limit=1,
             )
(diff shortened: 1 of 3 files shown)


simahawk (approved):
LGTM, @guewen can you have a look?

queue_job/job.py
```diff
@@ -300,7 +300,7 @@ def job_record_with_same_identity_key(self):
             .search(
                 [
                     ("identity_key", "=", self.identity_key),
-                    ("state", "in", [PENDING, ENQUEUED]),
+                    ("state", "in", [WAIT_DEPENDENCIES, PENDING, ENQUEUED, STARTED]),
```
simahawk, line 303: Wondering: shall we make optional to look for failed jobs too?
Example: I have some jobs to send out files. A cron is responsible to spawn jobs whenever a record needs to send out something. When jobs fail (eg: external service is off) they will get re-created because `FAILED` is not considered here.

simahawk: Actually the right approach seems to be to raise a RetryableJobError that will set the state as pending

simahawk:
@guewen gentle ping :)

guewen:
@simahawk I was still expecting this change: https://github.com/OCA/queue/pull/546#issuecomment-1596537052

richard-willdooit:
@guewen @simahawk Oh dear, guys, many apologies - I will aim to revisit this before the end of the month.

simahawk:
@richard-willdooit I really need to move this fwd. I hope you don't mind if I'm finishing your work on #581

PR #545, opened 2023-06-16
[15.0] [ENH] queue_job: identity_key enhancements

richard-willdooit:

### DIFF
--- a/queue_job/tests/common.py
+++ b/queue_job/tests/common.py
@@ -232,6 +232,7 @@ class JobsTrap:
                 self._perform_graph_jobs(jobs)
             else:
                 self._perform_single_jobs(jobs)
+        self.enqueued_jobs = []
 
     def _perform_single_jobs(self, jobs):
         # we probably don't want to replicate a perfect order here, but at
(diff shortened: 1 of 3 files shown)


queue_job/tests/common.py
```diff
@@ -251,11 +252,14 @@ def _perform_graph_jobs(self, jobs):
...
-        self._store_patchers.append(patcher)
-        patcher.start()
+        if not job.identity_key or all(
+            j.identity_key != job.identity_key for j in self.enqueued_jobs
+        ):
+            self.enqueued_jobs.append(job)
+
+            patcher = mock.patch.object(job, "store")
+            self._store_patchers.append(patcher)
+            patcher.start()
```
abhinavvaidya-tsd, line 262: Suggested change
```python
    with mock.patch.object(job, "store") as patcher:
        self._store_patchers.append(patcher)
```

**Context Manager:** Used the `with` statement to create a context for the mock patcher. This ensures that the patcher's cleanup (`stop()`) is automatically handled even if an exception occurs.

By using the `with` statement, you ensure that the patcher is properly started and stopped, avoiding potential issues with resource leaks or incorrect mocking behavior.

abhinavvaidya-tsd (approved):
Suggestion added

simahawk:
changes to be backported from https://github.com/OCA/queue/pull/587

simahawk:
Finished in https://github.com/OCA/queue/pull/590 as per #587 and #581

PR #557, opened 2023-08-16
[IMP] queue_job_batch: Improve perf of counters

yvaucher:
In the case of a batch of 2k jobs having large task definition a memory error would be reached easily.
Using a read saves some memory by avoiding the pre-fetching.

### DIFF
--- a/queue_job_batch/models/queue_job_batch.py
+++ b/queue_job_batch/models/queue_job_batch.py
@@ -116,18 +116,18 @@ class QueueJobBatch(models.Model):
         })
         return self.sudo().create(vals).sudo(self.env.uid)
 
-    @api.depends('job_ids')
+    @api.depends("job_ids", "job_ids.state")
     def _compute_job_count(self):
         for record in self:
-            job_count = len(record.job_ids)
-            failed_job_count = len(record.job_ids.filtered(
-                lambda r: r.state == 'failed'
-            ))
-            done_job_count = len(record.job_ids.filtered(
-                lambda r: r.state == 'done'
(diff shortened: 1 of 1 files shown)


queue_job_batch/models/queue_job_batch.py
```diff
@@ -116,18 +116,18 @@ def get_new_batch(self, name, **kwargs):
...
-            done_job_count = len(record.job_ids.filtered(
-                lambda r: r.state == 'done'
-            ))
-            record.job_count = job_count
-            record.finished_job_count = done_job_count
-            record.failed_job_count = failed_job_count
-            record.completeness = done_job_count / max(1, job_count)
-            record.failed_percentage = failed_job_count / max(1, job_count)
+            jobs = record.job_ids
+            states = [r["state"] for r in jobs.read(["state"])]
```
simahawk, line 123: any good reason to call read? Otherwise I'd go for
```suggestion
            states = jobs.mapped("state")
```

yvaucher: AFAIK in v12 `mapped` makes a pre-fetch which I want to avoid.

It seems to use the item getter, looping on each record thus IIUC it does a pre-fetch
https://github.com/odoo/odoo/blob/12.0/odoo/models.py#L4916

simahawk: but you are avoiding it by reading it straight from db... I'm not sure to get where is the improvement :thinking:

simahawk (approved):

PR #566, opened 2023-10-20
[14.0] [ADD] queue_job_batch_size

paradoxxxzero:
This module allows to seemlessly split a big job into smaller jobs.

It uses ``queue_job_batch`` to group the created jobs into a batch. 

Example:

```python
      # ...

      def copy_all_partners(self):
          # Duplicate all partners in batches of 30:
          self.with_delay(batch_size=30).copy()

  # ...
  self.env['res.partner'].search([], limit=1000).copy_all_partners()
```

This will create 34 jobs, each one copying 30 partners (except the last one which will copy 10) and will group them into a batch.

Instead of ``batch_size``, one can also use ``batch_count`` to specify the number of batches to create instead.


queue_job_batch_size/models/base.py
```diff
@@ -0,0 +1,132 @@
...
+    def with_delay(
+        self,
+        priority=None,
+        eta=None,
+        max_retries=None,
+        description=None,
+        channel=None,
+        identity_key=None,
+        batch_size=None,
+        batch_count=None,
```
amh-mw, line 110: This seems straightforward enough, but the corresponding change to `def delayable` is missing.

amh-mw (changes requested):
This generally feels like it should be an improvement to queue_job_batch, not its own module. I'm setting up a local test environment today and will report back with observations.

simahawk:
@paradoxxxzero shall we close this one in favor of https://github.com/OCA/queue/pull/658 ?

PR #567, opened 2023-10-24
[15.0][IMP] queue: test performance improvement

josep-tecnativa:
CC @Tecnativa TT45458

### DIFF
--- a/queue_job/tests/test_delayable.py
+++ b/queue_job/tests/test_delayable.py
@@ -8,9 +8,20 @@ from odoo.addons.queue_job.delay import Delayable, DelayableGraph
 
 
 class TestDelayable(unittest.TestCase):
-    def setUp(self):
-        super().setUp()
-        self.recordset = mock.MagicMock(name="recordset")
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        # Remove this variable in v16 and put instead:
+        # from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT
+        DISABLED_MAIL_CONTEXT = {
+            "tracking_disable": True,
+            "mail_create_nolog": True,
+            "mail_create_nosubscribe": True,
+            "mail_notrack": True,
+            "no_reset_password": True,
+        }
+        cls.env = cls.env(context=dict(cls.env.context, **DISABLED_MAIL_CONTEXT))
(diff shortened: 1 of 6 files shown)


chienandalu (approved):

pedrobaeza (approved):

queue_job/tests/test_delayable.py
```diff
@@ -8,9 +8,20 @@
 
 
 class TestDelayable(unittest.TestCase):
-    def setUp(self):
-        super().setUp()
-        self.recordset = mock.MagicMock(name="recordset")
+    @classmethod
+    def setUpClass(cls):
```
simahawk, line 12: IMO this change is useless. This class does not inherit from an odoo class and does not trigger any odoo functionality

chienandalu: Well, it avoids redeclaring the case for every test. Indeed not much gain here anyway and the context thing would be useless here, that's true

simahawk: It's a mocked object... changes nothing in terms of perf.
Anyway, I'd keep the setUpClass but I'd drop the ctx stuff because it gives the false idea that is needed for whatever  reason.

chienandalu: I agree :+1: Please, @josep-tecnativa :)

pedrobaeza: Well, maybe in terms of "consistency" to not wondering if put it or not.

chienandalu: The thing here is that this class test inherits directly from `unittest` so it's quite useless to set that

pedrobaeza: Yes, useless, but consistent to always read the same code everywhere in OCA (I wonder also why this is not using Odoo tests), and with minimum overhead. And the day this is switched to Odoo tests, you already have it.

simahawk: I disagree because you should always keep the setup minimal.
However.... Now that I think of this twice... I fear that since the refactoring of the test suite was done (14, 15, 16) these tests are not auto-discovered anymore (just search for this test class in the build).
Hence, we should use `BaseCase` at least.

pedrobaeza: @josep-tecnativa switch to `BaseCase` or `TransactionCase`, please.

PR #569, opened 2023-10-31
[FIX] Access error on ir.attachment create

hhgabelgaard:
As normal user only have read access to delay.export, we need sudo to chreate an attachment on this record

Use user 6 (Marc Demo) in the tests, not admin (2)


guewen (approved):
Makes sense, thanks!

PR #570, opened 2023-11-03
[14.0][IMP] export_async_schedule : Send attachments directly

quoc-pn:


export_async_schedule/models/export_async_schedule.py
```diff
@@ -45,7 +45,16 @@ class ExportAsyncSchedule(models.Model):
         default=lambda self: self.env.lang,
         help="Exports will be translated in this language.",
     )
-
+    mail_template_id = fields.Many2one(
+        "mail.template",
+        string="Email addresses",
```
cuongnmtm, line 50: Why email addresses?

cuongnmtm (changes requested):

PR #572, opened 2023-11-07
[14.0][IMP] export_async_schedule : Send attachments directly

quoc-pn:
Improvement:

- Add the ability to specify mail template for export.async.schedule.

- Add the ability to attach the export attachment to the email.

These two new features will help if you want to send the export file to someone who doesn't have access to do (so they cannot open the link).

### DIFF
--- a/base_export_async/models/delay_export.py
+++ b/base_export_async/models/delay_export.py
@@ -66,7 +66,9 @@ class DelayExport(models.Model):
             return xls.from_data(columns_headers, import_data)
 
     @api.model
-    def export(self, params):
+    def export(
+        self, params, mail_template=None, is_export_file_attached_to_email=False
+    ):
         """Delayed export of a file sent by email
 
         The ``params`` is a dict of parameters, contains:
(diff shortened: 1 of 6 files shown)


base_export_async/tests/test_base_export_async.py
```diff
@@ -64,22 +64,26 @@ def test_export_csv(self):
         params = json.loads(data_csv.get("data"))
         mails = self.env["mail.mail"].search([])
         attachments = self.env["ir.attachment"].search([])
-        self.delay_export_obj.export(params)
+        self.delay_export_obj.export(
+            params, self.env.ref("base_export_async.delay_export_mail_template"), True
```
cuongnmtm: `self.env.ref("base_export_async.delay_export_mail_template")` is default value for ` mail_template` param, you should write a new test for `is_export_file_attached_to_email=True`

base_export_async/tests/test_base_export_async.py
```diff
@@ -64,22 +64,26 @@ def test_export_csv(self):
...
+        )
         new_mail = self.env["mail.mail"].search([]) - mails
-        new_attachment = self.env["ir.attachment"].search([]) - attachments
+        new_attachments = self.env["ir.attachment"].search([]) - attachments
```
cuongnmtm: there should be only one attachment, isn't it?

cuongnmtm (changes requested):
Please help to add to the description what are the improvements.

- Add ability to specify mail template for export.async.schedule.
- Add ability to attach the export attachment to the email.

These two new features will help in case you want to send the export file to someone who doesn't have access to do (so they cannot open the link).

base_export_async/tests/test_base_export_async.py
quoc-pn: There will be 2 attachments, one is attached to the mail and one for the link in the mail.

cuongnmtm: And they are the same, right?  you created duplicate attachments.

cuongnmtm: I will add a comment on where should you change

base_export_async/models/delay_export.py
```diff
@@ -124,15 +126,24 @@ def export(self, params):
...
-                "recipient_ids": [(6, 0, users.mapped("partner_id").ids)],
-            },
-        )
+        email_values = {
+            "email_from": email_from,
+            "reply_to": email_from,
+            "recipient_ids": [(6, 0, users.mapped("partner_id").ids)],
+        }
+        if is_export_file_attached_to_email:
+            email_values.update({"attachments": [(name, base64.b64encode(content))]})
```
cuongnmtm: instead of updating `attachments`, try to update `attachment_ids`, and reuse the above `attachment` record.
```python
attachment_ids = attachment.ids
if mail_template:
    attachment_ids += mail_template.attachment_ids 
email_values.update({"attachment_ids": mail_template.attachment_ids + attachment.ids})
```

robyf70 (approved):
LGTM

PR #577, opened 2023-11-16
[12.0][IMP] queue_job: track error in chatter

MiquelRForgeFlow:
When job fails, it shows in chatter:

Selection_2215

Later, the job does a retry. If the retry succeeds, the error is deleted. If not succeeds but the error is different, the error changes. In both cases, you cannot know what was the first error. Let's put in the chatter to have better history.


PCatinean (approved):

PCatinean:
I second this, db size should not be an issue with the autovacuum and exception info is indeed quite important

MiquelRForgeFlow:
@simahawk @guewen please! Once merged, I will do forward ports...

etobella (approved):
Technically LGTM

simahawk (approved):
LG. Next time pls, move the pre-commit update to another PR :pray:

PR #571, opened 2023-11-03
[14.0] Port #379 to 14.0

sim6:
Port #379 to 14.0

PR #589, opened 2023-11-22
[16.0] Fix multi databases with list_db = false 

guewen:
Port of https://github.com/OCA/queue/pull/379 , https://github.com/OCA/queue/pull/571, #580

### DIFF
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -401,7 +401,7 @@ class QueueJobRunner(object):
         if config["db_name"]:
             db_names = config["db_name"].split(",")
         else:
-            db_names = odoo.service.db.exp_list(True)
+            db_names = odoo.service.db.list_dbs(True)
         return db_names
 
     def close_databases(self, remove_jobs=True):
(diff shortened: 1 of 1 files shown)


pcastelovigo:
thank you so much

rvalyi (approved):

queue_job/jobrunner/runner.py
```diff
@@ -398,10 +398,16 @@ def from_environ_or_config(cls):
         return runner
 
     def get_db_names(self):
+        """
+        >>> runner = QueueJobRunner()
+        >>> config["db_name"] = None
+        >>> runner.get_db_names()
+        ['odoo']
+        """
```
sim6, line 406: What about add `list_db = False` to test `odoo.service.db.list_dbs(True)` against `odoo.service.db.exp_list(True)`?
```
>>> config["list_db"] = False
```

pcastelovigo: i'm not sure about the implications, as I must confess I never checked what list_dbs or exp_list does

pcastelovigo: oh, i finally understood. Yes, thats a great idea, and a much better test approach
This only checks list_dbs gives something like untested exp_lists, your test checks if it works without list_dbs

sim6: :) Please see https://github.com/OCA/queue/pull/591

PR #594, opened 2023-11-27
[14.0] queue_job: split identity_key hasher to ease reuse

simahawk:
You can now create your own identity exact matcher using the default hasher as a base.


guewen (approved):
Makes sense! Thanks

etobella:
@simahawk Will you port this commits to 15 and  16?

Issue #585, opened 2023-11-21
[16.0] queue_job: Autovacuum doesn't clean up sub channels


yibudak:
To reproduce the error:

1) Add multiple sub-channels under the root channel. 
2) Create jobs past the deadline.
3) Run the "AutoVacuum Job Queue" job again in Scheduled Actions.

Autovacuum will only delete jobs with the "root" channel, it will not clean up jobs with sub-channels like "root.1" or "root.2".

Sample odoo.conf:
```
[queue_job]
channels = root:10, root.0:1, root.1:1, root.2:1, root.3:1, root.4:1, root.5:1, root.6:1, root.7:1, root.8:1, root.9:1
```
https://github.com/OCA/queue/blob/1ab7b4c477c8a71c837ae68d8b6dc7fdfce24f35/queue_job/models/queue_job.py#L407C28-L407C28

yibudak:
@guewen

guewen:
Hi, what do you have in the menu "Queue Job > Channels"? Also, do they match with the content of `queue_job.channel` fields?

yibudak:
![image](https://github.com/OCA/queue/assets/52182279/18db6d4d-4270-4725-9590-b10d287dbdfe)
![image](https://github.com/OCA/queue/assets/52182279/819d05f1-e60e-4c4b-a466-098ae053f0d3)

it's real life example. when we create a channel under a channel, I don't know if it should also appear on the channels screen, in all the work I have done in version 16, only the main channels appear.

guewen:
Interesting, thanks!

So first thing first, if you create corresponding channels in the UI, you will be able to set a number of days for the cleaning of jobs, per channel, and all jobs will eventually be cleaned.

Now, it shows a possible improvement, during cleanup, we could find the configuration of the closest parent channel in `queue.job.channel` _if no `queue.job.channel` exists for the job's channel_. The solution with an ilike would be too broad as when searching for instance for 'root.%'* it would apply its configuration for all subchannels, including those that already have a configuration.

guewen:
> it's real life example. when we create a channel under a channel

Oh but you mean you created a subchannel on the UI and it doesn't appear? That would be quite another issue 😓

yibudak:
I need to mention that I didn't create the sub channels within UI, I've just written them in config file. Aren't they supposed to be created on service initializing?

[...]

PR #596, opened 2023-11-27
[16.0][FIX] queue_job: Autovacuum for code generated channels

yibudak:
ref: https://github.com/OCA/queue/issues/585#issuecomment-1823958753

### DIFF
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -398,13 +398,19 @@ class QueueJob(models.Model):
         """
         for channel in self.env["queue.job.channel"].search([]):
             deadline = datetime.now() - timedelta(days=int(channel.removal_interval))
+            subchannels = self.env["queue.job.channel"].search(
+                [("parent_id", "=", channel.id)]
+            )
             while True:
                 jobs = self.search(
                     [
                         "|",
                         ("date_done", "<=", deadline),
                         ("date_cancelled", "<=", deadline),
+                        "|",
                         ("channel", "=", channel.complete_name),
+                        ("channel", "ilike", channel.complete_name + ".%"),
+                        ("channel", "not in", subchannels.mapped("complete_name")),
                     ],
                     limit=1000,
                 )


queue_job/models/queue_job.py
```diff
@@ -398,13 +398,19 @@ def autovacuum(self):
...
+            )
             while True:
                 jobs = self.search(
                     [
                         "|",
                         ("date_done", "<=", deadline),
                         ("date_cancelled", "<=", deadline),
+                        "|",
                         ("channel", "=", channel.complete_name),
+                        ("channel", "ilike", channel.complete_name + ".%"),
```
guewen, line 412:
```suggestion
                        ("channel", "=like", channel.complete_name + ".%"),
```
Otherwise it will add % before and after.

queue_job/models/queue_job.py
```diff
@@ -398,13 +398,19 @@ def autovacuum(self):
...
+                        ("channel", "not in", subchannels.mapped("complete_name")),
```
guewen, line 413: It would be worth to have a test that covers several levels of hierarchy, ex:

* root (has config: 7 days)
  * root.a (has config: 3 days)
    * root.a.z (has config: 10 days)
    * root.a.y (has no config)
  * root.b (has no config)
    * root.b.z (has no config)
    * root.b.y (has no config)
    * root.b.x (has config: 1 day)
 
When cleaning the channels sequentially, the expectation should be:

1. root: clean root, root.b, root.b.z, root.b.y with 7 days
2. root.a: clean root.a and root.a.y with 3 days
3. root.a.z: clean root.a.z with 10 days
4. root.b.x: clean root.b.x with 1 day

If I'm not mistaken, by the current logic, root.a.z and root.b.x would not respect their 10 days/1 day configuration.

guewen (commented):
Hey, thanks for moving forward on this topic!

iggups:
Any news on this? The deletion of jobs in subchannels is still not working.

PR #597, opened 2023-11-27
[13.0][IMP] queue_job: track error in chatter (forward port)

MiquelRForgeFlow:
Forward port of https://github.com/OCA/queue/pull/577.

PR #598, opened 2023-11-27
[14.0][IMP] queue_job: track error in chatter (forward port)

MiquelRForgeFlow:
Forward port of https://github.com/OCA/queue/pull/597.


pedrobaeza:
Isn't that message very long to be put there?

MiquelRForgeFlow:
Not necessarily.It is not the full traceback, just the error message.

pedrobaeza (approved):
OK, although I'm not very sure it will be helpful without the full trace if the error is for example just `ValidationError`.

PCatinean (approved):

guewen (approved):

Issue #608, opened 2023-12-14
The latest test change breaks tests on Odoo.sh


ventor-dev:
## Module
Queue Job (queue_job)

## Describe the bug
This commit leads to the errors with tests on Odoo.sh (check the screenshot below):

![image](https://github.com/OCA/queue/assets/10652233/785fa3ff-a924-4609-9fcb-83ced60e2f42)

<img width="1254" alt="image" src="https://github.com/OCA/queue/assets/10652233/7c14ad2e-afa9-42ee-ac82-8b85cb774522">

Looks like this can be easily fixed by removing doctests from this method (or at least removing new line added).

## To Reproduce
**Affected versions**: 15.0.2.5.5, 16.0.2.5.5

1. Create new project on Odoo.sh
2. Add queue_job module
3. Wait until tests will be finished

**Expected behavior**
No errors in tests

AlexanderPashuk:
Prepared two PRs with hotfix for this issue. I suggest removing this doctest because it makes no sense (at least for me it is not obvious why there should be database with name "odoo" as result of this code).

I am willing to refactor this test case if needed but need more details about what we want to check while testing?

AlexanderPashuk:
Hi all,

Prepared the same PR for 17.0 branch: https://github.com/OCA/queue/issues/608

@gurneyalex could you please merge them? They are the same as were done for 15.0 / 16.0

Thanks!

PR #610, opened 2023-12-14
[15.0] Remove doctest from get_db_names method to avoid issues with Odoo.sh

AlexanderPashuk:
This is the hotfix for this issue: https://github.com/OCA/queue/issues/608


sbidoul (approved):
I agree tests should not rely on a specific database name. So ok to remove this test for now (it was added recently in https://github.com/OCA/queue/pull/580).

gurneyalex (approved):

PR #479, opened 2022-10-31
[14.0][PORT] queue_job_cron_jobrunner: backport from 15.0 to 14.0 for queue_job_cron_jobrunner from 15.0 to 14.0

chrisb-c01:
This is a backport of the `queue_job_cron_jobrunner` added in 15.0 branch to 14.0.

Because 15.0 uses `ir.cron.trigger` to schedule cron jobs through the `cron._trigger(at)` private method, in this backport I am 'just' setting the `cron.nextcall` attribute to the current time `datetime.now()` in `[queue.job]._cron_trigger()`.

PR #611, opened 2023-12-19
[14.0] queue_job_cron_jobrunner:  avoid concurrent access on ir.cron

petrus-v:
As odoo add a lock while running cron task we should only try to write
on it. If the cron is running while create new queue job records
we will wait next call.

### DIFF
--- a/queue_job_cron_jobrunner/models/queue_job.py
+++ b/queue_job_cron_jobrunner/models/queue_job.py
@@ -150,7 +150,7 @@ class QueueJob(models.Model):
                 nextcall = at
         for cron in crons:
             if nextcall < cron.nextcall:
-                cron.nextcall = nextcall
+                cron.try_write({"nextcall": nextcall})
 
     def _ensure_cron_trigger(self):
         """Create cron triggers for these jobs"""
--- a/queue_job_cron_jobrunner/tests/test_queue_job.py
+++ b/queue_job_cron_jobrunner/tests/test_queue_job.py
@@ -2,8 +2,11 @@
 # @author Iván Todorovich <ivan.todorovich@camptocamp.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+from datetime import datetime
+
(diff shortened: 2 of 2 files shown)


damdam-s (approved):

petrus-v:
@ivantodorovich did you get similar issue on your odoo instances ?

ivantodorovich:
> @ivantodorovich did you get similar issue on your odoo instances ?

Hey @petrus-v , @damdam-s 👋🏻  it's nice to see your names here

Actually, I originally developed this module for Odoo 15.0, relying on cron triggers.
It was later backported to 14.0 by @chrisb-c01 (https://github.com/OCA/queue/pull/479). Instead of using cron triggers, it uses this `nextcall` field.

I've never used this module in `14.0`, though 😓

ivantodorovich (approved):
It makes sense 👍🏻

petrus-v:
> > @ivantodorovich did you get similar issue on your odoo instances ?
> 
> Hey @petrus-v , @damdam-s 👋🏻 it's nice to see your names here
> 
> Actually, I originally developed this module for Odoo 15.0, relying on cron triggers. It was later backported to 14.0 by @chrisb-c01 (#479). Instead of using cron triggers, it uses this `nextcall` field.
> 
> I've never used this module in `14.0`, though 😓

Thanks for your prompt reply, history regarding this topic and review !

Always a pleasure to see your name on so many projects around ! see you soon and happy end of year :)

gurneyalex:
are you aware of https://github.com/OCA/queue/pull/612?

petrus-v:
> are you aware of #612?

no I'm not... but I'm not sure how much it's related. As mentionned by @ivantodorovich this module in version 14.0 do not rely on `ir.cron.trigger` which exists from version 15.0.

So creating a new job in this version try to write on the ir.cron which is probably running to the row is lock, at the end the whole transaction get rollback before this PR.

guewen (approved):
/ocabot merge patch

PR #612, opened 2023-12-21
[16.0][IMP] queue_job_cron: Avoid parallel run

lmignon:
By default, odoo never runs the same cron job in parallel. This commit uses the identity key mechanism to enforce this mechanism when a cron job is run as a queue job. This behaviour can be controlled by a new setting on the cron definition but is activated by default to keep the original behaviour


rousseldenis (approved):
LGTM

sbidoul (approved):

gurneyalex (approved):

PR #613, opened 2023-12-21
[IMP] Add queue_job.keep_context ir.config_parameter

amh-mw:
When this parameter is set, queue_job always preserves the entire context. This honors the principle of least surprise, in that a developer can easily convert a record.method() call to record.with_delay().method() with the expectation that it will actually execute the same, simply at a later time.

Fixes #406.

### DIFF
--- a/queue_job/fields.py
+++ b/queue_job/fields.py
@@ -70,6 +70,8 @@ class JobEncoder(json.JSONEncoder):
     """Encode Odoo recordsets so that we can later recompose them"""
 
     def _get_record_context(self, obj):
+        if obj.env["ir.config_parameter"].sudo().get_param("queue_job.keep_context"):
+            return obj.env.context
         return obj._job_prepare_context_before_enqueue()
 
     def default(self, obj):
--- a/queue_job/tests/test_json_field.py
+++ b/queue_job/tests/test_json_field.py
@@ -31,6 +31,23 @@ class TestJson(common.TransactionCase):
         }
         self.assertEqual(json.loads(value_json), expected)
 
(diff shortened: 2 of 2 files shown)


amh-mw:
Just for general amusement, abridged from code that I wrote to import over a decade of users and data from a legacy system:

```python
self.with_context({
    'bus_no_notify': True, # custom
    'calendar_event_create_no_constrain': True, # custom
    'mail_auto_subscribe_no_notify': True, # odoo
    'mail_create_nolog': True, # odoo
    'mail_create_nosubscribe': True, # odoo
    'mail_notrack': True, # odoo
    'no_geo_complete': True, # custom
    'no_mail_to_attendees': True, # odoo
    'no_reset_password': True, # odoo
    'quickbooks_no_prepare': True, # custom
    'tracking_disable': True # odoo
}).with_delay(...)._cron_migrate()
```

It is really important that context is preserved or all kinds of side effects are triggered.

amh-mw:
@acsonefho @florian-dacosta Any interest in reviewing this `queue_job.keep_context` pull request?

sbidoul:
@amh-mw If my memory is correct, you should rather port https://github.com/OCA/queue/tree/14.0/queue_job_context from 14 to 15. In 16 it is not necessary anymore as it is the default.

amh-mw:
It looks like queue_job_context uses a hardcoded list of context keys that I would have to inherit the class and override?

https://github.com/OCA/queue/blob/089eb3ef9a9c9210384711b6e107433138f9809c/queue_job_context/models/base.py#L12

I don't prefer that. I just want to flip a switch and get all the context.

PR #616, opened 2024-01-18
[FIX]Changes in field job_directed_graph updates from 16 to 17

soos-cpu:


guewen:
Hi @soos-cpu, thanks for the correction! Pre-commit reports a warning and some lint changes, could you check?

PR #621, opened 2024-01-24
[16.0][Fix] ChannelManager.get_channel_by_name create wrong subchannels when pass parent_fallback=True

lingxiezfy:
```python
c = cm.get_channel_by_name('root.sub.not.configured', parent_fallback=True)

# c.fullname should be "root.sub.not.configured", instand of "root.sub.sub.not.configured"
assert c.fullname == "root.sub.not.configured"
```

### DIFF
--- a/queue_job/jobrunner/channels.py
+++ b/queue_job/jobrunner/channels.py
@@ -984,7 +984,7 @@ class ChannelManager(object):
         'root.sub'
         >>> c = cm.get_channel_by_name('root.sub.not.configured', parent_fallback=True)
         >>> c.fullname
-        'root.sub.sub.not.configured'
+        'root.sub.not.configured'
         """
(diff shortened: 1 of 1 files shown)


queue_job/jobrunner/channels.py
```diff
@@ -1012,7 +1012,9 @@ def get_channel_by_name(
                         parent_name,
                     )
                     break
-        for subchannel_name in channel_name.split(".")[1:]:
+        for subchannel_name in channel_name.split(".")[
+            parent.fullname.count(".") + 1 :
+        ]:
```
amh-mw, line 1017: Why is this necessary? Related unit tests could be illustrative.

amh-mw (changes requested):

PR #348, opened 2021-05-14
[ADD] [14.0] randomize retry interval

PieterPaulussen:
A randomized retry interval pattern can have the added benefit of solving concurrent transaction collisions on a database.

PR #624, opened 2024-02-01
Fix for randomized retry interval format

InigoGR:
queue_job: fix retry format with tuple values

https://github.com/OCA/queue/pull/348 added support for randomized retry intervals.
I could not set randomized retry intervals due to the formatting checks not being updated. This should fix it.

### DIFF
--- a/queue_job/models/queue_job_function.py
+++ b/queue_job/models/queue_job_function.py
@@ -149,10 +149,12 @@ class QueueJobFunction(models.Model):
         try:
             # as json can't have integers as keys and the field is stored
             # as json, convert back to int
-            retry_pattern = {
-                int(try_count): postpone_seconds
-                for try_count, postpone_seconds in self.retry_pattern.items()
-            }
+            retry_pattern = {}
+            for try_count, postpone_value in self.retry_pattern.items():
+                if isinstance(postpone_value, int):
+                    retry_pattern[int(try_count)] = postpone_value
+                else:
+                    retry_pattern[int(try_count)] = tuple(postpone_value)
         except ValueError:
             _logger.error(
                 "Invalid retry pattern for job function %s,"
@@ -181,8 +183,9 @@ class QueueJobFunction(models.Model):
     def _retry_pattern_format_error_message(self):
         return _(
             "Unexpected format of Retry Pattern for {}.\n"
-            "Example of valid format:\n"
-            "{{1: 300, 5: 600, 10: 1200, 15: 3000}}"
+            "Example of valid formats:\n"
+            "{{1: 300, 5: 600, 10: 1200, 15: 3000}}\n"
+            "{{1: (1, 10), 5: (11, 20), 10: (21, 30), 15: (100, 300)}}"
         ).format(self.name)
 
     @api.constrains("retry_pattern")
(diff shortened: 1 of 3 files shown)


queue_job/models/queue_job_function.py
```diff
@@ -149,10 +149,12 @@ def _parse_retry_pattern(self):
...
-            retry_pattern = {
-                int(try_count): postpone_seconds
-                for try_count, postpone_seconds in self.retry_pattern.items()
-            }
+            retry_pattern = {}
+            for try_count, postpone_value in self.retry_pattern.items():
+                if isinstance(postpone_value, int):
+                    retry_pattern[int(try_count)] = postpone_value
+                else:
+                    retry_pattern[int(try_count)] = tuple(postpone_value)
```
simahawk, line 157: is this conversion to tuple necessary?

simahawk (changes requested):
Thanks for your contrib!
I guess it failed because we didn't have specific tests, right?
Any chance you add that? If you don't have time that's fine, but it's a really nice to have if you can :wink: 

Also, could you please rewrite the commit msg to something like:

```
queue_job: fix retry format with tuple values

#348 added support for randomized retry intervals.
I could not set randomized retry intervals due to the formatting checks not being updated. This should fix it.
```
What changes: 
1. includes module name
2. includes the full description of the issue

queue_job/models/queue_job_function.py
InigoGR: If I leave that conversion out, the result in the computed value that is show in the `queue_job_function` form is a list. I believe it is a bit confusing if you write a tuple in the config and it results in a list, so I added that.

InigoGR:
> Thanks for your contrib! I guess it failed because we didn't have specific tests, right? Any chance you add that? If you don't have time that's fine, but it's a really nice to have if you can 😉
> 
> Also, could you please rewrite the commit msg to something like:
> 
> ```
> queue_job: fix retry format with tuple values
> 
> #348 added support for randomized retry intervals.
> I could not set randomized retry intervals due to the formatting checks not being updated. This should fix it.
> ```
> 
> What changes:
> 
> 1. includes module name
> 2. includes the full description of the issue

I will add test coverage for that case and change the commit message.

InigoGR:
The test coverage is added and the test are passing.
I think I messed up the commits when trying to change the message (not very experienced with git). @simahawk could you help me with that?

InigoGR:
Added some simple tests that should cover the changes in `_check_retry_pattern` and `_parse_retry_pattern`.

StefanRijnhart (approved):

queue_job/models/queue_job_function.py
```diff
@@ -195,12 +198,20 @@ def _check_retry_pattern(self):
...
                         record._retry_pattern_format_error_message()
                     )
 
+    def _retry_value_type_check(self, value):
+        if isinstance(value, (tuple, list)):
+            if len(value) != 2:
+                raise ValueError
+            [self._retry_value_type_check(element) for element in value]
+            return
+        int(value)
```
simahawk, line 213:
```suggestion
                except ValueError as exc:
                    raise exceptions.UserError(
                        record._retry_pattern_format_error_message()
                    ) from exc

    def _retry_value_type_check(self, value):
        if isinstance(value, (tuple, list)):
            if len(value) != 2:
                raise ValueError("Random retry pattern must be a tuple with 2 digits")
            [self._retry_value_type_check(element) for element in value]
            return
        int(value)
```

simahawk (approved):
Minor remark get a more clear exception. Pre-approving

dreispt:
@simahawk Is this good to merge?

amh-mw (approved):

PR #634, opened 2024-03-11
[13.0][FIX] base_import_async: add import context key as done in base_import

AnizR:
The override of the method `do` is missing some context keys introduced when importing a file (by [base_import](https://github.com/odoo/odoo/blob/13.0/addons/base_import/models/base_import.py#L923)) in "asynchronous mode".

This PR introduces those keys in the job context.

For example:
When importing a product and changing its `standard_price`, Odoo will trigger the function "_change_standard_price" ([code](https://github.com/odoo/odoo/blob/4cc086d330b73514fbc64a7fcf22d8a7a9f1b691/addons/stock_account/models/product.py#L188)). But, doing the same operation asynchronously won't have the same effect since it is missing the context key "import_file" .


acsonefho (approved):

guewen (approved):
/ocabot merge patch

PR #635, opened 2024-03-13
[17.0] Remove doctest from get_db_names method to avoid issues with Odoo.sh

AlexanderPashuk:
PR for https://github.com/OCA/queue/issues/608


ventor-dev (approved):

simahawk (approved):
makes sense even if we should have a unit tests to replace this

CC @guewen

PR #639, opened 2024-03-27
[16.0][FW] [14.0] queue_job: triggers stored computed fields before calling 'set_done()'

sebalix:
Port of #637 from 14.0 to 16.0.


sebalix:
cc @simahawk

simahawk (approved):

PR #644, opened 2024-04-17
[16.0][FIX] prevent a warning message

benwillig:
Use args[0] instead of name to get the exception message as the name attribute is deprecated

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -160,7 +160,7 @@ class RunJobController(http.Controller):
         exception_name = orig_exception.__class__.__name__
         if hasattr(orig_exception, "__module__"):
             exception_name = orig_exception.__module__ + "." + exception_name
-        exc_message = getattr(orig_exception, "name", str(orig_exception))
+        exc_message = orig_exception.args[0] if orig_exception.args else str(orig_exception)
         return {
             "exc_info": traceback_txt,
             "exc_name": exception_name,


ThomasBinsfeld (approved):

FrancoMaxime (approved):

queue_job/controllers/main.py
```diff
@@ -160,7 +160,7 @@ def _get_failure_values(self, job, traceback_txt, orig_exception):
         exception_name = orig_exception.__class__.__name__
         if hasattr(orig_exception, "__module__"):
             exception_name = orig_exception.__module__ + "." + exception_name
-        exc_message = getattr(orig_exception, "name", str(orig_exception))
+        exc_message = getattr(orig_exception, "args[0]", str(orig_exception))
```
simahawk: Are you sure the this instance has an attribute called `"args[0]"`? This seems broken to me :thinking: 
`args` will be always there and very likely you'll have a first arg.

I guess you want to have something like `orig_exception.args[0] if orig_exception.args else str(orig_exception)`.

simahawk (changes requested):
Additionally, could you pls rewrite the commit msg to something like 
```
[fix] queue_job: fix exception msg handling

.... long explanation goes here...
```

queue_job/controllers/main.py
amh-mw: Per 8.3 Handling Exceptions [^1],

> The except clause may specify a variable after the exception name. The variable is bound to the exception instance which typically has an args attribute that stores the arguments. For convenience, builtin exception types define [__str__()](https://docs.python.org/3/reference/datamodel.html#object.__str__) to print all the arguments without explicitly accessing .args.

"Typically" suggests that `args` will be absent some of the time.

1. At the risk of having a slightly uglier error message, why not just use `str(orig_exception)` instead of trying to preserve formatting?
2. If the exception handling behavior is changing, it should be covered by unit tests.

[^1]: https://docs.python.org/3/tutorial/errors.html#handling-exceptions

benwillig: Indeed the syntax looks weird so your proposal would be better. However, @amh-mw made a point so I think it would be better to simply use ```str(orig_exception)```.

simahawk: Maybe I can give you a little background as I'm the author of this piece of code :wink: 

The aim was to clearly rip out of the msg the exception class. That's why we have `exc_name` and `exc_message`. Exceptions are already messy for end users and this is an attempt to have a better msg.
Not to mention that this proposal will cause duplication of info because the exc name will be repeated.
Hence, calling `str` was just a safe fallback to make sure you had a msg.

An alternative is to call `str` and drop surrounding chars like exc name, parentesis, etc but I find this more complex than the little check here.

benwillig: I updated the commit

PR #652, opened 2024-05-19
[FIX] `JobsTrap` `perform_enqueued_jobs` not working with jobs creating new jobs (issue OCA#651)

fd-oncodna:
`JobsTrap` `perform_enqueued_jobs` not working with jobs creating new jobs. After a call to `JobsTrap.perform_enqueued_jobs` the `enqueued_jobs` attribute is empty, even if one of the jobs executed had the effect of creating new jobs. That is because after executing the jobs of the queue in a loop, `JobsTrap.perform_enqueued_jobs` sets the `enqueued_jobs` to `[]`ignoring the newly created jobs.

### DIFF
--- a/queue_job/tests/common.py
+++ b/queue_job/tests/common.py
@@ -228,12 +228,12 @@ class JobsTrap:
             return job.graph_uuid or ""
 
         sorted_jobs = sorted(self.enqueued_jobs, key=by_graph)
+        self.enqueued_jobs = []
         for graph_uuid, jobs in groupby(sorted_jobs, key=by_graph):
             if graph_uuid:
                 self._perform_graph_jobs(jobs)
             else:
                 self._perform_single_jobs(jobs)
-        self.enqueued_jobs = []
 
     def _perform_single_jobs(self, jobs):
         # we probably don't want to replicate a perfect order here, but at


guewen (approved):
Makes sense, thanks

queue_job/tests/common.py
```diff
@@ -227,13 +227,13 @@ def perform_enqueued_jobs(self):
         def by_graph(job):
             return job.graph_uuid or ""
 
-        sorted_jobs = sorted(self.enqueued_jobs, key=by_graph)
+        sorted_jobs = sorted(self.enqueued_jobs[:], key=by_graph)
```
amh-mw: I don't believe the extra [:] slicing here is necessary.

amh-mw (changes requested):

queue_job/tests/common.py
simahawk: I agree, since `sorted` returns already a new list.

simahawk (approved):
Thanks for your contrib! 
Could you please rewrite the commit msg to something like
```
[fix] queue_job: JobsTrap.perform_enqueued_jobs

Closes #651
```
pre-approving

fd-oncodna:
@simahawk everything's done ;)

simahawk:
@fd-oncodna tnx, any chance you can backoprt this to 15 and 16?

fd-oncodna:
> @fd-oncodna tnx, any chance you can backoprt this to 15 and 16?

Yes, see #655 and #656

PR #653, opened 2024-05-23
[14.0][IMP] queue_job: add cron to purge dead jobs.

hparfr:
I'm trying to improve the remediation of started job stucked.

Cron Jobs Garbage Collector, with the second parameter, let you requeue started job. But if the root cause like a CPU limit error is still present after the requeue, the issue always reappear.

With this new cron, the job is marked as failed and not requeued.

### DIFF
--- a/queue_job/data/queue_data.xml
+++ b/queue_job/data/queue_data.xml
@@ -10,6 +10,15 @@
             <field name="state">code</field>
             <field name="code">model.requeue_stuck_jobs()</field>
         </record>
+        <record id="ir_cron_queue_job_fail_dead_jobs()" model="ir.cron">
+            <field name="name">Jobs Garbage Collector</field>
+            <field name="interval_number">15</field>
+            <field name="interval_type">minutes</field>
+            <field name="numbercall">-1</field>
+            <field ref="model_queue_job" name="model_id" />
+            <field name="state">code</field>
+            <field name="code">model.fail_dead_jobs(240)</field>
+        </record>
         <!-- Queue-job-related subtypes for messaging / Chatter -->
         <record id="mt_job_failed" model="mail.message.subtype">
             <field name="name">Job failed</field>
--- a/queue_job/data/queue_data.xml
+++ b/queue_job/data/queue_data.xml
@@ -11,11 +11,11 @@
             <field name="code">model.requeue_stuck_jobs()</field>
         </record>
         <record id="ir_cron_queue_job_fail_dead_jobs()" model="ir.cron">
-            <field name="name">Jobs Garbage Collector</field>
+            <field name="name">Take care of unresponsive jobs</field>
             <field name="interval_number">15</field>
             <field name="interval_type">minutes</field>
             <field name="numbercall">-1</field>
-            <field ref="model_queue_job" name="model_id" />
+            <field name="model_id" ref="model_queue_job" />
             <field name="state">code</field>
             <field name="code">model.fail_dead_jobs(240)</field>
         </record>
--- a/queue_job/data/queue_data.xml
+++ b/queue_job/data/queue_data.xml
@@ -10,7 +10,7 @@
             <field name="state">code</field>
             <field name="code">model.requeue_stuck_jobs()</field>
         </record>
-        <record id="ir_cron_queue_job_fail_dead_jobs()" model="ir.cron">
+        <record id="ir_cron_queue_job_fail_dead_jobs" model="ir.cron">
             <field name="name">Take care of unresponsive jobs</field>
             <field name="interval_number">15</field>
             <field name="interval_type">minutes</field>
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -418,7 +418,7 @@ class QueueJob(models.Model):
         ).requeue()
         return True
 
-    def fail_dead_jobs(self, started_delta):
+    def fail_dead_jobs(self, started_delta, force_low_delta=False):
         """Set as failed job started since too long ago.
 
         Workers can be dead without anyone noticing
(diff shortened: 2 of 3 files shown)


queue_job/data/queue_data.xml
```diff
@@ -10,6 +10,15 @@
             <field name="state">code</field>
             <field name="code">model.requeue_stuck_jobs()</field>
         </record>
+        <record id="ir_cron_queue_job_fail_dead_jobs()" model="ir.cron">
+            <field name="name">Jobs Garbage Collector</field>
```
amh-mw: This has the same name as the job above.

queue_job/models/queue_job.py
```diff
@@ -418,6 +418,48 @@ def requeue_stuck_jobs(self, enqueued_delta=5, started_delta=0):
...
+
+        This function, mark jobs started longtime ago
+        as failed.
+
+        Cause of death can be CPU Time limit reached
+        a SIGTERM, a power shortage, we can't know, etc.
+        """
+        now = fields.datetime.now()
+        started_dl = now - timedelta(minutes=started_delta)
+        if started_delta <= 10:
```
amh-mw: Instead of hardcoded 10 minutes here, perhaps use `config['limit_time_real']` seconds?

amh-mw (changes requested):

queue_job/models/queue_job.py
hparfr: You may have a limit_time_real short for the regular http workers and a long for queue workers on a separate instance.

The idea is to use this cron as a last line of defence against poorly configured instance.

amh-mw: I still think this would be better served with something like
```python
if started_delta <= int(self.env['ir.config_parameter'].sudo().get_param('queue_job.limit_time_dead', 10)):
```

hparfr: I added an argument to bypass the 10 min limit without having  to change the code

queue_job/data/queue_data.xml
```diff
@@ -10,6 +10,15 @@
             <field name="state">code</field>
             <field name="code">model.requeue_stuck_jobs()</field>
         </record>
+        <record id="ir_cron_queue_job_fail_dead_jobs()" model="ir.cron">
```
amh-mw: Just noticed the `()` in the id here. Not sure if this could be problematic?

queue_job/models/queue_job.py
```diff
@@ -418,6 +418,61 @@ def requeue_stuck_jobs(self, enqueued_delta=5, started_delta=0):
         ).requeue()
         return True
 
+    def fail_dead_jobs(self, started_delta, force_low_delta=False):
```
amh-mw, line 421: Not a required change, but I think it would be reasonable to `force_low_delta=True` here to turn your safety check on by default. I'm going to override
```xml
            <field name="code">model.fail_dead_jobs(240)</field>
```
anyway to significantly decrease the started_delta for my own purposes, so it will be no inconvenience to me to also disable that check.

amh-mw (approved):

queue_job/models/queue_job.py
hparfr: It's the opposite, you will have to change the cron to something like : 
```py
model.fail_dead_jobs(5, force_low_delta=True)
```

BTW, why are you planning to put a low value here ? What's your use case ?

amh-mw: > BTW, why are you planning to put a low value here ? What's your use case ?

I'm impatient. 😉 I don't use a dedicated queue_job server, so all my jobs run within the default 60/120 cpu/real time limits. Ten minutes is an eternity. 😉 I also plan to decrease the Dead Job cron interval.

queue_job/data/queue_data.xml
```diff
@@ -10,6 +10,15 @@
...
             <field name="code">model.requeue_stuck_jobs()</field>
         </record>
+        <record id="ir_cron_queue_job_fail_dead_jobs" model="ir.cron">
+            <field name="name">Take care of unresponsive jobs</field>
+            <field name="interval_number">15</field>
+            <field name="interval_type">minutes</field>
+            <field name="numbercall">-1</field>
+            <field name="model_id" ref="model_queue_job" />
+            <field name="state">code</field>
+            <field name="code">model.fail_dead_jobs(240)</field>
```
florian-dacosta, line 20: I think it would be nice to add an advice on how to choose the value somewhere if someone wants to adapt it to its config?
In the Readme or in the methode description ? The ideal value would be the cpu_time limit from the queue job server config right ?

florian-dacosta (approved):
LGTM
I believe in a new version (18?) it would be nice to merge both cron.

The cron would then : requeue the enqueued jobs + set to fail the jobs started for too long (+ put this behavior by default).
But it is better not to do it in an already released version.

queue_job/models/queue_job.py
```diff
@@ -418,6 +418,61 @@ def requeue_stuck_jobs(self, enqueued_delta=5, started_delta=0):
...
+
+        This function, mark jobs started longtime ago
+        as failed.
+
+        Cause of death can be CPU Time limit reached
+        a SIGTERM, a power shortage, we can't know, etc.
+
+        This mechanism should be very exceptionnal.
+        It may help, for instance, if someone forget to configure
+        properly his system.
```
simahawk, line 436:
```suggestion
        This function marks jobs started longtime ago
        as failed.

        Cause of death can be CPU Time limit reached
        a SIGTERM, a power shortage, etc. We can't know.

        This mechanism should be very exceptional.
        It may help, for instance, if someone forgot to configure
        properly his system.
```

queue_job/models/queue_job.py
```diff
@@ -418,6 +418,61 @@ def requeue_stuck_jobs(self, enqueued_delta=5, started_delta=0):
...
+
+        :param started_delta: lookup time in minutes for jobs
+                                that are in started state,
+
+        :param force_low_delta: force a started_delta in less 10min
+                                = you know what you do
+        """
+        now = fields.datetime.now()
+        started_dl = now - timedelta(minutes=started_delta)
```
simahawk, line 445:
```suggestion
        started_dl = fields.Datetime.subtract(now, minutes=started_delta)
```

queue_job/models/queue_job.py
simahawk, line 421:
```suggestion
    def gc_dead_jobs(self, started_delta, force_low_delta=False):
```

queue_job/data/queue_data.xml
simahawk: I agree. I think a comment here in the code would be nice too :wink:

queue_job/data/queue_data.xml
```diff
@@ -10,6 +10,15 @@
             <field name="state">code</field>
             <field name="code">model.requeue_stuck_jobs()</field>
         </record>
+        <record id="ir_cron_queue_job_fail_dead_jobs" model="ir.cron">
+            <field name="name">Take care of unresponsive jobs</field>
```
simahawk, line 14: While thinking of a better name for this cron I wonder.... why don't we add another method and use only one cron?

Eg:

```
<field name="code">model.gc_stuck_jobs()</field>
[...]

def gc_stuck_jobs(self, started_delta=None, force_low_delta=False):
    self.requeue_stuck_jobs()
    if started_delta:
        self.gc_dead_jobs(started_delta, force_low_delta=force_low_delta)
```

WDYT?

simahawk (changes requested):

PR #658, opened 2024-06-04
[IMP] queue_job: Add split method

paradoxxxzero:
This is a draft PR trying an alternative approach for https://github.com/OCA/queue/pull/566

```python
      # ...

      def copy_all_partners(self):
          # Duplicate all partners in batches of 30:
          self.with_delay(split=30).copy()

    # or

     def copy_all_partners(self):
          # Duplicate all partners in batches of 30:
          self.delayable().set(description='Copy Partners').copy().split(30).delay()

  # ...
  self.env['res.partner'].search([], limit=1000).copy_all_partners()
```

The `split` function on `Delayable` returns a `DelayableGroup`.

This is an attempt and I am open to suggestions

### DIFF
--- a/queue_job/delay.py
+++ b/queue_job/delay.py
@@ -526,6 +526,52 @@ class Delayable:
         """Delay the whole graph"""
         self._graph.delay()
 
+    def split(self, size, chain=False):
+        """Split the Delayables.
+
+        Use `DelayableGroup` or `DelayableChain`
+        if `chain` is True containing batches of size `size`
+        """
+        if not self._job_method:
+            raise ValueError("No method set on the Delayable")
+
+        total_records = len(self.recordset)
+
+        delayables = []
+        for index in range(0, total_records, size):
+            recordset = self.recordset[index : index + size]
+            delayable = Delayable(
+                recordset,
+                priority=self.priority,
+                eta=self.eta,
+                max_retries=self.max_retries,
+                description=self.description,
+                channel=self.channel,
+                identity_key=self.identity_key,
+            )
+            # Update the __self__
+            delayable._job_method = getattr(recordset, self._job_method.__name__)
+            delayable._job_args = self._job_args
+            delayable._job_kwargs = self._job_kwargs
+
+            delayables.append(delayable)
+
+        description = self.description or (
+            self._job_method.__doc__.splitlines()[0].strip()
+            if self._job_method.__doc__
+            else "{}.{}".format(self.recordset._name, self._job_method.__name__)
+        )
+        for index, delayable in enumerate(delayables):
+            delayable.set(
+                description="%s (split %s/%s)"
+                % (description, index + 1, len(delayables))
+            )
+
+        # Prevent warning on deletion
+        self._generated_job = True
+
+        return (DelayableChain if chain else DelayableGroup)(*delayables)
+
     def _build_job(self):
         if self._generated_job:
             return self._generated_job
--- a/queue_job/tests/test_delayable.py
+++ b/queue_job/tests/test_delayable.py
@@ -1,15 +1,15 @@
 # copyright 2019 Camptocamp
 # license agpl-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 
-import unittest
-
 import mock
+
+from odoo.tests import common
 
 # pylint: disable=odoo-addons-relative-import
 from odoo.addons.queue_job.delay import Delayable, DelayableGraph
 
 
-class TestDelayable(unittest.TestCase):
+class TestDelayable(common.BaseCase):
     def setUp(self):
         super().setUp()
         self.recordset = mock.MagicMock(name="recordset")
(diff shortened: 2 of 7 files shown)


queue_job/delay.py
```diff
@@ -534,6 +534,45 @@ def delay(self):
         """Delay the whole graph"""
         self._graph.delay()
 
+    def split(self, size=None):
+        """Split the Delayable into a DelayableGroup containing batches
+        of size `size`
+        """
+
+        total_records = len(self.recordset)
+        count = 1 + total_records // size
```
amh-mw: For a division without remainder, won't this return one too many?

queue_job/delay.py
```diff
@@ -534,6 +534,45 @@ def delay(self):
...
+
+        delayables = []
+        for batch in range(count):
```
amh-mw: Perhaps simpler to let `range` do more of the work?

```python
for start in range(0, total_records, size):
```

queue_job/delay.py
```diff
@@ -607,14 +647,21 @@ def __init__(
...
     def __getattr__(self, name):
         def _delay_delayable(*args, **kwargs):
-            getattr(self.delayable, name)(*args, **kwargs).delay()
+            delayable = getattr(self.delayable, name)(*args, **kwargs)
+            if self.split:
+                group = delayable.split(self.split)
+                group.delay()
+                return [d._generated_job for d in group._delayables]
+
+            delayable.delay()
```
amh-mw: Is this necessary? Couldn't a developer just call `delayable.split()` themselves?

amh-mw (changes requested):

queue_job/delay.py
paradoxxxzero: Removed

queue_job/delay.py
paradoxxxzero: Good idea

queue_job/delay.py
paradoxxxzero: Yes, indeed. I have rewritten the loop and added some tests.

amh-mw (approved):
LGTM

paradoxxxzero:
The coverage does not seem to include the unittest :/

amh-mw:
> The coverage does not seem to include the unittest :/

Import the new test in `tests/__init__.py`

paradoxxxzero:
They weren't run as they were `unitest.TestCase`, I changed them to `TransactionCase` for odoo to run them.

paradoxxxzero:
I added a chain parameter to split into a chain instead of a group. 

Is there anything lacking for this to be merged?

amh-mw:
> Is there anything lacking for this to be merged?

Just the attention of @OCA/connector-maintainers

hparfr (approved):

queue_job/tests/test_delayable.py
```diff
@@ -1,15 +1,15 @@
...
 import mock
 
+from odoo.tests import common
+
 # pylint: disable=odoo-addons-relative-import
 from odoo.addons.queue_job.delay import Delayable, DelayableGraph
 
 
-class TestDelayable(unittest.TestCase):
+class TestDelayable(common.TransactionCase):
```
simahawk: There's no transaction required here. Use [TestCase](https://github.com/odoo/odoo/blob/16.0/odoo/tests/case.py#L95)  or BaseCase

queue_job/tests/test_delayable_split.py
```diff
@@ -0,0 +1,94 @@
...
+# @author Florian Mounier <florian.mounier@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import common
+
+# pylint: disable=odoo-addons-relative-import
+from odoo.addons.queue_job.delay import Delayable
+
+
+class TestDelayableSplit(common.TransactionCase):
```
simahawk: same regarding the base class

queue_job/delay.py
```diff
@@ -534,6 +534,50 @@ def delay(self):
         """Delay the whole graph"""
         self._graph.delay()
 
+    def split(self, size, chain=False):
+        """Split the Delayable into a DelayableGroup or DelayableChain
+        if `chain` is True containing batches of size `size`
+        """
```
simahawk:
```suggestion
        """Split the Delayables.
        
        Use `DelayableGroup` or `DelayableChain`
        if `chain` is True containing batches of size `size`
        """
```

simahawk (commented):
Nice, thanks! Just some minor remarks :)

@guewen ping :)

simahawk:
As this is not touching any existing feature and is not adding a lot of code IMO we can move on w/ the merge.
Yet, I think we miss a bit of documentation. Can you add something to the README pls?

paradoxxxzero:
Done

simahawk (approved):

PR #669, opened 2024-07-05
[16.0][IMP] queue_job: Requeue stuck jobs channel-based config

hildickethan:
- Add the option to configure the `enqueued_delta` and `started_delta` parameters on a channel by channel basis. If not set it will use the global value set in the cron
- The use case we have for this is that we use the cron to catch any jobs that get stuck in `started` when the Odoo gets automatically deployed, in order for it to be mostly invisible to the end user and not have a blocked queue for long, the value should be low (5 or so). But we have some other processes that due to the length get sent to background tasks as jobs and might exceed this 5 minute mark, causing them to requeue infinitely. The ideal situation is optimizing the job or refactoring the whole thing to have <5min jobs but being able to allow some jobs to be longer is much more convenient and often will be enough of a solution to the issue.

Some extra notes
- The domain it generates is much bigger than the previous one depending on the number of channels, but the cron execution was still instantaneous during my tests on a database with 26 channels and 3.5m job records
- The `[("channel", "not in", searched_channels)]` part of the domain is for jobs that are created in a specific channel but without creating the corresponding `queue.job.channel` record, I would deem it technically a poor usage by the job creator (I have done it before) but it would be a regression for these cases if not taken into account

### DIFF
--- a/queue_job/README.rst
+++ b/queue_job/README.rst
@@ -144,6 +144,8 @@ Configuration
   * ``started_delta``: Spent time in minutes after which a started job is considered stuck.
     This parameter should not be less than ``--limit-time-real // 60`` parameter in your configuration.
     Set it to 0 to disable this check. Set it to -1 to automate it, based in the server's ``--limit-time-real`` config parameter.
+  * These parameters are also configurable on a channel by channel basis, if set to 0 it will use the global value.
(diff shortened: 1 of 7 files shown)


sbidoul:
Work on a more robust solution is in progress in #716

PR #672, opened 2024-07-12
[16.0][FIX] queue_job: fix deadlock

BT-sschmid:
Fix a deadlock condition when there are dependent stored fields on `queue.job:state`.

Example:
 - Sale-Order has a stored computed field which depends on job-state.
 - A job updates a sale-order record.
 - A exception is raised and in the final error-handling with a child cursor in https://github.com/OCA/queue/blob/c93a25366d2c51339ae4530a42eb7393bdc98b12/queue_job/controllers/main.py#L144 the job's state is updated, which also triggers an update in the sale order. Result is a deadlock.

This commit adds a rollback in case of error during job execution, so that any locks are released.

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -32,11 +32,15 @@ class RunJobController(http.Controller):
         job.store()
         env.cr.commit()
         _logger.debug("%s started", job)
+        try:
+            job.perform()
+            # Triggers any stored computed fields before calling 'set_done'
+            # so that will be part of the 'exec_time'
+            env.flush_all()
+        except Exception:
+            env.cr.rollback()
+            raise
 
-        job.perform()
-        # Triggers any stored computed fields before calling 'set_done'
-        # so that will be part of the 'exec_time'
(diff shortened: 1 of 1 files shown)


amh-mw (commented):
Though I generally agree with the intent of this change, there is already some rollback code in the enclosing `runjob` call, specifically a rollback in the event of a `RetryableJobError`. It makes me question why the `except (FailedJobError, Exception)` stanza doesn't rollback as well, choosing instead to grab a different cursor to commit the job state.

BT-sschmid:
> Though I generally agree with the intent of this change, there is already some rollback code in the enclosing `runjob` call, specifically a rollback in the event of a `RetryableJobError`. It makes me question why the `except (FailedJobError, Exception)` stanza doesn't rollback as well, choosing instead to grab a different cursor to commit the job state.

now I got your comment ^^ 
Indeed I also don't see the need to use a new cursor if the tx is rolled back. If you agree I change it to just rollback.

Issue #618, opened 2024-01-19
Run job in a chain when previous job/jobs failed?


oerp-odoo:
When you need to handle failed jobs, there is no clear way to do it. For example, I have group of jobs that are part of chain. And that chain ends with postprocess job. If all group jobs were successful, then postprocess job is run and we can handle success path (e.g. sending message to users informing about it etc). But it can't handle failure path.

If any jobs fail, the last job that handles group jobs, is not run (which is OK in some cases). But then jobs are just stuck. I know there are notifications about failed jobs, but these kind of notifications are mainly for technical people and in my case, simple users don't use queue jobs directly. They don't even have access to queue jobs application. These jobs are run without them even knowing it.

So it would be great if there would be a way to trigger job on failure, so it could postprocess all failed jobs in a graph. And my case is handling group of jobs, not single job as those group jobs do same thing, just process separate batches of data (but such feature most likely would handle any number of jobs, its just an example).

**Describe the solution you'd like**

**GIVEN**
Jobs graph, we can create job in a chain, that is run on failure (as far as I know, jobs currently proceed only on done, but not failed state). 
**WHEN**
Any job fails that "fail processing job" depends on
**THEN**
It will be run only if any previous job it depends on, failed (but it should wait till all dependent jobs are run regardless if there are some failed jobs already).

**GIVEN**
Jobs graph, we can create job in a chain, that is run on failure (as far as I know, jobs currently proceed only on done, but not failed state). 
**WHEN**
All dependent jobs are done
**THEN**
 failure handling job is skipped (maybe would need new state skipped?)

**Describe alternatives you've considered**

I've considered hooking on failure message posting that already exists with queue jobs, which is related to my current problem (to send aggregated message about all failed jobs, not for specific job), but job that handles failed jobs would be more universal as there could be various cases when failure needs to be handled.

Also, maybe simpler solution would be to allow postprocessing job to just run on both done and failed states (with some extra parameter passed). So this way it would still be possible to handle failed job, but then developer would be responsible for recognizing what to do inside that job and detect whether related jobs failed or all passed.

amh-mw:
See https://github.com/OCA/queue/pull/674?

PR #674, opened 2024-07-26
[16.0][IMP] queue_job: run specific hook method after max_retries

QuocDuong1306:
When a job has been tried for max_retries but still fails, final `FailedJobError` is raised and job is set to `Failed`. 

This PR enables developers to run a specific hook method when this happens.

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -527,6 +527,21 @@ class Job(object):
             elif not self.max_retries:  # infinite retries
                 raise
             elif self.retry >= self.max_retries:
+                hook = f"{self.method_name}_on_max_retries_reached"
+                if hasattr(self.recordset, hook):
+                    recordset = self.recordset.with_context(
+                        job_uuid=self.uuid, exc_info=self.exc_info
+                    )
+                    try:
+                        getattr(recordset, hook)()
+                    except Exception as ex:
+                        _logger.debug(
+                            "Exception on %s:%s() for Job UUID: %s",
+                            self.recordset,
+                            hook,
+                            self.uuid,
+                            ex,
+                        )
                 type_, value, traceback = sys.exc_info()
                 # change the exception type but keep the original
                 # traceback and message:
--- a/queue_job/README.rst
+++ b/queue_job/README.rst
@@ -397,6 +397,33 @@ Based on this configuration, we can tell that:
 * retries 10 to 15 postponed 30 seconds later
 * all subsequent retries postponed 5 minutes later
 
+**Job function: reach max retryable times**
+
+When a job has reached the maximum number of retries and still fails, 
+the job's status will be set to ``Failed``. 
+You can define a specific method to handle this event. 
+The method should be named ``{method_name}_on_max_retries_reached``.
+
+Here's an example:
+
+.. code-block:: python
+
+    from odoo import models, fields, api
+
+    class MyModel(models.Model):
+        _name = 'my.model'
+
+        def button_done(self):
+            self.env['my.model'].with_delay().my_method('a', k=2)
+
+        def my_method_on_max_retries_reached(self):
+            # This method is called when the job reaches the maximum retries and fails
+            # Add your custom logic here
+
+In this example, ``my_method_on_max_retries_reached`` is the method 
+that will be called when the job my_method fails after reaching the maximum retries. 
+You can add your custom logic inside this method to handle the event.
+
 **Job Context**
 
 The context of the recordset of the job, or any recordset passed in arguments of
--- a/queue_job/readme/USAGE.rst
+++ b/queue_job/readme/USAGE.rst
@@ -242,6 +242,33 @@ Based on this configuration, we can tell that:
 * retries 10 to 15 postponed 30 seconds later
 * all subsequent retries postponed 5 minutes later
 
+**Job function: reach max retryable times**
+
+When a job has reached the maximum number of retries and still fails, 
+the job's status will be set to ``Failed``. 
+You can define a specific method to handle this event. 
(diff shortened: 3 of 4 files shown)


queue_job/job.py
```diff
@@ -527,6 +527,21 @@ def perform(self):
             elif not self.max_retries:  # infinite retries
                 raise
             elif self.retry >= self.max_retries:
+                hook = f"{self.method_name}_on_max_retries_reached"
```
amh-mw, line 530: I am generally not a fan of interpolating method names. Pass `on_exception` as an additional argument to `delayable`/`with_delay` instead?

Perhaps the scope could be slightly broader as well? Give the developer a chance to handle all types of exception, not just `FailedJobError`?

amh-mw (changes requested):

queue_job/job.py
nilshamerlinck: - interpolating method names is quite a common pattern in odoo code: see lots of `getattr` in the codebase :)
- quite elegant imho to be able to define `method_name` and `method_name_on_max_retries_reached` nearby, but of course it's a bit subjective
- regarding your last point, that's an interesting idea but it feels quite natural to handle exceptions in the job code itself, e.g. in the EDI framework [here](https://github.com/OCA/edi-framework/blob/b6a69b55593cf0b2adbf8d1c24a23a73fb6ef323/edi_oca/models/edi_backend.py#L303)

simahawk: A more declarative approach could be to use a decorator but it will likely add complexity.
@QuocDuong1306 could you please update the docs?

QuocDuong1306: Hi @simahawk , I updated the docs

versadabot: I would say whenever job reaches failed state, it would be useful to have a hook, to do something, not when it just failed after max retries, but failed for any reason?

For example, issue described here: https://github.com/OCA/queue/issues/618

oerp-odoo: I would say whenever job reaches failed state, it would be useful to have a hook, to do something, not when it just failed after max retries, but failed for any reason?

For example, issue described here: https://github.com/OCA/queue/issues/618

simahawk: That's a good point. Yet, I think you can subscribe to that particular event easily (job switching to failed).
In fact we could subscribe even in this case and check the max retry counter.
@guewen did you have something in mind regarding handling failures?

guewen: Previously this is the kind of thing we would add to the `@job` decorator, things that were configured on this decorator are now on `queue.job.function`. This is akin to the "related actions" where we store the method to execute there. Different jobs can be pointed to the same error handler, and we would be able to use an handler on "no-code jobs" easily (e.g. I call an existing method with `with_delay` in a script, and I want to notify slack when the max failure is reached using a handler that already exists in the code, I can create a queue job function and set this handler from the UI).

I agree with your points on triggering when switching to failed, not considering retries, then it would be worth to provide the max retry and current retry count to the handler as well.

Something to pay really attention to in the implementation is the transaction handling: I think in the current form, if the job failed with any error that causes a rollback (such as a serialization error for example), the transaction is unusable and the handler will probably fail as well! We should probably execute it in a new transaction, but then be aware that it will not be up-to-date with whatever happened in the current transaction, and could be subject to deadlocks depending of what the failed job did and the failure handler does...

Considering that, I'd also be more confortable if the handling happens somewhere in 

```
    def _try_perform_job(self, env, job):
        """Try to perform the job."""
        job.set_started()
        job.store()
        env.cr.commit()
        _logger.debug("%s started", job)

        job.perform()
        # Triggers any stored computed fields before calling 'set_done'
        # so that will be part of the 'exec_time'
        env.flush_all()
        job.set_done()
        job.store()
        env.flush_all()
        env.cr.commit()
        _logger.debug("%s done", job)
```

So the transactional flow is more straightforward

trisdoan: > Previously this is the kind of thing we would add to the `@job` decorator, things that were configured on this decorator are now on `queue.job.function`. This is akin to the "related actions" where we store the method to execute there. Different jobs can be pointed to the same error handler, and we would be able to use an handler on "no-code jobs" easily (e.g. I call an existing method with `with_delay` in a script, and I want to notify slack when the max failure is reached using a handler that already exists in the code, I can create a queue job function and set this handler from the UI).
> 
> I agree with your points on triggering when switching to failed, not considering retries, then it would be worth to provide the max retry and current retry count to the handler as well.
> 
> Something to pay really attention to in the implementation is the transaction handling: I think in the current form, if the job failed with any error that causes a rollback (such as a serialization error for example), the transaction is unusable and the handler will probably fail as well! We should probably execute it in a new transaction, but then be aware that it will not be up-to-date with whatever happened in the current transaction, and could be subject to deadlocks depending of what the failed job did and the failure handler does...
> 
> Considering that, I'd also be more confortable if the handling happens somewhere in
> 
> ```
>     def _try_perform_job(self, env, job):
>         """Try to perform the job."""
>         job.set_started()
>         job.store()
>         env.cr.commit()
>         _logger.debug("%s started", job)
> 
>         job.perform()
>         # Triggers any stored computed fields before calling 'set_done'
>         # so that will be part of the 'exec_time'
>         env.flush_all()
>         job.set_done()
>         job.store()
>         env.flush_all()
>         env.cr.commit()
>         _logger.debug("%s done", job)
> ```
> 
> So the transactional flow is more straightforward

Hello @guewen, I started in https://github.com/OCA/queue/pull/734, would you mind taking a look please?

simahawk:
@QuocDuong1306 @trisdoan  shall we close this one in favor of #734 ?

PR #676, opened 2024-08-01
[14.0][IMP] queue_job: automatically update dependent jobs when state changes manually

QuocDuong1306:
Context:

- If a user explicitly cancels a job, he expects its child jobs to be canceled as well;

- However if a job fails, child jobs will stay in WAIT_DEPENDENCIES state; an action is expected from user to fix it and job might succeed after being re-queued

Note

- No need to  forward port for the `FIX` commit, it was fixed already as part of migration to 16.0+ ([here](https://github.com/OCA/queue/blob/7e5efe7263ba4f79f704b9abc7d565de8285511f/queue_job/models/queue_job.py#L323))

### DIFF
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -308,6 +308,7 @@ class QueueJob(models.Model):
             if state == DONE:
                 job_.set_done(result=result)
                 job_.store()
+                record.env["queue.job"].flush()
                 job_.enqueue_waiting()
             elif state == PENDING:
                 job_.set_pending(result=result)
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -316,6 +316,8 @@ class QueueJob(models.Model):
             elif state == CANCELLED:
                 job_.set_cancelled(result=result)
                 job_.store()
+                record.env["queue.job"].flush()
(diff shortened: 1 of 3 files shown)


queue_job/models/queue_job.py
```diff
@@ -308,13 +308,16 @@ def _change_job_state(self, state, result=None):
             if state == DONE:
                 job_.set_done(result=result)
                 job_.store()
+                record.env["queue.job"].flush()
```
amh-mw, line 311: Why is this flush necessary? This isn't the code path that was changed.

QuocDuong1306: Hello @amh-mw , It demonstrates the need for the flush: if you remove it, newly added tests won't pass:
- https://github.com/OCA/queue/pull/677

The parent jobs need to be flushed before updating the child. Like v16 as is ([here](https://github.com/OCA/queue/blob/7e5efe7263ba4f79f704b9abc7d565de8285511f/queue_job/models/queue_job.py#L323))

guewen (approved):
Thanks!

amh-mw (approved):

simahawk (approved):

simahawk:
/ocabot merge minor
Merci! :)

PR #679, opened 2024-08-23
[IMP] queue_job: Don't raise a warning for valid context

rousseldenis:
As the 'queue_job__no_delay' context key is the valid one to use, don't raise warnings during tests, but log it as information level.

@simahawk Warning level should be only used to show something that could goes wrong (e.g.: deprecated methods, wrong definitions, ...). In this case, the use of the correct key context should not raise a warning and log as information only.

As I try to keep an eye on real warnings, they are lost in the crowd like:

https://github.com/OCA/sale-workflow/actions/runs/10522761814/job/29156065217#step:8:3025

### DIFF
--- a/test_queue_job/tests/test_job_auto_delay.py
+++ b/test_queue_job/tests/test_job_auto_delay.py
@@ -32,11 +32,16 @@ class TestJobAutoDelay(JobCommonCase):
 
     def test_auto_delay_force_sync(self):
         """method forced to run synchronously"""
-        result = (
-            self.env["test.queue.job"]
-            .with_context(_job_force_sync=True)
-            .delay_me(1, kwarg=2)
+        with self.assertLogs(level="WARNING") as log_catcher:
+            result = (
+                self.env["test.queue.job"]
+                .with_context(_job_force_sync=True)
+                .delay_me(1, kwarg=2)
+            )
+        self.assertEqual(
+            len(log_catcher.output), 1, "Exactly one warning should be logged"
         )
+        self.assertIn(" ctx key found. NO JOB scheduled. ", log_catcher.output[0])
         self.assertTrue(result, (1, 2))
 
     def test_auto_delay_context_key_set(self):
--- a/queue_job/utils.py
+++ b/queue_job/utils.py
@@ -36,5 +36,5 @@ def must_run_without_delay(env):
             return True
 
     if env.context.get("queue_job__no_delay"):
-        _logger.warning("`queue_job__no_delay` ctx key found. NO JOB scheduled.")
+        _logger.info("`queue_job__no_delay` ctx key found. NO JOB scheduled.")
         return True


guewen (approved):
Maybe it should be logged as warning if test_enable is not set? (e.g. run on production). That was maybe the original intent?

rousseldenis:
> Maybe it should be logged as warning if test_enable is not set? (e.g. run on production). That was maybe the original intent?

And if we really want (I don't do it) to run without delay ? I think this could be done without warnings.

amh-mw (approved):
LGTM, but adding the `test_enable` check would be superior.

rousseldenis:
@simahawk Your review on this would be welcome. Thanks

rousseldenis:
@guewen @amh-mw Adapted to warn if not testing.

simahawk:
> Maybe it should be logged as warning if test_enable is not set? (e.g. run on production). That was maybe the original intent?

AFAIR yes :)

> And if we really want (I don't do it) to run without delay ? I think this could be done without warnings.

AFAIR I was thinking of this as well. 

Thinking again about this right now: in fact, if you do it explicitly there's no reason to log a warning no matter if in tests or in prod mode. IMO we can skip the distinction and always log the same info message.

rousseldenis:
> Thinking again about this right now: in fact, if you do it explicitly there's no reason to log a warning no matter if in tests or in prod mode. IMO we can skip the distinction and always log the same info message.

Indeed, putting it in the context is sufficient

amh-mw:
> Thinking again about this right now: in fact, if you do it explicitly there's no reason to log a warning no matter if in tests or in prod mode. IMO we can skip the distinction and always log the same info message.

I'm good with never warning.

gurneyalex (approved):

rousseldenis:
So, just one info message if context used, correct (I need to change this PR) ?

FrancoMaxime (approved):

test_queue_job/tests/test_job_auto_delay.py
```diff
@@ -32,11 +32,16 @@ def test_auto_delay_inside_job(self):
 
     def test_auto_delay_force_sync(self):
         """method forced to run synchronously"""
-        result = (
-            self.env["test.queue.job"]
-            .with_context(_job_force_sync=True)
-            .delay_me(1, kwarg=2)
+        with self.assertLogs(level="WARNING") as log_catcher:
```
simahawk, line 35: this test needs an update no?

rousseldenis: Nope, corresponding to true warnings: https://github.com/OCA/queue/pull/679/files#diff-6e33951ef77e0428f3edcd938952cf20d30ca84e703406c08685f3d70d68da86R28

PR #680, opened 2024-09-03
[14.0][BKP] queue_job: unify no delay option

simahawk:
Current situation:

* multiple keys for no good reason
* half baked: not all of them used everywhere
* no centralization
* poor naming

With this change we'll have:

* 1 and only one key to disable via ctx: ``queue_job__no_delay``
* 1 and only one key to disable via os env: ``QUEUE_JOB__NO_DELAY``
* backward compatibility with deprecation for old keys

Backport of #521

### DIFF
--- a/queue_job/utils.py
+++ b/queue_job/utils.py
@@ -1,0 +1,40 @@
+# Copyright 2023 Camptocamp
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
+
+import logging
+import os
+
+_logger = logging.getLogger(__name__)
+
+
+def must_run_without_delay(env):
+    """Retrun true if jobs have to run immediately.
+
+    :param env: `odoo.api.Environment` instance
+    """
+    # TODO: drop in v17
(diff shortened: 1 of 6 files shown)


queue_job/utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+    if os.getenv("QUEUE_JOB__NO_DELAY"):
+        _logger.warning("`QUEUE_JOB__NO_DELAY` env var found. NO JOB scheduled.")
+        return True
+
+    # TODO: drop in v17
+    deprecated_keys = ("_job_force_sync", "test_queue_job_no_delay")
+    for key in deprecated_keys:
+        if env.context.get(key):
+            _logger.warning(
```
simahawk, line 31: Probably is better to silent this log and use `debug`.

simahawk, line 31: NOTE: the warning here is the same we had already + the deprecation message.

amh-mw (approved):
LGTM

PR #689, opened 2024-10-02
[fix][17.0] To show dependency graph

mostafabarmshory:
The dependency graph widget is faced with error due to new updates of odoo 17.0. I fix and check it with new concepts of odoo web.


mostafabarmshory:
See issue #690

guewen (approved):
Thanks!

PR #683, opened 2024-09-16
[14.0][IMP] queue_job: add filter on Date Created

florentx:
Search filter for "Date created" and related "Group by"

PR #657, opened 2024-06-04
[IMP] queue_job: Display warning before displaying big dependency graphs

paradoxxxzero:
Big dependency graphs currently slow down web page, this PR adds a warning in this case:

 
20240604_09h13m27s_grim

PR #629, opened 2024-03-06
[14.0] queue_job: fix partial index to add `wait_dependencies` state

sebalix:
Some details about the search performed by the method `job_record_with_same_identity_key` in file `queue_job/job.py`, the partial index wasn't used before:

### Before
```sql
 Limit  (cost=6874.52..6874.53 rows=1 width=20) (actual time=0.123..0.123 rows=0 loops=1)
   ->  Sort  (cost=6874.52..6874.53 rows=1 width=20) (actual time=0.121..0.121 rows=0 loops=1)
         Sort Key: date_created DESC, date_done DESC
         Sort Method: quicksort  Memory: 25kB
         ->  Index Scan using queue_job_state_index on queue_job  (cost=0.42..6874.51 rows=1 width=20) (actual time=0.113..0.113 rows=0 loops=1)
               Index Cond: ((state)::text = ANY ('{wait_dependencies,pending,enqueued}'::text[]))
               Filter: ((identity_key)::text = 'f75c2b628243d651bad356030a3021a95dbe7725'::text)
 Planning Time: 0.375 ms
 Execution Time: 0.161 ms
```

### After
```sql
 Limit  (cost=8.15..8.16 rows=1 width=20) (actual time=0.032..0.032 rows=0 loops=1)
   ->  Sort  (cost=8.15..8.16 rows=1 width=20) (actual time=0.030..0.030 rows=0 loops=1)
         Sort Key: date_created DESC, date_done DESC
         Sort Method: quicksort  Memory: 25kB
         ->  Index Scan using queue_job_identity_key_state_partial_index on queue_job  (cost=0.12..8.14 rows=1 width=20) (actual time=0.019..0.019 rows=0 loops=1)
               Index Cond: ((identity_key)::text = 'f75c2b628243d651bad356030a3021a95dbe7725'::text)
 Planning Time: 0.844 ms
 Execution Time: 0.094 ms
```

PR #692, opened 2024-10-07
[18.0][MIG] queue_job: Migrate to 18.0

thienvh332:
### Ported PRs

- #413
- #387
- #466
- #495
- #594
- #598
- #629
- #624
- #657
- #676
- #683

### Changes
- Cleaned TODO tags

- Refactor the JobSerialized field on top of odoo.fields.Json

    - Result
   
```
>>> self.env.cr.execute("SELECT column_name, data_type FROM information_schema.columns WHERE table_name = 'queue_job' AND column_name = 'kwargs';")
>>> self.env.cr.fetchall()
[('kwargs', 'jsonb')]

```
- Example for searching:
```
>>> self.env.cr.execute("SELECT kwargs->>'a' FROM queue_job WHERE kwargs ? 'a'")
>>> self.env.cr.dictfetchall()
[{'?column?': 'abc'}, {'?column?': 'abc'}]
```


simahawk:
@thienvh332 can you please check pre-commit?
Also, I'd say that the test module should stay together w/ the main one https://github.com/OCA/queue/pull/694

queue_job/fields.py
```diff
@@ -23,7 +24,7 @@ class JobSerialized(fields.Field):
     """
 
     type = "job_serialized"
-    column_type = ("text", "text")
+    _column_type = ("jsonb", "jsonb")
```
simahawk, line 27: to preserve existing jobs this should be migrated, no?

thienvh332: I will add it after fixing the pre-commit bug

sbidoul:
In 18, warnings in logs cause tests to fail because they often reveal real issues. Two in this case:

```
errors that caused failure (2):
2024-10-07 11:47:52,343 245 WARNING odoo odoo.addons.queue_job.delay: Delayable Delayable(test.queue.job().testing_method((1,), {})) was prepared but never delayed 
2024-10-07 11:47:56,895 245 WARNING odoo odoo.addons.queue_job.utils: `queue_job__no_delay` ctx key found. NO JOB scheduled. 
```

amh-mw:
Forward porting https://github.com/OCA/queue/pull/679 should handle `queue_job__no_delay`.

queue_job/tests/common.py
```diff
@@ -444,9 +443,10 @@ def load_tests(loader, tests, ignore):
         Also extend the DocTestCase class trivially to fit the class teardown
         that Odoo backported for its own test classes from Python 3.8.
         """
-        if sys.version_info < (3, 8):
-            doctest.DocTestCase.doClassCleanups = lambda: None
-            doctest.DocTestCase.tearDown_exceptions = []
+        # UP036 Version block is outdated for minimum Python version
```
simahawk: do we need this?

simahawk (approved):
LGTM overall

simahawk:
@thienvh332 reminder: if you don't quote w/ back-ticks the PR references, we get links automatically :wink:

sebalix (commented):
Not sure we should port https://github.com/OCA/queue/pull/387 to revert it afterwards by porting https://github.com/OCA/queue/pull/466

sebalix:
Should we also port these PRs?
- https://github.com/OCA/queue/pull/676
- https://github.com/OCA/queue/pull/683

They are missing when I check ports from 14.0 to 17.0

EDIT: fixed the 2nd PR ref

simahawk:
> Not sure we should port #387 to revert it afterwards by porting #466

Good point. We should blacklist those PRs.

> Should we also port these PRs?

I'd say yes.

@thienvh332 ping

thienvh332:
> > Not sure we should port #387 to revert it afterwards by porting #466

> Good point. We should blacklist those PRs.

In my opinion, these 2 commits have 1 more test case so we should keep it. I think the more test cases the better. Of course, let me know if you still want to ignore it. I will drop it.
> 
> > Should we also port these PRs?
> 
> I'd say yes.
> 
I have ported both of those commits. I missed adding them to the description.

simahawk:
> In my opinion, these 2 commits have 1 more test case so we should keep it. I think the more test cases the better. Of course, let me know if you still want to ignore it. I will drop it.


ahh you are right, the PR was not fully reverted. Tests were kept. Fine for me :+1:

sebalix:
To ease review, the migration commit is 21eb44d99a5a6bfbcce8ea1d8af2c508eba3f52f

Seems pre-commit stuff have been included in it, not sure?

pedrobaeza:
@sebalix in this version, migrating from 17.0, there shouldn't be pre-commit changes, as most linters continue the same, the README files are already MD, and the pyproject.toml file remains the same.

sbidoul:
> most linters continue the same

With the exception of ruff that was upgraded, and does some more python syntax upgrades plus some minor tweaks along blank lines multiple `with` context managers, etc.

sebalix (approved):

simahawk:
@thienvh332 one last check

image

This change is included in the [migration commit](https://github.com/OCA/queue/pull/692/commits/21eb44d99a5a6bfbcce8ea1d8af2c508eba3f52f) but does not sounds like related to migration.
I guess is by error? Is it something we have to backport?

thienvh332:
> This change is included in the [migration commit](https://github.com/OCA/queue/pull/692/commits/21eb44d99a5a6bfbcce8ea1d8af2c508eba3f52f) but does not sounds like related to migration. I guess is by error? Is it something we have to backport?

- This problem happened because the new test would raise an **`ERROR`** when a **`WARNING`** occurred in the test case. So I added it. 
- I separated it into a commit so I could backport it. (https://github.com/OCA/queue/pull/692/commits/d308edc4efe97df81ce61dfa840c9f5a587f57a4)

sbidoul:
@thienvh332 please don't change the behaviour based on test_enabled. Better use `self.assertLogs` to trap the log and test that it actually logged what was expected. See for example: https://github.com/OCA/server-tools/pull/3050/commits/7d2e3d88c20175a91fcd5cf223417d9cf5da2f49

thienvh332:
> @thienvh332 please don't change the behaviour based on test_enabled. Better use `self.assertLogs` to trap the log and test that it actually logged what was expected. See for example: [OCA/server-tools@7d2e3d8](https://github.com/OCA/server-tools/commit/7d2e3d88c20175a91fcd5cf223417d9cf5da2f49)

Hi @sbidoul 
I think you are right, I tried the update but it didn't work for me. Could you help me see what I am missing?

thienvh332:
As [commented in unittest](https://github.com/OCA/queue/blob/c062b727ddfe984c68ef1bf4d16765e633da02d3/test_queue_job/tests/test_delayable.py#L282-L287), I am not able to control the printing of WARNING. So I will revert to the original handling. If you guys have any suggestions to handle it I will fix it.

sbidoul:
The problem is we don't see what you tried anymore, so it's difficult to help.

thienvh332:
Hi @sbidoul , indeed I should have pushed a fixup commit instead of force-pushing
- So here is what happened:
  - the **`.assertLogs`** was trapping the log as expected when calling **`node.__del__`** explicitely
  - but python was still garbage collecting the **`node`** after the test, so an additional warning was emitted
 
- To fix it, I now explicitly run the garbage collection with **`gc`**

Thanks for your help

sbidoul:
@thienvh332 looks good, thanks!

test_queue_job/tests/test_delayable.py
```diff
@@ -279,20 +282,23 @@ def test_delayable_group_of_chain(self):
         )
 
     def test_log_not_delayed(self):
-        logger_name = "odoo.addons.queue_job"
-        with self.assertLogs(logger_name, level="WARN") as test:
+        with self.assertLogs(level=logging.WARNING) as test:
```
mmequignon, line 285: Any WARNING log would be catched here, we should keep the logger filter.

test_queue_job/tests/test_job_auto_delay.py
```diff
@@ -32,16 +32,7 @@ def test_auto_delay_inside_job(self):
 
     def test_auto_delay_force_sync(self):
         """method forced to run synchronously"""
-        with self.assertLogs(level="WARNING") as log_catcher:
-            result = (
-                self.env["test.queue.job"]
-                .with_context(_job_force_sync=True)
-                .delay_me(1, kwarg=2)
-            )
```
mmequignon: why aren't we keeping this `_job_force_sync` context key in the `test_auto_delay_force_sync` test method ?

mmequignon (commented):
A few remarks, LG overall

test_queue_job/tests/test_job_auto_delay.py
thienvh332: Hi @mmequignon ,
- Simone introduced that standardization in 16.0 (https://github.com/OCA/queue/pull/521). It should have been dropped as part of 17.0 migration but wasn't.
- So it's dropped in 18.0

omalbastin:
Till v17 we can use the NothingtoDoJob exception to conditions that doesnot fall under a rule so that it will set the job to done. As the class NothingToDoJob has been removed from Exception in v18, how can we handle the same?

john-herholz-dt:
@omalbastin did you find a way to replace it? We are encountering the same issue right now.

@sbidoul you might also give some hints here. Can you explain the removal of NothingToDoJob and whats the best way to replace it?

It is a bit unfortunate that most connector implementations are abandoned or in private forks - the framework itself is quite cool.

sbidoul:
> Can you explain the removal of NothingToDoJob 

I'm afraid not. It seems it has been removed in this PR but I don't know why.

> and whats the best way to replace it?

The use case of that exception is unclear to me, to be honest.  If the job has nothing to do, why not simply... let it do nothing and end normally?

PR #698, opened 2024-10-14
[15.0][IMP] queue_job: Configure default subchannel capacity.

amh-mw:
This adds a new `subcapacity` option to channels that allows the
configuration of the default capacity for autocreated child channels.

For example, environment `ODOO_QUEUE_JOB_CHANNELS=root:8:subcapacity=1`
would set the capacity of an autocreated `root.sub` channel to 1.

### DIFF
--- a/queue_job/jobrunner/channels.py
+++ b/queue_job/jobrunner/channels.py
@@ -381,7 +381,9 @@ class Channel(object):
     without risking to overflow the system.
     """
 
-    def __init__(self, name, parent, capacity=None, sequential=False, throttle=0):
+    def __init__(
+        self, name, parent, capacity=None, sequential=False, throttle=0, subcapacity=0
+    ):
         self.name = name
         self.parent = parent
         if self.parent:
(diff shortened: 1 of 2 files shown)


amh-mw:
My own use case is that I run multi-company odoo with an environment like ODOO_QUEUE_JOB_CHANNELS=root:8,single:1. The single channel is for problem children that tend to generate an excess of SerializationFailure if executed in parallel. I've recently determined that those SerializationFailure are company-specific, so a channel per company would improve parallelism.

Per @sbidoul in OCA Discord
> Undeclared channels are auto created at runtime, but with an unlimited capacity.

This setting allows me to limit the capacity of auto-created channels.

sbidoul:
Thanks for addressing this @amh-mw 

I would have a preference for configuring this in the parent channel, instead of globally.
Like this is done, for instance for the `throttle` config.

amh-mw:
> I would have a preference for configuring this in the parent channel, instead of globally. Like this is done, for instance for the `throttle` config.

Understood. I will unfortunately not have time to circle back around to that request until next year. Would it be possible to mark this issue`nostale`?

sbidoul:
@amh-mw Count on me to repoen it if the stale bot closes it. ~Can you set it Draft?~ [done]

amh-mw:
@sbidoul Version two ready for review.

amh-mw:
Reopening against 18.0 https://github.com/OCA/queue/pull/767

PR #708, opened 2024-11-18
[15.0][FIX] queue_job: Post failure after write

hhgabelgaard:
To make it possible to include the exception info in the failure notification


guewen (approved):
It makes sense, thanks!

PR #713, opened 2024-12-03
[IMP] queue_job: detect jobs runned by workers that have been killed

AnizR:
## Goal
Do not keep jobs in 'started' status if their associated worker have been killed by a timeout and directly set it as 'failed'

### DIFF
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -5,6 +5,8 @@ import logging
 import random
 from datetime import datetime, timedelta
 
+import psutil
+
 from odoo import _, api, exceptions, fields, models
 from odoo.osv import expression
 from odoo.tools import config, html_escape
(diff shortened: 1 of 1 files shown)


queue_job/models/queue_job.py
```diff
@@ -417,6 +419,37 @@ def autovacuum(self):
...
+                ("state", "=", "started"),
+                ("worker_pid", "!=", False),
+            ]
+        )
+
+        failed_jobs = self.env["queue.job"].browse()
+
+        for job in jobs:
+            if not psutil.pid_exists(job.worker_pid):
+                _logger.info("a process with pid %d does not exist" % job.worker_pid)
```
amh-mw: If I saw this in the logs, I don't think I'd know what to look for. Maybe say "a queue_job process" instead?

queue_job/models/queue_job.py
```diff
@@ -417,6 +419,37 @@ def autovacuum(self):
...
+                failed_jobs += job
+        if failed_jobs:
+            for job in failed_jobs:
```
amh-mw: Why a second loop? I don't see any value in accumulating failed_jobs; just set the jobs failed as they are found?

amh-mw (approved):
This is a recurring problem for me, so I will backport to 15.0 shortly after merge.

queue_job/models/queue_job.py
AnizR: Oops, i forgot to remove it, it to ensure that it was working during my tests :smile:

queue_job/models/queue_job.py
AnizR: Indeed, not necessary, thanks!

queue_job/models/queue_job.py
amh-mw: I do think there is value in logging unexpected queue_job workers deaths with greater specificity than the generic werkzeug worker timeout message.

AnizR: I see, I updated the message to give the worker pid and the job uuid.

guewen (commented):
Hi, sadly the pid may be on another host and still executing

AnizR:
> Hi, sadly the pid may be on another host and still executing

What would you advise ?
Making the server update the job before killing the worker? (extending the function[ `process_timeout` ](https://github.com/odoo/odoo/blob/16.0/odoo/service/server.py#L881)? I have no idea if it is possible )

amh-mw:
> Hi, sadly the pid may be on another host and still executing

Record the hostname / other signature of the host and filter accordingly? Config parameter gate this feature?

sbidoul:
@AnizR I think https://github.com/OCA/queue/pull/423 was promising and is still the best approach I can think of.

AnizR:
This wasn't a good approach.
I have created another PR #716 that seems more promising.

PR #716, opened 2024-12-06
[16.0][IMP] queue_job: remove dead jobs requeuer cron and automatically requeue dead jobs

AnizR:
## Goal

Automatically re-queue jobs that have been started but whose worker have been killed.
And rely on Odoo's limit_time_cpu and limit_time_real for job execution.

## Technical explanation

Everything relies on a new table `queue_job_locks` which contains ids of jobs that have been started.
When a job is executed, its row `queue_job_locks` is locked.

If a row is in `queue_job_locks` with a `state='started'` but not locked, it is either:
 - a job killed
 - a job not yet started (happens in a very small time frame)

Using this information, we can re-queue these jobs.

## Why not lock directly in the `queue_job' table?

This was tried in #423  but it wasn't working when a job was raising an error.
It seems that the row was locked and it tried to write on the row to set as `failed` before committing.
 
## Improve current behavior

Re-queue jobs that have been killed but increment their 'retries' to avoid having a job that is always get killed in infinite re-queuing.


queue_job/readme/CONFIGURE.rst
```diff
@@ -47,15 +47,4 @@
...
-    Set it to 0 to disable this check.
-  * ``started_delta``: Spent time in minutes after which a started job is considered stuck.
-    This parameter should not be less than ``--limit-time-real // 60`` parameter in your configuration.
-    Set it to 0 to disable this check. Set it to -1 to automate it, based in the server's ``--limit-time-real`` config parameter.
-
-  .. code-block:: python
-
-    # `model` corresponds to 'queue.job' model
-    model.requeue_stuck_jobs(enqueued_delta=1, started_delta=-1)
+* Jobs in timeout will be automatically be requeued
```
sbidoul:
```suggestion
* Jobs that remain in `enqueued` or `started` state (because, for instance, their worker has been killed) will be automatically re-queued.
```

sbidoul (commented):
This looks very promising!

A few thoughts:
- There is a Caveat comment in `runner.py` that needs updating.
- This should handle the `enqueued` state too, which is the state when the runner has decided that a jobs needs to run but then `/queue_job/runjob` controller has not set it started yet. This state normally exists for a very short time, but I have seen situations where the workers are overloaded and take time to accept `/queue_job/runjob` requests then die, leaving jobs in `enqueued` state forever.
- Since there are two race conditions between `enqueued` and `started`, and between `started` and the time the job transaction actually starts with the lock, I wonder if we should not introduce a small elapsed time condition (10s?) in `reset_dead_jobs`. May be based on `date_enqueued`.

queue_job/jobrunner/runner.py
```diff
@@ -343,6 +343,60 @@ def set_job_enqueued(self, uuid):
                 (ENQUEUED, uuid),
             )
 
+    def reset_dead_jobs(self):
```
```suggestion
    def requeue_dead_jobs(self):
```

?

queue_job/jobrunner/runner.py
```diff
@@ -343,6 +343,60 @@ def set_job_enqueued(self, uuid):
...
+                            lock.id = job.id
+                    WHERE
+                        job.state  = 'started'
+                    FOR UPDATE SKIP LOCKED
+                );
+        """
+        with closing(self.conn.cursor()) as cr:
+            cr.execute(query)
+
+    def clean_job_lock_table(self):
```
sbidoul: Is this necessary? Since we have the on delete cascade on that table, it will be cleaned automatically when the jobs are deleted by the vacuum cleaner.

queue_job/migrations/16.0.2.6.9/pre-migration.py
```diff
@@ -0,0 +1,33 @@
...
+                SELECT res_id
+                FROM
+                    ir_model_data
+                WHERE
+                    module='queue_job'
+                    AND model='ir.cron'
+                    AND name='ir_cron_queue_job_garbage_collector'
+            );
+        """
+    )
```
sbidoul: This query does not seem to do what the comment above it says?

queue_job/models/queue_job.py
```diff
@@ -417,24 +417,6 @@ def autovacuum(self):
...
-                              that are in started state,
-                              0 means that it is not checked,
-                              -1 will use `--limit-time-real` config value
-        """
-        if started_delta == -1:
-            started_delta = (config["limit_time_real"] // 60) + 1
-        return self._get_stuck_jobs_to_requeue(
-            enqueued_delta=enqueued_delta, started_delta=started_delta
-        ).requeue()
-
```
sbidoul, line 439: You can remove `_get_stuck_jobs_domain` and `_get_stuck_jobs_to_requeue` too.

queue_job/jobrunner/runner.py
AnizR: seems more accurate

queue_job/jobrunner/runner.py
AnizR: Indeed, it is not necessary, I 'll remove it.

queue_job/migrations/16.0.2.6.9/pre-migration.py
AnizR: Well Spotted! It should be an update. Thanks.

AnizR:
> This looks very promising!
> 
> A few thoughts:
> 
>     * There is a Caveat comment in `runner.py` that needs updating.
> 
>     * This should handle the `enqueued` state too, which is the state when the runner has decided that a jobs needs to run but then `/queue_job/runjob` controller has not set it started yet. This state normally exists for a very short time, but I have seen situations where the workers are overloaded and take time to accept `/queue_job/runjob` requests then die, leaving jobs in `enqueued` state forever.
> 
>     * Since there are two race conditions between `enqueued` and `started`, and between `started` and the time the job transaction actually starts with the lock, I wonder if we should not introduce a small elapsed time condition (10s?) in `reset_dead_jobs`. May be based on `date_enqueued`.

Thank you for your suggestions. I have implemented the necessary corrections.
While I have not encountered a case where a job is left in the 'enqueued' state, I have made the appropriate correction to ensure the code is more robust.

queue_job/__manifest__.py
```diff
@@ -2,7 +2,7 @@
 
 {
     "name": "Job Queue",
-    "version": "16.0.2.6.8",
+    "version": "16.0.2.6.9",
```
sbidoul: It's worth a minor bump

queue_job/controllers/main.py
```diff
@@ -31,6 +31,10 @@ def _try_perform_job(self, env, job):
         job.set_started()
         job.store()
         env.cr.commit()
+        locked = job.lock()
+        if not locked:
+            # When job is considered as 'dead' before actually locking the job
+            raise OperationalError("Trying to lock job that wasn't started")
```
sbidoul: Since this is not a postgres error, maybe use another exception type?
Also wondering if it would be slightly easier to read if the raise was in the `lock()` function.

queue_job/job.py
```diff
@@ -238,6 +238,31 @@ def load_many(cls, env, job_uuids):
...
+                    FROM
+                        queue_job
+                    WHERE
+                        uuid = %s
+                        AND state='started'
+                )
+            FOR UPDATE;
+        """,
+            [self.uuid],
+        )
```
sbidoul:
```suggestion
            WHERE
                id = %s
            FOR UPDATE;
        """,
            [self.id],
        )
```

Since the job lock id is the job id?

queue_job/job.py
```diff
@@ -517,6 +542,9 @@ def perform(self):
 
         The job is executed with the user which has initiated it.
         """
+        if self.max_retries and self.retry >= self.max_retries:
+            raise FailedJobError("Max. retries (%d) reached" % (self.max_retries))
```
sbidoul: Why adding this? This sounds unrelated.

queue_job/jobrunner/runner.py
```diff
@@ -343,6 +327,48 @@ def set_job_enqueued(self, uuid):
...
+                    FROM
+                        queue_job_locks AS lock
+                        LEFT JOIN
+                            queue_job AS job
+                        ON
+                            lock.id = job.id
+                    WHERE
+                        job.state  in ('started','enqueued')
+                        AND date_enqueued <
+                        date_trunc('seconds',now() AT TIME ZONE 'utc' - INTERVAL '10 sec')
```
sbidoul: Is date_trunc necessary?

queue_job/jobrunner/runner.py
```diff
@@ -343,6 +327,48 @@ def set_job_enqueued(self, uuid):
...
+        However, when the Odoo server crashes or is otherwise force-stopped,
+        running jobs are interrupted while the runner has no chance to know
+        they have been aborted.
+        """
+        query = """
+            UPDATE
+                queue_job
+            SET
+                state='pending',
+                retry=retry+1
```
sbidoul: Strictly speaking when requeuing a job that is in `enqueued` state, it has not been tried (started) yet, so increasing the retry count in that  case is not 100% correct. We could probably ignore that case for the sake of simplicitly but it's worth a comment at least.

queue_job/jobrunner/runner.py
```diff
@@ -424,6 +450,11 @@ def initialize_databases(self):
                         self.channel_manager.notify(db_name, *job_data)
                 _logger.info("queue job runner ready for db %s", db_name)
 
+    def requeue_dead_jobs(self):
+        for db in self.db_by_name.values():
+            if db.has_queue_job:
+                db.requeue_dead_jobs()
+
```
sbidoul, line 477:
```suggestion
```

Copy paste issue?

sbidoul (commented):
A few more minor comments.

queue_job/jobrunner/runner.py
```diff
@@ -343,6 +327,48 @@ def set_job_enqueued(self, uuid):
...
+                        queue_job_locks AS lock
+                        LEFT JOIN
+                            queue_job AS job
+                        ON
+                            lock.id = job.id
+                    WHERE
+                        job.state  in ('started','enqueued')
+                        AND date_enqueued <
+                        date_trunc('seconds',now() AT TIME ZONE 'utc' - INTERVAL '10 sec')
+                    FOR UPDATE SKIP LOCKED
```
sbidoul, line 359: I have a doubt here. Since it is now a join, does the `skip locked` apply to the queue_job_locks table only, or the queue_job table, or both? If it's both, that might be problematic.

queue_job/job.py
AnizR: The job lock id is the job id.
But, I need to ensure that the job is `started`.
So, I'll keep my query as it is.

queue_job/job.py
AnizR: The check that `max_retries` is greater than `retry` was done only  when a `RetryableJobError` was raised during the execution of the job.

Since I am increasing the `retry` when re-queuing a job that was killed, I need to do the check before peforming the job, not only when raising an exception.

Moreover, I think that it doesn't make sense to perform the check only when an exception has been raised :thinking:

queue_job/jobrunner/runner.py
AnizR: `date_enqueued` is truncated to the seconds but yes, it is not necessary.
I don't see it as an issue, do you want me to remove it?

queue_job/jobrunner/runner.py
AnizR: I have modified this part, I have split it in 2 queries:
1. Set to 'pending' and increment `retry` if job was in `state` 'started'.
2. Set to 'pending' if job was in `state` 'enqueued'.

queue_job/jobrunner/runner.py
AnizR: No, it is a function on class `QueueJobRunner` that will call the function `requeue_dead_jobs` on class `Database`.
What might confuse you is that both these two classes are defined in the same class.

queue_job/jobrunner/runner.py
sbidoul: Ah, another thought here. It would be nice to log job uuids dans we requeue. With info level (or maybe even warning, since this is most likely a sign that something went wrong).

queue_job/jobrunner/runner.py
sbidoul: If it's not needed let's remove it. It will be ~more performant~ easier to read if anything.

queue_job/jobrunner/runner.py
sbidoul: Indeed!

queue_job/jobrunner/runner.py
AnizR: I have slightly modified the query (I know use a inner join) and I have checked:
When doing an inner join, line where a lock is applied on `queue_job_locks` are skipped (even if no lock is defined on `queue_job`)

queue_job/jobrunner/runner.py
AnizR: Ok, I have done:
- Query every job that are in enqueued/started but not locked and older that 10sec
- Set jobs that are started to pending and increment retry + logging
- Set jobs that are enqueued to pending + logging

queue_job/controllers/main.py
AnizR: A `RetryableJobError` seems more appropriate

queue_job/job.py
```diff
@@ -238,6 +238,34 @@ def load_many(cls, env, job_uuids):
...
+                )
+            FOR UPDATE;
+        """,
+            [self.uuid],
+        )
+
+        # 1 job should be locked
+        if not 1 == len(self.env.cr.fetchall()):
+            raise RetryableJobError(
+                "Trying to lock job that wasn't started, uuid: %s" % self.uuid
```
sbidoul:
```suggestion
                f"Trying to lock job that wasn't started, uuid: {self.uuid}"
```

queue_job/migrations/16.0.2.6.9/pre-migration.py
```diff
@@ -0,0 +1,35 @@
...
```
sbidoul: Rename the migration directory to 16.0.2.7.0

queue_job/migrations/16.0.2.6.9/pre-migration.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def migrate(cr, version):
+    # Create job lock table
+    cr.execute(
+        """
+            CREATE TABLE IF NOT EXISTS queue_job_locks (
+                id INT PRIMARY KEY,
+                CONSTRAINT
+                    foreign_key_queue_job
```
sbidoul: Maybe follow the same naming style as Odoo? `queue_job_locks_queue_job_id_fkey`

queue_job/jobrunner/runner.py
```diff
@@ -343,6 +327,62 @@ def set_job_enqueued(self, uuid):
...
+                        )
+                    FOR UPDATE SKIP LOCKED
+                )
+            RETURNING uuid
+            """
+
+            cr.execute(query)
+
+            job_uuids_to_requeue = [job_uuid[0] for job_uuid in cr.fetchall()]
+            if job_uuids_to_requeue:
```
sbidoul: nit: this if is not necessary

queue_job/jobrunner/runner.py
```diff
@@ -343,6 +327,62 @@ def set_job_enqueued(self, uuid):
...
+                for uuid in job_uuids_to_requeue:
+                    _logger.warning(
+                        "Re-queued job with uuid: %s",
+                        str(uuid),
```
```suggestion
                        uuid,
```

nit: no need for str

guewen (approved):
Thanks for tackling this old issue, I like this elegant solution. It should be quite optimized also. Congrats for this work

queue_job/data/queue_data.xml
```diff
@@ -1,17 +1,6 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
     <data noupdate="1">
-        <record id="ir_cron_queue_job_garbage_collector" model="ir.cron">
```
amh-mw, line 4: Should this become
```xml
<delete id="ir_cron_queue_job_garbage_collector" model="ir.cron"/>
```
to clean up upgrades / avoid filling cron logs with errors since requeue_stuck_jobs method is gone?

amh-mw (changes requested):

sbidoul:
@AnizR I think we can also remove the `set_job_pending` function, as the new mechanism will take care of that part.

queue_job/data/queue_data.xml
AnizR: I have archived the cron in the `pre-migration.py`. Therefore, there won't be any error.
I always prefer to archive (set active to false) rather than deleting.

AnizR:
> @AnizR I think we can also remove the `set_job_pending` function, as the new mechanism will take care of that part.

Yes, jobs that have not been started will be re-queued by my new mechanism.
Removing this part will provide a more "uniform" process of re-queuing jobs.

queue_job/jobrunner/runner.py
```diff
@@ -248,11 +210,8 @@ def urlopen():
             # for HTTP Response codes between 400 and 500 or a Server Error
             # for codes between 500 and 600
             response.raise_for_status()
-        except requests.Timeout:
-            set_job_pending()
```
sbidoul, line 252: A timeout here is normal behaviour, so we don't want to log it as an exception.

AnizR: right :+1:

sbidoul (approved):

queue_job/jobrunner/runner.py
```diff
@@ -343,6 +305,61 @@ def set_job_enqueued(self, uuid):
...
+                        )
+                    FOR UPDATE SKIP LOCKED
+                )
+            RETURNING uuid
+            """
+
+            cr.execute(query)
+
+            job_uuids_to_requeue = [job_uuid[0] for job_uuid in cr.fetchall()]
+            for uuid in job_uuids_to_requeue:
```
```suggestion
            for uuid, in cr.fetchall():
```

A little detail. Using tuple unpacking, this can be shortened.

AnizR: I'm not used to this notation but here you are :smile:

adrienpeiffer (approved):
LGTM. Thanks for that.

adrienpeiffer:
@guewen Could you merge this one ?

queue_job/migrations/16.0.2.7.0/pre-migration.py
guewen, line 1: In the manifest, the version is 2.8.0, could you update this file accordingly @AnizR?

adrienpeiffer:
@guewen  We're finally going to do a few more tests before we merge this. We'll keep you informed

florian-dacosta (approved):
Review LGTM
I also did some tests with job exceeding the time_cpu_limit and it works as expected.

queue_job/job.py
```diff
@@ -238,6 +238,34 @@ def load_many(cls, env, job_uuids):
         recordset = cls.db_records_from_uuids(env, job_uuids)
         return {cls._load_from_db_record(record) for record in recordset}
 
+    def lock(self):
+        self.env.cr.execute(
```
hparfr, line 269: please add the intent of this `def` as comment

queue_job/jobrunner/runner.py
hparfr: please add a comment on why there is `pass` here.

queue_job/migrations/16.0.2.7.0/pre-migration.py
```diff
@@ -0,0 +1,35 @@
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
+
+
+def migrate(cr, version):
+    # Create job lock table
+    cr.execute(
+        """
+            CREATE TABLE IF NOT EXISTS queue_job_locks (
+                id INT PRIMARY KEY,
```
hparfr: why not use `uuid` here or at the opposit only use `id` and get rid of all the joins ?

hparfr (changes requested):
Thanks for addressing this problem.

I'm asking for more comments to give explicit intent.

And I have a question about: why all these joins instead of use id or uuid directly.

queue_job/migrations/16.0.2.7.0/pre-migration.py
sbidoul: @hparfr we probably could but I don't think it would significantly change the main query in `requeue_dead_jobs`. And since we want a foreign key with on delete cascade between job locks and jobs, it's perhaps more intuitive to have it on the primary key, as usual.

hparfr: thanks for the clarification

queue_job/job.py
```diff
@@ -820,6 +853,23 @@ def set_started(self):
         self.date_started = datetime.now()
         self.worker_pid = os.getpid()
 
+        # add job to list of lockable jobs
+        self.env.cr.execute(
```
rousseldenis: Put this in a separate function to clarify code.

queue_job/jobrunner/runner.py
```diff
@@ -343,6 +305,60 @@ def set_job_enqueued(self, uuid):
                 (ENQUEUED, uuid),
             )
 
+    def requeue_dead_jobs(self):
```
rousseldenis: Wasn't it possible to test that ?

queue_job/job.py
```diff
@@ -517,6 +545,11 @@ def perform(self):
 
         The job is executed with the user which has initiated it.
         """
+        if self.max_retries and self.retry >= self.max_retries:
+            raise FailedJobError(
+                "Job: %s, Max. retries (%d) reached" % (self.uuid, self.max_retries)
```
rousseldenis: This should be enhanced. Maybe doing a new exception as the case is quite different as this is triggered by a job execution that has not ended well but without exception (e.g.: worker timeout).

The best (but I don't know if it is feasible) should be to have the exact cause of retry.

rousseldenis (commented):
Some comments

rousseldenis:
@AnizR Could you check last @hparfr comments ?

queue_job/job.py
sbidoul: Perhaps we should simply remove this new logic on retries.

queue_job/jobrunner/runner.py
```diff
@@ -343,6 +305,60 @@ def set_job_enqueued(self, uuid):
...
+        running jobs are interrupted while the runner has no chance to know
+        they have been aborted.
+        """
+
+        with closing(self.conn.cursor()) as cr:
+            query = """
+            UPDATE
+                queue_job
+            SET
+                state='pending',
```
sbidoul: And here set state to error if max retries has been reached?

AnizR:
Thank you for all your valuable feedback!

I appreciate the time you've taken to review this PR.
I will rework the code based on your suggestions and push an update as soon as possible.

sbidoul:
@AnizR another feedback I got today is that the new `queue_job_locks` table got deleted by `database_cleanup`. Maybe we should create a model for it with a proper foreign key, even if we don't actually use the ORM to manipulate it?

queue_job/job.py
AnizR: In deed, I'll remove this

queue_job/job.py
AnizR: Done :smile:

queue_job/job.py
AnizR: Done :+1:

queue_job/migrations/16.0.2.7.0/pre-migration.py
AnizR: Well spotted!

AnizR:
I think that I've addressed every remarks. Thank you all!

I've managed to add some tests on my newly introduced mechanism.
I was forced to introduce some `commit()` in my tests since I am relying on rows and locks in the db.

queue_job/job.py
```diff
@@ -238,6 +238,61 @@ def load_many(cls, env, job_uuids):
         recordset = cls.db_records_from_uuids(env, job_uuids)
         return {cls._load_from_db_record(record) for record in recordset}
 
+    def add_lock_record(self):
+        """
+        Create row in db to be locked once the job is performed
```
sbidoul:
```suggestion
        Create row in db to be locked while the job is being performed.
```

sbidoul:
@AnizR I realized that the AbstractModel technique to create the lock table would not prevent it to be purged by `database_cleanup`. So I made the change to use a regular model on the train back from the sprint. Could you check?

pedrobaeza:
You can add the exception in the `database_cleanup` module as well, as we have done with some columns, although it's true that if there's another way to not require that manual list, it's better.

AnizR:
> @AnizR I realized that the AbstractModel technique to create the lock table would not prevent it to be purged by `database_cleanup`. So I made the change to use a regular model on the train back from the sprint. Could you check?

I did a small test to simulate re-queuing of 'dead jobs'.
It still works correctly, thanks!

hparfr (approved):
Comments added as I requested, thanks @AnizR

rousseldenis:
> > @AnizR I realized that the AbstractModel technique to create the lock table would not prevent it to be purged by `database_cleanup`. So I made the change to use a regular model on the train back from the sprint. Could you check?
> 
> I did a small test to simulate re-queuing of 'dead jobs'. It still works correctly, thanks!

We'll do tests today

lmignon (approved):
Thank you @AnizR for all your hard work. For the last 3 weeks, this PR has been working perfectly in production. We no longer encounter dead  jobs and the retry error handling works even better than before.

AnizR:
> Thank you @AnizR for all your hard work. For the last 3 weeks, this PR has been working perfectly in production. We no longer encounter dead jobs and the retry error handling works even better than before.

I am happy to contribute and see that it works and is used :+1: 

I have also used this PR on a production environment without any issue.
I think that it can be merged

sbidoul:
@amh-mw have your concerns been addressed?

amh-mw (approved):
LGTM

sbidoul:
Let's go.

/ocabot merge minor

sbidoul:
FW port to 17.0 in #748

queue_job/migrations/16.0.2.9.0/pre-migration.py
```diff
@@ -0,0 +1,22 @@
...
+            SET
+                active=False
+            WHERE id IN (
+                SELECT res_id
+                FROM
+                    ir_model_data
+                WHERE
+                    module='queue_job'
+                    AND model='ir.cron'
+                    AND name='ir_cron_queue_job_garbage_collector'
```
baguenth, line 19: @AnizR I guess we have in issue with this migration. I was wondering why I get error messages in my local:

AttributeError: 'queue.job' object has no attribute 'requeue_stuck_jobs'

I figured out, that the migration did no actually deactivate the cron job. The issue is, that as "name" in the sql query you put the id. 

So either you need to provide the actual name which is "Jobs Garbage Collector" or we need to use "openupgrade.delete_records_safely_by_xml_id" with the proper external id. 

Would be great, if you could may create a follow up pull request :)

PR #719, opened 2024-12-16
[14.0][REF] queue_job: Use set.discard(...) instead of SafeSet

florentx:
Simplify


sbidoul (approved):
Yep. Really curious why my previous self did that 10 years ago...

amh-mw (approved):

PR #722, opened 2024-12-17
[18.0][REF] queue_job: Use set.discard(...) instead of SafeSet

florentx:


sbidoul (approved):

amh-mw (changes requested):
Why is the code coverage different on this pull request versus the other ones?

sbidoul:
> Why is the code coverage different on this pull request versus the other ones?

Maybe the doc tests are not running anymore?

PR #732, opened 2024-12-27
[18.0][REF] Forward port from 17.0

florentx:
Only minor divergence

### DIFF
--- a/queue_job/README.rst
+++ b/queue_job/README.rst
@@ -434,7 +434,7 @@ running Odoo**
 When you are developing (ie: connector modules) you might want to bypass
 the queue job and run your code immediately.
 
-To do so you can set QUEUE_JOB\__NO_DELAY=1 in your enviroment.
+To do so you can set QUEUE_JOB\__NO_DELAY=1 in your environment.
 
 **Bypass jobs in tests**
 
--- a/queue_job/delay.py
+++ b/queue_job/delay.py
@@ -232,7 +232,7 @@ class DelayableGraph(Graph):
         elif jobs_count == 1:
             if jobs[0].graph_uuid:
                 raise ValueError(
-                    f"Job {jobs[0]} is a single job, it should not" " have a graph uuid"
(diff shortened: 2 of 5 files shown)


amh-mw (changes requested):
This seems to have lost attribution to the original committer(s)? I don't think that is allowed per AGPL v3 section 7:

```
    b) Requiring preservation of specified reasonable legal notices or
    author attributions in that material or in the Appropriate Legal
    Notices displayed by works containing it; or
```

florentx:
> This seems to have lost attribution to the original committer(s)? I don't think that is allowed per AGPL v3 section 7:
> 
> ```
>     b) Requiring preservation of specified reasonable legal notices or
>     author attributions in that material or in the Appropriate Legal
>     Notices displayed by works containing it; or
> ```

Well, I'm not sure I understand what it refers to.
These are only minor fixes, and I've packed them in a single commit for simplicity.
I am author of the fixes ported from the other branch to 18.0.

leemannd (approved):

sebalix (commented):
@florentx we should avoid the squash of commits when backporting/forward porting commits from one branch to another, otherwise it is very difficult to list fixes/features that are missing.

I would encourage the use of https://github.com/OCA/oca-port/ to port changes from one branch to another, it compares commits between OCA branches based on several criteria like:
- author
- commit date
- commit message

And from there it regroup missing commits by PRs to better understand what needs to be ported, in the right order to avoid conflicts if possible.

sebalix:
I did an attempt here from 14.0 to 18.0 for instance:
- https://github.com/OCA/queue/pull/736

sebalix (changes requested):
Prefer to block this PR, as stated history needs to be preserved.

florentx:
History restored, as suggested.

sebalix (approved):

StephaneMangin (approved):

PR #729, opened 2024-12-27
[16.0][UPD] Forward port changes from 14.0 and 15.0

florentx:
Following PRs were missing in 16.0 branch. They are now forward ported from 14.0 and 15.0 branches:
 - #382
 - #449
 - #466 (keep and update tests)
 - #569
 - #624
 - #643
 - #676
 - #683
 - #728

PR #736, opened 2025-01-13

sebalix:
Port from 14.0 to 16.0:
- #676
- #683

The following PRs have been blacklisted:
- #658: (auto) Nothing to port

### DIFF
--- a/.oca/oca-port/blacklist/queue_job.json
+++ b/.oca/oca-port/blacklist/queue_job.json
@@ -8,6 +8,7 @@
     "440": "Already fixed",
     "466": "Reverting #387",
     "511": "Icon already updated for v16",
-    "537": "Already ported"
+    "537": "Already ported",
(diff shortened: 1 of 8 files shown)


florentx:
Hello @sebalix 

I'm surprised you open this PR without considering this one first:
- #729 

Even if it does not use oca-port, I believe it is more complete, since I tried to be exhaustive for all the missing parts.

sebalix:
@florentx it was not on purpose, sorry, I didn't check first if there was an existing PR!
Yours is more complete because it ports commits of multiple modules at a time indeed.

EDIT: I spotted two commits squashed into one, which is creating false-positive to detect afterwards what is missing, didn't checked the rest.
The goal of `oca-port` is not to only ease the developer life, but to provide reports so we know what to focus on with hundreds of OCA modules to maintain/migrate among different projects.

PR #738, opened 2025-01-17
[14.0][IMP] queue_job: perform_enqueued_jobs should filter the context

florentx:
This enhancement is needed to write more accurate tests when the delayed method is context-sensitive.

We had a bug not identified during tests because of this :(

### DIFF
--- a/test_queue_job/tests/test_delay_mocks.py
+++ b/test_queue_job/tests/test_delay_mocks.py
@@ -268,6 +268,27 @@ class TestDelayMocks(common.SavepointCase):
             self.assertEqual(logs[2].message, "test_trap_jobs_perform graph 3")
             self.assertEqual(logs[3].message, "test_trap_jobs_perform graph 1")
 
+    def test_trap_jobs_prepare_context(self):
+        """Verify that context is filtered consistently.
+
+        Method '_job_prepare_context_before_enqueue_keys' is redefined on 'test.queue.job'.
+        Key 'lang' is whitelisted, while key 'config_key' is filtered out.
+        """
+        # pylint: disable=context-overridden
+        with trap_jobs() as trap:
+            model1 = self.env["test.queue.job"].with_context({"config_key": 42})
+            model2 = self.env["test.queue.job"].with_context(
+                {"config_key": 42, "lang": "it_IT"}
+            )
+            model1.with_delay().testing_method("0", "K", return_context=1)
+            model2.with_delay().testing_method("0", "K", return_context=1)
+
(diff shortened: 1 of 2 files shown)


StefanRijnhart (dismissed):

test_queue_job/tests/test_delay_mocks.py
```diff
@@ -268,6 +268,22 @@ def test_trap_jobs_perform(self):
             self.assertEqual(logs[2].message, "test_trap_jobs_perform graph 3")
             self.assertEqual(logs[3].message, "test_trap_jobs_perform graph 1")
 
+    def test_trap_jobs_prepare_context(self):
+        # pylint: disable=context-overridden
+        with trap_jobs() as trap:
+            model1 = self.env["test.queue.job"].with_context({"config_key": 42})
```
petrus-v, line 279: [`config_key` is not part of keys to forward](https://github.com/OCA/queue/blob/14.0/test_queue_job/models/test_models.py#L42), isn't that your problem ?

I suppose you should rely on queue_job_context module: https://github.com/OCA/queue/blob/14.0/queue_job_context/models/base.py#L8 (which this test_queue_job to not depends on)

petrus-v: hum... in fact your test demonstrate that config_key is not propagated ! so maybe just an explicit dependency to the queue_job_context module would be nice which define the lang attribute.

StefanRijnhart: I believe the issue that is fixed here is that the context is now also filtered of trapped jobs, as is the case in queued jobs in production environments. I agree with the original submitter that this is an improvement because it makes the test simulate production environments more realistically.

StefanRijnhart: Huh you're right in saying that the `lang` key shold not survive the test, because `_job_prepare_context_before_enqueue_keys` does indeed return an empty list in this version.

florentx: Test model `test.queue.job` redefines the whitelisted keys of the context:
https://github.com/florentx/queue/blob/14.0_trap_context/test_queue_job/models/test_models.py#L42-L43

StefanRijnhart: Thanks for clarifying! It would be good to have it documented in the test's docstring that this method is implemented by the test model.

StefanRijnhart (approved):

petrus-v (approved):

PR #742, opened 2025-01-24
[FIX][18.0] queue_job: indicate that run_job need a read/write connection

vincent-hatakeyama:
Without this fix, if db_replica_host is set, Odoo might pass a readonly database cursor and the FOR UPDATE in the method would fail.

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -73,7 +73,13 @@ class RunJobController(http.Controller):
             else:
                 break
 
-    @http.route("/queue_job/runjob", type="http", auth="none", save_session=False)
+    @http.route(
+        "/queue_job/runjob",
+        type="http",
+        auth="none",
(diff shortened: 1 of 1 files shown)


guewen (approved):

sbidoul (approved):

aallouche:
Thanks for the quick merge! Much appreciated.

StefanRijnhart:
@vincent-hatakeyama Can I ask you, I'm curious why this is necessary. https://github.com/odoo/odoo/blob/b2d709b/odoo/http.py#L701-L703 seems to imply that the cursor will default to read/write. Where is the default determined as True?

StefanRijnhart:
@vincent-hatakeyama Oh, never mind. Found it in https://github.com/odoo/odoo/blob/b2d709b/odoo/http.py#L863, when auth='none'.

PR #743, opened 2025-01-24
[18.0][IMP] queue_job: perform_enqueued_jobs should filter the context

florentx:
This enhancement is needed to write more accurate tests when the delayed method is context-sensitive.

Same as:
- Odoo 14 #738

### DIFF
--- a/test_queue_job/tests/test_delay_mocks.py
+++ b/test_queue_job/tests/test_delay_mocks.py
@@ -271,6 +271,12 @@ class TestDelayMocks(common.TransactionCase):
             self.assertEqual(logs[3].message, "test_trap_jobs_perform graph 1")
 
     def test_trap_jobs_prepare_context(self):
+        """Context is transferred to the job according to an allow-list.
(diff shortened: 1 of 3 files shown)


test_queue_job/tests/test_delay_mocks.py
```diff
@@ -270,6 +270,22 @@ def test_trap_jobs_perform(self):
             self.assertEqual(logs[2].message, "test_trap_jobs_perform graph 3")
             self.assertEqual(logs[3].message, "test_trap_jobs_perform graph 1")
 
+    def test_trap_jobs_prepare_context(self):
```
StefanRijnhart, line 273: To help others like me, who are not immediately aware of how context is treated in the queue, can you clarify in a docstring that it is checked that only the whitelisted context keys are preserved when running the job?

StefanRijnhart (approved):
Thanks!

hbrunn (approved):

PR #733, opened 2025-01-08
[18.0][MIG] queue_job_batch, test_queue_job_batch: Migration to 18.0

yankinmax:
**HELP IS NEEDED**

I’ve started the migration and JS code analysis.

The “messaging” JS code was heavily refactored starting v17 with this commit:
- [ ] https://github.com/odoo/odoo/commit/7710c3331ebd22f8396870bd0731f8c1152d9c41

I’m not familiar with the module and don’t know how it’s expected to work.

I’ve checked the JS changes and write some pseudo code:

https://github.com/OCA/queue/pull/733/files#diff-bc23bdd925614b2acfed835f19bb33416f7a6f4ee37a15938971d2c309aa2a0e

ping @ivantodorovich

PR #744, opened 2025-01-27
[18.0][MIG] queue_job_batch, test_queue_job_batch: Migration to 18.0

ivantodorovich:
Superseeds:
- https://github.com/OCA/queue/pull/733

Changes:

* The batch enqueued method is removed, to better reflect the fact that the batch's jobs are executed immediately.
* The batch state `draft` is is renamed to `pending`, for the same purpose.
* The batch's `is_read` field is initialized as `False` (default value), and only set to `True` when it's actually read.
* We now use `identity_exact` for the batch `check_state` delayed method.
* The widget has been reimplemented due to the heavy refactoring made in the core `mail` app: https://github.com/odoo/odoo/commit/7710c333


queue_job_batch/README.rst
```diff
@@ -38,15 +38,13 @@ Example:
 .. code:: python
 
    from odoo import models, fields, api
-   from odoo.addons.queue_job.job import job
+
 
    class MyModel(models.Model):
-      _name = 'my.model'
+       _name = 'my.model'
```
yankinmax, line 44: There will be no documentation problem, but you can probably remove falsy space?

yankinmax (approved):
Technical migration seems ok! Thanks @ivantodorovich

queue_job_batch/models/queue_job_batch.py
```diff
@@ -52,72 +54,75 @@ class QueueJobBatch(models.Model):
...
+                "company_id": self.env.company.id or self.env.user.company_id.id,
             }
         )
-        return self.sudo().create(vals).with_user(self.env.uid)
+        record = self.sudo().create(vals).with_user(self.env.uid)
+        record.user_id._bus_send("queue.job.batch/updated", {"batch_created": True})
+        return record
 
-    @api.depends("job_ids")
+    @api.depends("job_ids.state")
```
simahawk, line 102: do we really need this depends? Job count is not stored.

simahawk (approved):

queue_job_batch/models/queue_job_batch.py
ivantodorovich: Not really, for the UI. I'd be necessary when reading the field programatically, though, without invalidating caches. It's a habit 😓

mpurnell1:
Any update on when this might be merged?

PR #748, opened 2025-02-28
[17.0][PORT] queue_job: remove dead jobs requeuer cron and automatically requeue dead jobs

sbidoul:
FW port of #716


AnizR (changes requested):
Thanks for the port forward!

No remarks to do regarding the logic, it still works: queue_job isn't different between v16 and v17.

Small question: Shouldn't we need to change the module version since we are adding a migration script?

sbidoul:
> Shouldn't we need to change the module version since we are adding a migration script?

I was thinking to merge with a minor bump, but changing it explicitly is safer. Done.

AnizR (approved):
Ok, nice. Everything looks good to me now :+1:

guewen (approved):

hparfr (approved):

paulius-sladkevicius:
Note that it wasn't ported to v18 yet.

queue_job/jobrunner/runner.py
```diff
@@ -343,6 +306,93 @@ def set_job_enqueued(self, uuid):
...
 
+    def _query_requeue_dead_jobs(self):
+        return """
+            UPDATE
+                queue_job
+            SET
+                state=(
+                    CASE
+                        WHEN
+                            max_retries IS NOT NULL AND
```
sersanchus, line 317: Maybe there's a problem here? max_retries can be 0 and is well documented that it has to retry forever...

sbidoul: Possibly, indeed. cc/ @AnizR

zamberjo: There's a possible solution in https://github.com/OCA/queue/pull/816, mind taking a look? 

@sbidoul @AnizR

PR #749, opened 2025-03-03
[18.0][PORT] queue_job: remove dead jobs requeuer cron and automatically requeue dead jobs

sbidoul:
FW port of #748 + #816


hbrunn:
@sbidoul looks like you have to rewrite this to use savepoints?

sbidoul:
I'm note sure it's possible to test that with savepoints, since we test database locks.

@AnizR

queue_job/tests/test_requeue_dead_job.py
```diff
@@ -0,0 +1,133 @@
...
+
+        job_obj.set_started()
+        job_obj.store()
+
+        locks = self.get_locks(job_obj.uuid)
+
+        self.assertEqual(1, len(locks))
+
+        # commit to update queue_job records in DB
+        self.env.cr.commit()  # pylint: disable=E8102
```
sbidoul: @AnizR in this test, is it necessary to commit?

queue_job/tests/test_requeue_dead_job.py
```diff
@@ -0,0 +1,133 @@
...
+        queue_job = self.create_dummy_job(uuid)
+        job_obj = Job.load(self.env, queue_job.uuid)
+
+        job_obj.set_enqueued()
+        # simulate enqueuing was in the past
+        job_obj.date_enqueued = datetime.now() - timedelta(minutes=1)
+        job_obj.set_started()
+
+        job_obj.store()
+        self.env.cr.commit()  # pylint: disable=E8102
```
sbidoul: Here we could maybe create a job with demo data, and run this test post install.

queue_job/tests/test_requeue_dead_job.py
AnizR: Yes, otherwise, the row representing a 'lock' will not be created within the db

queue_job/tests/test_requeue_dead_job.py
AnizR: Using data to create some records within the db might be a way to avoid to commit within tests, I'll think about it :+1:

queue_job/migrations/18.0.1.3.0/pre-migration.py
```diff
@@ -0,0 +1,22 @@
...
+            SET
+                active=False
+            WHERE id IN (
+                SELECT res_id
+                FROM
+                    ir_model_data
+                WHERE
+                    module='queue_job'
+                    AND model='ir.cron'
+                    AND name='ir_cron_queue_job_garbage_collector'
```
baguenth: You may directly want to consider this fix: https://github.com/OCA/queue/pull/762/commits/46b0d8f91ef54b777dde55eb0d6881c90671d170

AnizR:
@hbrunn, I came up with the solution of loading test `queue.job` records within `demo`.

What do you think about it?

queue_job/demo/queue_job_test_lock.xml
```diff
@@ -0,0 +1,18 @@
...
+        name="_create_test_started_job"
+        model="queue.job"
+        eval="('test_started_job',)"
+    />
+    <function
+        name="_create_test_enqueued_job"
+        model="queue.job"
+        eval="('test_enqueued_job',)"
+    />
+</odoo>
```
sbidoul: @AnizR this file should be removed as it was moved to test_queue_job?

AnizR: Yes, well spotted!

simahawk (approved):
LGTM. Remember to merge w/ `nobump`.

sbidoul:
Setting as draft. There is a conflicts and the tests need some rework.

AnizR:
I reworked the tests and decided to isolate the tests in `test_queue_job`.
Moreover, I managed to remove the `.commit()` calls.

FrancoMaxime (commented):
LGTM: code review

lmarion-source (approved):

PR #668, opened 2024-07-02
[16.0][IMP] queue_job: HA job runner using session level advisory lock

sbidoul:
Another attempt.

closes #422

PR #750, opened 2025-03-05
[IMP] queue_job_cron_jobrunner: channel

Koudja:
Implement channel on queue_job_cron_jobrunner

### DIFF
--- a/queue_job_cron_jobrunner/models/queue_job.py
+++ b/queue_job_cron_jobrunner/models/queue_job.py
@@ -8,6 +8,7 @@ from collections import defaultdict
 from datetime import datetime
 from io import StringIO
 
+import psutil
 from psycopg2 import OperationalError
 
 from odoo import _, api, fields, models, tools
@@ -162,6 +163,7 @@ class QueueJob(models.Model):
     @api.model
     def _job_runner(self, commit=True):
         """Short-lived job runner, triggered by async crons"""
+        self._release_started_jobs(commit=commit)
         job = self._acquire_one_job(commit=commit)
         while job:
             job._process(commit=commit)
@@ -214,6 +216,24 @@ class QueueJob(models.Model):
         delayed_etas = {rec.eta for rec in records if rec.eta}
         if delayed_etas:
             self._cron_trigger(at=list(delayed_etas))
+
+    @api.model
+    def _release_started_jobs(self, commit=False):
+        pids = [x.pid for x in psutil.process_iter()]
+        for record in self.search(
+            [("state", "=", "started"), ("worker_pid", "not in", pids)]
+        ):
+            job = Job._load_from_db_record(record)
+            job.set_pending()
+            job.store()
+            _logger.info(
+                "release started job %s[channel=%s,uuid=%s]",
+                record.id,
+                record.channel,
+                record.uuid,
+            )
+        if commit:  # pragma: no cover
+            self.env.cr.commit()  # pylint: disable=invalid-commit
 
     @api.model_create_multi
     def create(self, vals_list):
--- a/queue_job_cron_jobrunner/models/queue_job.py
+++ b/queue_job_cron_jobrunner/models/queue_job.py
@@ -4,6 +4,7 @@
 
 import logging
 import traceback
+from collections import defaultdict
 from datetime import datetime
 from io import StringIO
 
@@ -19,6 +20,7 @@ from odoo.addons.queue_job.exception import (
     RetryableJobError,
 )
 from odoo.addons.queue_job.job import Job
+from odoo.addons.queue_job.jobrunner import QueueJobRunner
 
 _logger = logging.getLogger(__name__)
 
@@ -27,14 +29,12 @@ class QueueJob(models.Model):
     _inherit = "queue.job"
 
     @api.model
-    def _acquire_one_job(self):
+    def _acquire_one_job(self, commit=False):
         """Acquire the next job to be run.
 
         :returns: queue.job record (locked for update)
         """
-        # TODO: This method should respect channel priority and capacity,
-        #       rather than just fetching them by creation date.
-        self.flush()
+        runner = QueueJobRunner.from_environ_or_config()
         self.env.cr.execute(
             """
             SELECT id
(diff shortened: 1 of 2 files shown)


queue_job_cron_jobrunner/models/queue_job.py
```diff
@@ -32,42 +33,77 @@ def _acquire_one_job(self):
...
         )
-        row = self.env.cr.fetchone()
-        return self.browse(row and row[0])
+        rows = self.env.cr.fetchall()
+
+        channels = {}
+        for queue_job in self.search([("state", "=", "started")]):
+            if not queue_job.channel:
+                continue
+            channels[queue_job.channel] = channels.get(queue_job.channel, 0) + 1
```
alexandregaldeano: **suggestion:** what do you think?
```suggestion
        channels = defaultdict(int)
        for queue_job in self.search([("state", "=", "started")]):
            if not queue_job.channel:
                continue
            channels[queue_job.channel] += 1
```

queue_job_cron_jobrunner/models/queue_job.py
```diff
@@ -32,42 +33,77 @@ def _acquire_one_job(self):
...
+            _logger.info(
+                "_acquire_one_job queue.job %s[channel=%s,uuid=%s] started",
+                row[0],
+                job.channel,
+                job.uuid,
+            )
+            result = queue_job
+            break
+        self.flush()
+        self.env.cr.commit()  # pylint: disable=E8102
```
alexandregaldeano: **suggestion:** Maybe add a comment to explain why you can't use a different cursor (lock never released by Odoo)?

alexandregaldeano (approved):
Once the tests are fixed, LGTM! 🚀

petrus-v (approved):
Nice one, thanks !

petrus-v:
This PR require https://github.com/OCA/queue/pull/754 to avoid opened file descriptor leak !

damdam-s (approved):

hbrunn:
@Koudja this is still in draft state, is this intentional?

petrus-v:
> @Koudja this is still in draft state, is this intentional?

In fact we are still supervising the system and keep improving it as long we met difficulties ! yesterday we've implement the stop_processing feature to avoid to be kill by the cpu time limit in case there is a lot of job in the queue, this is not yet deployed to our production (and I'm going to somme logging).

So I suppose keeping it in draft state as long we have not get some stabilities is probably better ! But review and feedback are welcome otherwise so not sure what suits most OCA conventions in such case ?

hbrunn:
draft is good I think, because of the label I nearly merged it

queue_job_cron_jobrunner/models/queue_job.py
```diff
@@ -113,18 +162,47 @@ def _process(self, commit=False):
...
+            )
+        )
+        stop_processing_threshold_seconds = int(
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param(
+                "queue_job_cron_jobrunner.stop_processing_threshold_seconds", "5"
+            )
+        )
+        if next_cron_job_runner_trigger.nextcall <= (
```
petrus-v: after few investigation, I've the feeling the nextcall is not yet uppdated so we are probably dealing with the current started date so we would have to compute it...

petrus-v: because odoo changed it on a different cursor !...

sbidoul:
I'm wondering... now that the job scheduler (misnamed jobrunner) can run on odoo.sh (#668), queue_job_cron_jobrunner could maybe be simplified to only run jobs that are in state `enqueued`, and keep all the scheduling logic (channels, priorities, eta, etc) in the regular jobrunner ?

petrus-v:
> I'm wondering... now that the job scheduler (misnamed jobrunner) can run on odoo.sh (#668), queue_job_cron_jobrunner could maybe be simplified to only run jobs that are in state `enqueued`, and keep all the scheduling logic (channels, priorities, eta, etc) in the regular jobrunner ?

As of today, we are not using the mentioned patch and have configured our setup to avoid spawning the job scheduler/runner, as suggested in the module’s documentation.

To be honest, we haven’t spent much time evaluating this patch, so we might have missed some key information. However, we do have some concerns:

  *  A preference for using a cron worker to process queue jobs.
  *  Uncertainty about how Odoo.sh manages process lifecycles, especially regarding jobrunner. If it gets killed (due to inactivity, lack of HTTP requests, or other obscure platform limits), how does the system handle it? Would it be properly restarted by Odoo.sh?
  
If I understand correctly, your idea is to have the queue jobrunner focus on managing job states—ensuring that tasks are enqueued while respecting channel capacities—while letting the ir.cron execute them based on priority. This approach would remove the need for HTTP workers to process jobs.

That sounds like an interesting direction and avoid to duplicate the channel implementation logic. Would you be open to further designing this together or discussing the long-term vision for the module?

sbidoul:
> If I understand correctly, your idea is to have the queue jobrunner focus on managing job states—ensuring that tasks are enqueued while respecting channel capacities—while letting the ir.cron execute them based on priority. This approach would remove the need for HTTP workers to process jobs.

Yes, that's the idea.

> Uncertainty about how Odoo.sh manages process lifecycles, especially regarding jobrunner. If it gets killed (due to inactivity, lack of HTTP requests, or other obscure platform limits), how does the system handle it? Would it be properly restarted by Odoo.sh?

I'm not sure either. On the project where we use the HA jobrunner on odoo.sh, we don't observe issues with this, but yeah, "obscure platform limits" are the key words. But I tend to think this is not a blocker for that approach.

> That sounds like an interesting direction and avoid to duplicate the channel implementation logic. Would you be open to further designing this together or discussing the long-term vision for the module?

We don't need this so far :crossed_fingers:, but I'd be happy to discuss.

queue_job_cron_jobrunner/models/queue_job.py
Koudja: (doesn't seem useful anymore)

Koudja:
I updated PR with latest comments. It's been on our production for few weeks now without any issue (with thousands jobs per day). I updated this PR state to ready for merge based on this.

Koudja:
We've migrated on HTTP worker, so I am closing this PR.

PR #754, opened 2025-03-13
[FIX] queue_job: job runner open pipe that are never closed properly

petrus-v:
this make a lot of open file descriptors that we got the limit while instentiate jobrunner from queue_job_cron_jobrunner in the current channel implementation https://github.com/OCA/queue/pull/750

### DIFF
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -367,6 +367,16 @@ class QueueJobRunner:
         self._stop = False
         self._stop_pipe = os.pipe()
 
+    def __del__(self):
+        try:
+            os.close(self._stop_pipe[0])
+        except OSError:
+            pass
+        try:
+            os.close(self._stop_pipe[1])
+        except OSError:
+            pass
+
     @classmethod
     def from_environ_or_config(cls):
         scheme = os.environ.get("ODOO_QUEUE_JOB_SCHEME") or queue_job_config.get(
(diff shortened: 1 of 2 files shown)


damdam-s (approved):

amh-mw (approved):
Code review only, LGTM. If this gets forward-ported to 15 or 18, I would be willing to graph file descriptor counts in one of my long running test environments and report back.

petrus-v:
@amh-mw Thanks for review

> Code review only, LGTM. If this gets forward-ported to 15 or 18, I would be willing to graph file descriptor counts in one of my long running test environments and report back.
 
I don't expect you get that mush open fd using queue_job as it. We discover it by using queue_job_cron_jobrunner and this recent PR  https://github.com/OCA/queue/pull/750 where a [`QueueJobRunner` instance is created while acquiring job](https://github.com/OCA/queue/pull/750/files#diff-e0bfcb5a4868e4856304442bdf2a6372bbd3668e4101150804e280fbe8df5dabR37) so we get the limit while handling thousands of jobs quickly. Indeed this is a bit mitigate as workers are restart at some point.

Anyway it's an esay way to mitigate a denial of service attack (assuming attacker is able to execute such code `from odoo.addons.queue_job.jobrunner import QueueJobRunner; [QueueJobRunner.from_environ_or_config() for _  in range(2048)]` ... well I suppose if an attacker is able to do that it can do more dommage on the server ! ).

simahawk (approved):

PR #758, opened 2025-04-04
[17.0][IMP] queue_job: add Job to recordset ctx

SilvioC2C:
Having the UUID is enough for retrieving the original Job instance via ``Job.load(env, uuid)`` and read its attributes. But the returned Job instance is actually a different one than the Job instance that triggered the function, because loading creates a new instance instead. Therefore, operations beside reading the Job attributes will result in unexpected behaviors (for example: dependencies are not stored correctly).

This commit allows retrieving the Job instance from the context itself to make it usable from within the running method.

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -756,7 +756,7 @@ class Job:
 
     @property
     def func(self):
-        recordset = self.recordset.with_context(job_uuid=self.uuid)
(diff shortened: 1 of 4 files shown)


Camille0907 (approved):
LGTM, thanks

simahawk (changes requested):
I'm not sure to understand why you need this.

PR #759, opened 2025-04-10
[FIX] queue_job: Adapted migration to properly delete the invalid que…

baguenth:
In https://github.com/OCA/queue/commit/3e5645bc23d03a4aff4b9713d7485c21ee2c05ff we got rid of the garbage collector logic for stuck queue jobs and therefore deactivated the old cron job, which was handling the garbage collection in the past. 

Seems like the migration added, to deactivate the cron job, doesn't properly work. At least in my environments the cron job is still active and lead to error messages in the log like:

image

Since we actually don't need this cron job anymore, I guess we can simply remove it entirely.


sbidoul:
@AnizR

AnizR (approved):
Thanks, for fixing this issue!

sbidoul:
@baguenth do you plan to forward port?

baguenth:
> @baguenth do you plan to forward port?

@sbidoul @AnizR Here is the port forward for 17.0: https://github.com/OCA/queue/pull/762 

In 18.0, the logic for stuck jobs was not yet adapted.

PR #762, opened 2025-04-15
[FIX] queue_job: Adapted migration to properly delete the invalid cro…

baguenth:
…n job

Port forward for https://github.com/OCA/queue/pull/759


sbidoul:
Thanks. For 18, there is https://github.com/OCA/queue/pull/749 where the tests need to be reworked a bit.

/ocabot merge nobump

PR #764, opened 2025-04-28
[16.0][UPD] add empty batch remove when job vaccuum

julienCozo:
When queue jobs performing its vaccuum, delete also queue batch where job_ids is 0.

### DIFF
--- a/queue_job_batch/models/queue_job.py
+++ b/queue_job_batch/models/queue_job.py
@@ -30,3 +30,18 @@ class QueueJob(models.Model):
             # to work with the same batch
             batch.with_delay().check_state()
         return super().write(vals)
+
+    def autovacuum(self):
+        """Delete batches if all jobs are unlink after job autovacuum
+
+        Called from a cron.
+        """
+        super().autovacuum()
+        batches = self.env["queue.job.batch"].search([])
+
+        for batch in batches:
+            if not batch.job_ids:
+                batch.unlink()
+        self.env.cr.commit()  # pylint: disable=E8102
+
+        return True


queue_job_batch/models/queue_job.py
```diff
@@ -30,3 +30,18 @@ def write(self, vals):
...
+
+        Called from a cron.
+        """
+        super().autovacuum()
+        batches = self.env["queue.job.batch"].search([])
+
+        for batch in batches:
+            if not batch.job_ids:
+                batch.unlink()
+        self.env.cr.commit()  # pylint: disable=E8102
```
AnizR, line 45: Why doing a commit there?

queue_job_batch/models/queue_job.py
```diff
@@ -30,3 +30,18 @@ def write(self, vals):
             # to work with the same batch
             batch.with_delay().check_state()
         return super().write(vals)
+
+    def autovacuum(self):
+        """Delete batches if all jobs are unlink after job autovacuum
+
+        Called from a cron.
+        """
+        super().autovacuum()
```
AnizR, line 39: A good practice is to return the call of the super:
```suggestion
        ret = super().autovacuum()
        
        ...
        
        return ret
```

queue_job_batch/models/queue_job.py
```diff
@@ -30,3 +30,18 @@ def write(self, vals):
...
+        batches = self.env["queue.job.batch"].search([])
```
AnizR, line 40:
```suggestion
        batches = self.env["queue.job.batch"].search([('job_ids','=',False)])
```

AnizR (changes requested):
Nice initiative, I think that it makes sense to remove old batches

Can you check why your tests aren't working?

remi-filament:
@julienCozo why would you remove your branch ? Could you please attend comments from @AnizR and reopen PR ?

PR #767, opened 2025-05-07
[18.0][IMP] queue_job: Default subchannel capacity and sequential.

amh-mw:
This adds `capacity_default` and `sequential_default` options that set `capacity` and `sequential` for autocreated subchannels. It also allows non-root channel configurations to omit capacity.

### DIFF
--- a/queue_job/jobrunner/channels.py
+++ b/queue_job/jobrunner/channels.py
@@ -408,7 +408,15 @@ class Channel:
     """
 
     def __init__(
-        self, name, parent, capacity=None, sequential=False, throttle=0, paused=False
+        self,
+        name,
+        parent,
+        capacity=None,
+        sequential=False,
+        throttle=0,
+        paused=False,
+        capacity_default=None,
+        sequential_default=False,
     ):
         self.name = name
         self.parent = parent
@@ -419,9 +427,13 @@ class Channel:
         self._running = set()
         self._failed = set()
         self._pause_until = 0  # utc seconds since the epoch
-        self.capacity = capacity
+        self.capacity = (
+            capacity if (capacity is not None) else (parent and parent.capacity_default)
+        )
+        self.capacity_default = capacity_default
         self.throttle = throttle  # seconds
-        self.sequential = sequential
+        self.sequential = sequential or (parent and parent.sequential_default)
+        self.sequential_default = sequential_default
         self.paused = paused
 
     @property
@@ -438,13 +450,17 @@ class Channel:
         Supported keys are:
 
         * capacity
+        * capacity_default (default for sub channels)
         * sequential
+        * sequential_default (default for sub channels)
         * throttle
         * paused
         """
         assert self.fullname.endswith(config["name"])
         self.capacity = config.get("capacity", None)
+        self.capacity_default = config.get("capacity_default", None)
         self.sequential = bool(config.get("sequential", False))
+        self.sequential_default = config.get("sequential_default", False)
         self.throttle = int(config.get("throttle", 0))
         self.paused = bool(config.get("paused", False))
         if self.sequential and self.capacity != 1:
@@ -902,22 +918,23 @@ class ChannelManager:
                 continue
             config = {}
             config_items = split_strip(channel_config_string, ":")
-            name = config_items[0]
-            if not name:
+            if not (name := config_items.pop(0)):
                 raise ValueError(
                     f"Invalid channel config {config_string}: missing channel name"
                 )
             config["name"] = name
-            if len(config_items) > 1:
-                capacity = config_items[1]
+            if len(config_items) > 0:
                 try:
-                    config["capacity"] = int(capacity)
+                    config["capacity"] = int(config_items[0])
+                    config_items.pop(0)
                 except Exception as ex:
-                    raise ValueError(
-                        f"Invalid channel config {config_string}: "
-                        f"invalid capacity {capacity}"
-                    ) from ex
-                for config_item in config_items[2:]:
+                    if name == "root":
+                        raise ValueError(
+                            f"Invalid channel config {config_string}: "
+                            f"invalid capacity {config_items[0]}"
+                        ) from ex
+
+                for config_item in config_items:
                     kv = split_strip(config_item, "=")
                     if len(kv) == 1:
                         k, v = kv[0], True
@@ -932,7 +949,16 @@ class ChannelManager:
                         raise ValueError(
                             f"Invalid channel config {config_string}: duplicate key {k}"
                         )
-                    config[k] = v
+                    if k == "capacity_default":
+                        try:
+                            config[k] = int(v)
+                        except Exception as ex:
+                            raise ValueError(
+                                f"Invalid channel config {config_string}: "
+                                f"invalid capacity_default {v}"
+                            ) from ex
+                    else:
+                        config[k] = v
             else:
                 config["capacity"] = 1
             res.append(config)
@@ -945,6 +971,17 @@ class ChannelManager:
         >>> c = cm.get_channel_by_name('root')
         >>> c.capacity
         1
+
+        >>> cm.simple_configure('root:bogus')
+        Traceback (most recent call last):
+            ...
+        ValueError: Invalid channel config root:bogus: invalid capacity bogus
+
+        >>> cm.simple_configure('root:4,:2')
+        Traceback (most recent call last):
+            ...
+        ValueError: Invalid channel config root:4,:2: missing channel name
+
         >>> cm.simple_configure('root:4,autosub.sub:2,seq:1:sequential')
         >>> cm.get_channel_by_name('root').capacity
         4
@@ -961,7 +998,28 @@ class ChannelManager:
         1
         >>> cm.get_channel_by_name('seq').sequential
         True
-        """
+
+        >>> cm.simple_configure('root:4:capacity_default=bogus')
+        Traceback (most recent call last):
+            ...
+        ValueError: Invalid channel config root:4:capacity_default=bogus: invalid capacity_default bogus
+
+        >>> cm.simple_configure('root:4,sub:3:capacity_default=2')
+        >>> cm.get_channel_by_name('root.sub').capacity
+        3
+        >>> cm.get_channel_by_name('root.sub.auto', autocreate=True).capacity
+        2
+
+        >>> cm.simple_configure('root:4,seq:2:sequential')
+        Traceback (most recent call last):
+            ...
+        ValueError: A sequential channel must have a capacity of 1
+
+        >>> cm.simple_configure('root:4,seq:sequential_default')
+        >>> cm.get_channel_by_name('root.seq.auto', autocreate=True).sequential
+        True
+
+        """  # noqa: E501
         for config in ChannelManager.parse_simple_config(config_string):
             self.get_channel_from_config(config)
 
--- a/queue_job/tests/test_runner_channels.py
+++ b/queue_job/tests/test_runner_channels.py
@@ -12,6 +12,7 @@ from odoo.addons.queue_job.jobrunner import channels
 @tagged("doctest")
 class TestDoctest(BaseCase):
     def test_doctest(self):
-        doctest.testmod(
+        results = doctest.testmod(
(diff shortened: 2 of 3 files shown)


queue_job/tests/test_runner_channels.py
```diff
@@ -1,10 +1,43 @@
...
 load_tests = load_doctests(channels)
+
+
+class TestChannelManager(BaseCase):
+    def test_subcapacity_default(self):
+        cm = channels.ChannelManager()
+        cm.simple_configure("root:4")
+        root = cm.get_channel_by_name("root")
+        self.assertEqual(root.capacity, 4)
+        self.assertEqual(root.subcapacity, 0)
```
sbidoul: What is the reason for the default subcapacity to be 0 instead of None?

amh-mw: Personal preference for subcapacity value to be contained by the set of whole numbers rather than the set of counting numbers plus None.

sbidoul: But the default for capacity is `None` so I'd do the same for subcapacity, if only for consistency.
Also, I think we could say there is a distinction between a capacity of 0 (a channel that is paused?) vs an undefined capacity.

queue_job/jobrunner/channels.py
```diff
@@ -410,11 +419,13 @@ def configure(self, config):
 
         * capacity
         * sequential
+        * subcapacity
         * throttle
         """
         assert self.fullname.endswith(config["name"])
         self.capacity = config.get("capacity", None)
         self.sequential = bool(config.get("sequential", False))
+        self.subcapacity = int(config.get("subcapacity", 0)) or None
```
amh-mw: I perceive this `0 or None` falsy coercion as contrived. I'm open to a better idiom.

florentx: Doing it during configuration parsing is an alternative. Then you can have same thing as for `capacity`.

```suggestion
        self.subcapacity = config.get("subcapacity", None)
```

```diff
--- a/queue_job/jobrunner/channels.py
+++ b/queue_job/jobrunner/channels.py
@@ -897,7 +897,16 @@ class ChannelManager:
                             f"Invalid channel config {config_string}: "
                             f"duplicate key {k}"
                         )
-                    config[k] = v
+                    if k == "subcapacity":
+                        try:
+                            config[k] = int(v)
+                        except Exception as ex:
+                            raise ValueError(
+                                f"Invalid channel config {config_string}: "
+                                f"invalid subcapacity {v}"
+                            ) from ex
+                    else:
+                        config[k] = v
             else:
                 config["capacity"] = 1
             res.append(config)
```

sbidoul (commented):
This looks good, thanks!

Could expand the docstring of `parse_simple_config` a bit (around line 823).

The only question I'd have is if we can come up for a better name to convey "default subchannel capacity". I don't have any better idea for now.

amh-mw:
> The only question I'd have is if we can come up for a better name to convey "default subchannel capacity". I don't have any better idea for now.

Given existing documentation:
> Channels: give a capacity for the root channel and its sub-channels and segregate jobs in them. Allow for instance to restrict heavy jobs to be executed one at a time while little ones are executed 4 at a times.

Noodling around in tests:
```python
cm = channels.ChannelManager()
cm.simple_configure("root:2,child")
root = cm.get_channel_by_name("root")
self.assertEqual(root.capacity, 2)
child = cm.get_channel_by_name("child")
self.assertEqual(child.capacity, 1)
auto = cm.get_channel_by_name("auto", autocreate=True)
self.assertEqual(auto.capacity, None)
```

The difference in capacity between an autocreated subchannel and a configured subchannel was the impetus for the creation of this pull request. Maybe rename `subcapacity` to `autocapacity`? Testing shows that the new setting does not change the capacity for configured subchannels.

```python
cm = channels.ChannelManager()
cm.simple_configure("root:4:subcapacity=2,child")
root = cm.get_channel_by_name("root")
self.assertEqual(root.capacity, 4)
child = cm.get_channel_by_name("child")
self.assertEqual(child.capacity, 1)
```

VecSec:
Hi, awesome feature great job. May you merge? @amh-mw

hoangtrann (approved):

queue_job/jobrunner/channels.py
```diff
@@ -874,7 +885,16 @@ def parse_simple_config(cls, config_string):
...
                         )
-                    config[k] = v
+                    if k == "subcapacity":
+                        try:
+                            config[k] = int(v)
+                        except Exception as ex:
+                            raise ValueError(
+                                f"Invalid channel config {config_string}: "
+                                f"invalid subcapacity {v}"
+                            ) from ex
```
sbidoul, line 895: This error handling is nice but I think it's better to move it to the `configure()` method.

sbidoul (commented):
I still don't have a better idea for a config name. `default_subchannel_capacity` might be better but it's a bit long?

Can you add doctests in the `ChannelManager` class? You can look at how it is done for the `throttle` config to get inspiration.

queue_job/jobrunner/channels.py
sbidoul: Oh I had not seen this resolved comment. I personally prefer to keep this in the `configure` method, if only for symmetry with how the other options are handled.

How about `self.subcapacity = int(config.get("subcapacity") if "subcapacity" in config else None` ?

queue_job/jobrunner/channels.py
amh-mw: I don't follow. This is nearly identical to the "capacity" error handling earlier in this same method.

sbidoul: Hm, right, so this was already messy before, with the type conversions spread between `parse_simple_config()` and `configure()` for different options.

I'd prefer to have all the type conversions in the same place, though, so let's not make it worse here? I can refactor in a followup, so as you prefer.

abhinavvaidya-tsd (approved):

sbidoul:
While discussing this with a colleague we found a use case where it would be interesting to configure subchannels with default capacity 1 + sequential, so maybe we should generalize this PR.

amh-mw:
> While discussing this with a colleague we found a use case where it would be interesting to configure subchannels with default capacity 1 + sequential, so maybe we should generalize this PR.

@sbidoul If you would be so kind as to create a new issue for that use case, I'd be happy to take a look, but I am loathe to add features to an `approved` pull request that is about to turn one year old. 🎉

sbidoul:
Hi @amh-mw, I've  looked at this again and I think my last comments still stand.

- I would like the test to be added as doctests to `ChannelManager.simple_configure`, like other similar tests, instead of new test classes. I'm aware the doctests currently don't run in CI but they can be run manually, and it's important to keep similar things together.
- Can you please move the parsing of the `subcapacity` config value to the `configure` method, like it is done for `throttle`, to keep the same code style. Except for `capacity` which is special for the config parser, the `parse_simple_config` method does not have specific knowledge of the config keys.
- Since we will likely want to extend to other default config keys for subchannels, I propose to rename it to `def_sub_capacity`. So we can later add `def_sub_sequential`, etc.

queue_job/jobrunner/channels.py
amh-mw: Let's do it in a follow up. The "int or None" behavior deserves a dedicated refactoring to pure integers.

amh-mw:
> While discussing this with a colleague we found a use case where it would be interesting to configure subchannels with default capacity 1 + sequential, so maybe we should generalize this PR.

I went ahead and added `sequential_default`.

queue_job/jobrunner/channels.py
```diff
@@ -404,7 +404,16 @@ class Channel:
...
-    def __init__(self, name, parent, capacity=None, sequential=False, throttle=0):
+    def __init__(
+        self,
+        name,
+        parent,
+        capacity=None,
+        sequential=False,
+        throttle=0,
+        capacity_default=None,
+        sequential_default=None,
```
sbidoul:
```suggestion
        sequential_default=False,
```

or is it `self.sequential_default = config.get("sequential_default", False)` at line 457 that should have None as default?

queue_job/jobrunner/channels.py
```diff
@@ -414,9 +423,13 @@ def __init__(self, name, parent, capacity=None, sequential=False, throttle=0):
         self._running = set()
         self._failed = set()
         self._pause_until = 0  # utc seconds since the epoch
-        self.capacity = capacity
+        self.sequential = sequential or (parent and parent.sequential_default)
+        self.sequential_default = sequential_default
+        self.capacity = (
+            (self.sequential and 1) or capacity or (parent and parent.capacity_default)
+        )
```
sbidoul, line 432: I worry this looses the capability to set `capacity=0`, which is a way to pause a channel.

queue_job/jobrunner/channels.py
```diff
@@ -432,12 +445,16 @@ def configure(self, config):
         Supported keys are:
 
         * capacity
+        * capacity_default
```
sbidoul:
```suggestion
        * capacity_default (default for sub channels)
```

queue_job/jobrunner/channels.py
```diff
@@ -432,12 +445,16 @@ def configure(self, config):
...
         * sequential
+        * sequential_default
```
sbidoul:
```suggestion
        * sequential_default (default for sub channels)
```

queue_job/jobrunner/channels.py
amh-mw: Sequential is a boolean; None was copy pasta error.

queue_job/jobrunner/channels.py
amh-mw: Roger that; will honor explicit capacity first.

queue_job/jobrunner/channels.py
```diff
@@ -926,7 +964,35 @@ def simple_configure(self, config_string):
...
+            ...
+        ValueError: Invalid channel config root:4:capacity_default=bogus: invalid capacity_default bogus
+
+        >>> cm.simple_configure('root:4,sub:3:capacity_default=2')
+        >>> cm.get_channel_by_name('root.sub').capacity
+        3
+        >>> cm.get_channel_by_name('root.sub.auto', autocreate=True).capacity
+        2
+
+        `sequential`
```
sbidoul:
```suggestion
```

This looks strange?

queue_job/jobrunner/channels.py
```diff
@@ -926,7 +964,35 @@ def simple_configure(self, config_string):
         1
         >>> cm.get_channel_by_name('seq').sequential
         True
-        """
+
+        `capacity_default`
```
sbidoul: ### 
```suggestion
```

queue_job/jobrunner/channels.py
```diff
@@ -926,7 +964,35 @@ def simple_configure(self, config_string):
...
+        2
+
+        `sequential`
+        >>> cm.simple_configure('root:4,seq:2:sequential')
+        Traceback (most recent call last):
+            ...
+        ValueError: A sequential channel must have a capacity of 1
+
+        `sequential_default`
+
```
sbidoul:
```suggestion
```

?

amh-mw: Removing low budget headers.

sbidoul (approved):
Thanks! This is a useful feature.

amh-mw:
Hold off for a second, test logs have
```
**********************************************************************
File "/mnt/extra-addons/oca/queue/queue_job/jobrunner/channels.py", line 986, in odoo.addons.queue_job.jobrunner.channels.ChannelManager.simple_configure
Failed example:
    cm.get_channel_by_name('root.seq.auto', autocreate=True).capacity
Expected:
    1
Got nothing
**********************************************************************
1 items had failures:
   1 of  23 in odoo.addons.queue_job.jobrunner.channels.ChannelManager.simple_configure
***Test Failed*** 1 failures.
```
but no error.

amh-mw:
I removed the defaulting of capacity=1 when sequential=True and gave doctest its teeth back.

sbidoul (approved):

guewen (approved):

PR #768, opened 2025-05-12
[18.0][IMP] queue_job: Improve context management.

amh-mw:
When the `queue_job_keep_context` context variable is set, always
    preserve the entire context. This honors the principle of least
    surprise, in that a developer can easily convert a record.method()
    call to record.with_delay().method() with the expectation that it will
    actually execute the same, simply at a later time.
    
    When the `queue_job_context_keys` context variable is set, preserve
    the given keys in addition to those normally specified by
    `_job_prepare_context_before_enqueue_keys`.

Fixes #406. Formerly #613.

### DIFF
--- a/queue_job/models/base.py
+++ b/queue_job/models/base.py
@@ -245,15 +245,23 @@ class Base(models.AbstractModel):
 
     @api.model
     def _job_prepare_context_before_enqueue_keys(self):
-        """Keys to keep in context of stored jobs
-        Empty by default for backward compatibility.
-        """
-        return ("tz", "lang", "allowed_company_ids", "force_company", "active_test")
+        """Keys to keep in context of stored jobs"""
+        keys = self.env.context.get("queue_job_context_keys")
+        return (
+            "tz",
+            "lang",
+            "allowed_company_ids",
+            "force_company",
+            "active_test",
+            *(keys or ()),
+        )
 
     def _job_prepare_context_before_enqueue(self):
         """Return the context to store in the jobs
         Can be used to keep only safe keys.
         """
+        if self.env.context.get("queue_job_keep_context"):
+            return self.env.context
         return {
             key: value
             for key, value in self.env.context.items()
--- a/queue_job/readme/USAGE.md
+++ b/queue_job/readme/USAGE.md
@@ -288,13 +288,24 @@ The default allow-list is ("tz", "lang", "allowed_company_ids",
 
 When the `queue_job_context_keys` context variable is set, preserve
 the given keys in addition to those normally specified by
-`_job_prepare_context_before_enqueue_keys`.
+`_job_prepare_context_before_enqueue_keys`. Examples:
+
+```python
+# existing context
+record.with_context(queue_job_context_keys=['a']).with_delay().func()
+# new context
+record.with_context(b=1, queue_job_context_keys=['b']).with_delay().func()
+```
 
 When the `queue_job_keep_context` context variable is set, always
 preserve the entire context. This honors the principle of least
 surprise, in that a developer can easily convert a record.method()
 call to record.with_delay().method() with the expectation that it will
-actually execute the same, simply at a later time.
+actually execute the same, simply at a later time. Example:
+
+```python
+record.with_context(queue_job_keep_context=True).with_delay().func()
+```
 
 **Bypass jobs on running Odoo**
 
--- a/queue_job/readme/USAGE.md
+++ b/queue_job/readme/USAGE.md
@@ -276,7 +276,7 @@ Based on this configuration, we can tell that:
 - retries 10 to 15 postponed 30 seconds later
 - all subsequent retries postponed 5 minutes later
 
-**Job Context**
+#### Job Context
 
 The context of the recordset of the job, or any recordset passed in
 arguments of a job, is transferred to the job according to an
@@ -284,8 +284,19 @@ allow-list.
 
 The default allow-list is ("tz", "lang", "allowed_company_ids",
 "force_company", "active_test"). It can be customized in
-`Base._job_prepare_context_before_enqueue_keys`. **Bypass jobs on
-running Odoo**
+`Base._job_prepare_context_before_enqueue_keys`.
+
+When the `queue_job_context_keys` context variable is set, preserve
+the given keys in addition to those normally specified by
+`_job_prepare_context_before_enqueue_keys`.
+
+When the `queue_job_keep_context` context variable is set, always
+preserve the entire context. This honors the principle of least
+surprise, in that a developer can easily convert a record.method()
+call to record.with_delay().method() with the expectation that it will
+actually execute the same, simply at a later time.
+
+**Bypass jobs on running Odoo**
 
 When you are developing (ie: connector modules) you might want to bypass
 the queue job and run your code immediately.
--- a/queue_job/tests/test_json_field.py
+++ b/queue_job/tests/test_json_field.py
@@ -32,6 +32,52 @@ class TestJson(common.TransactionCase):
         }
         self.assertEqual(json.loads(value_json), expected)
 
+    def test_encoder_recordset_context_keys(self):
+        demo_user = self.env.ref("base.user_demo")
+        context = {
+            **demo_user.context_get(),
+            "foo": "bar",
+            "baz": "zeb",
+            "queue_job_context_keys": ["foo"],
+        }
+        partner = self.env(user=demo_user, context=context).ref("base.main_partner")
+        value = partner
+        value_json = json.dumps(value, cls=JobEncoder)
+        expected_context = context.copy()
+        expected_context.pop("baz")
+        expected_context.pop("queue_job_context_keys")
+        expected_context.pop("uid")
+        expected = {
+            "uid": demo_user.id,
+            "_type": "odoo_recordset",
+            "model": "res.partner",
+            "ids": [partner.id],
+            "su": False,
+            "context": expected_context,
+        }
+        self.assertEqual(json.loads(value_json), expected)
+
+    def test_encoder_recordset_keep_context(self):
+        demo_user = self.env.ref("base.user_demo")
+        context = {
+            **demo_user.context_get(),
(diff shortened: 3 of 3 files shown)


ryantran-novobi (approved):

hildickethan:
Maybe it would be better as a parameter to the `with_delay()` itself so you can choose to keep the full context on a job by job basis rather than globally, something like using `with_delay(keep_full_context=True)`
The original reasoning for not always preserving context is [here](https://github.com/OCA/queue/pull/432)

amh-mw:
> Maybe it would be better as a parameter to the `with_delay()` ... something like using `with_delay(keep_full_context=True)`

@hildickethan I like this idea in principle, but haven't had enough coffee this morning to see how to pass `keep_full_context` from `Base.with_delay` -> `DelayableRecordset` -> ??? -> `JobSerialized.convert_*` -> `JobEncoder.default`.

dannyadair:
What about things that can't be serialized? That would at least be the asterisk on "entire".

I feel like the moment you call with_delay you generally know what context keys you could be interested in (or intentionally want to ignore!). Preserving the entire* context is all or nothing.
Maybe it would be more flexible to allow passing a dictionary (`with_context=` ?), and the job context gets updated with that dictionary. Then if you want you can just pass the context itself to preserve it as-is.

dannyadair:
>> to see how to pass keep_full_context from Base.with_delay -> DelayableRecordset -> ??? -> JobSerialized.convert_* -> JobEncoder.default.

Hmm looking at the "how" I think
> allow passing a dictionary

a new field that gets passed through is overkill,
I think it would be much simpler if that itself was just in the context :)

Currently:
```
    def _job_prepare_context_before_enqueue(self):
        """Return the context to store in the jobs
        Can be used to keep only safe keys.
        """
        return {
            key: value
            for key, value in self.env.context.items()
            if key in self._job_prepare_context_before_enqueue_keys()
        }
  ```
  
  So this sets the context dictionary for the job from the current context, but only the key that are in the allow list.
  This could be adjusted to look for a context key that "does more".
  Could be a `queue_job_preserve_context` being True, or `queue_job_context` a dictionary that updates that little default one. Maybe
  
  ```
      def _job_prepare_context_before_enqueue(self): 
        """Return the context to store in the jobs
        Can be used to keep only safe keys.
        """
        if self.env.context.get("queue_job_preserve_context"):
            context = self.env.context.copy()
        else:
            context = {
                key: value
                for key, value in self.env.context.items()
                if key in self._job_prepare_context_before_enqueue_keys()
            }

        if "queue_job_context" in self.env.context:
            context.update(self.env.context["queue_job_context"])

        return context
  ```
  
  or maybe more concise:
  ```
    context = (
        self.env.context.copy()
        if self.env.context.get("queue_job_preserve_context")
        else {k: v for k, v in self.env.context.items() if k in self._job_prepare_context_before_enqueue_keys()}
    )

    context.update(self.env.context.get("queue_job_context") or {})

    return context
  ```
  
  then you could just
  ```
  record.with_context(queue_job_preserve_context=True).with_delay().method()
  ```
  or
  ```
  record.with_context(
      queue_job_context=dict(
          mykey=self.env.context["mykey"]
      )
    ).with_delay().method()
  ```
  or, I don't know, if you have lots of jobs just set
  ```
  record = record.with_context(queue_job_preserve_context=True)
  ```
to save repetition.

**Comment by amh-mw** [comment]
> What about things that can't be serialized? That would at least be the asterisk on "entire".

Absolutely. That has been raised before in https://github.com/OCA/queue/pull/432. Personally, I have never encountered a non-serializable context across the few dozen Odoo and OCA modules that I use, so... I didn't code around it.

> I feel like the moment you call with_delay you generally know what context keys you could be interested in (or intentionally want to ignore!). Preserving the entire* context is all or nothing. Maybe it would be more flexible to allow passing a dictionary (`with_context=` ?), and the job context gets updated with that dictionary. Then if you want you can just pass the context itself to preserve it as-is.

This approach isn't a great fit for me personally, as I have a large number of context variables that I set at the top of large import jobs (per https://github.com/OCA/queue/pull/613#issuecomment-1866983992), and my actual `with_delay` calls are nested arbitrarily deep (and sometimes recursively) in my code.

That said, `with_context` mirrors the familiar Odoo Python `self.env[model].with_context(...)` idiom, so walking that out a little bit, couldn't a developer rely on the core Odoo context manipulation mechanism to do something like:

```python
# addition
self.env[model].with_context(a=1, b=2).with_delay(keep_full_context=True).func()
# replacement
self.env[model].with_context({'c': 3, 'd': 4}).with_delay(keep_full_context=True).func()
```

Then the `queue_job` module doesn't gain context manipulation complexity at all and just relies on core Odoo capabilities. My only concern then is how to play nicely with `_job_prepare_context_before_enqueue` and its default keys ("tz", "lang", "allowed_company_ids", "force_company", "active_test")

amh-mw:
> a new field that gets passed through is overkill, I think it would be much simpler if that itself was just in the context :)

Great suggestion. Give me a bit and let me see what I can work up.

dannyadair:
>> to see how to pass keep_full_context from Base.with_delay -> DelayableRecordset -> ??? -> JobSerialized.convert_* -> JobEncoder.default.

Hmm looking at the "how" I think
> allow passing a dictionary

a new field that gets passed through is overkill,
I think it would be much simpler if that itself was just in the context :)

Currently:
```
    def _job_prepare_context_before_enqueue(self):
        """Return the context to store in the jobs
        Can be used to keep only safe keys.
        """
        return {
            key: value
            for key, value in self.env.context.items()
            if key in self._job_prepare_context_before_enqueue_keys()
        }
  ```
  
  So this sets the context dictionary for the job from the current context, but only the key that are in the allow list.
  This could be adjusted to look for a context key that "does more".
  Could be `queue_job_context` dictionary that updates that little default one. Maybe
  
  ```
      def _job_prepare_context_before_enqueue(self): 
        """Return the context to store in the jobs
        Can be used to keep only safe keys.
        """
        context = {
            key: value
            for key, value in self.env.context.items()
            if key in self._job_prepare_context_before_enqueue_keys()
        }

        if "queue_job_context" in self.env.context:
            context.update(self.env.context["queue_job_context"])

        return context
  ```
    
  then you could just
  ```
  record.with_context(
      queue_job_context=dict(mykey=self.env.context["mykey"])
    ).with_delay().method()
  ```
  or for the full context
  ```
record.with_context(queue_job_context=self.env.context.copy()).with_delay().method()
  ```
  
  
  (or set it on `record = record.with_context()` once to do that for multiple jobs)

**Comment by dannyadair** [comment]
> > a new field that gets passed through is overkill, I think it would be much simpler if that itself was just in the context :)
> 
> Great suggestion. Give me a bit and let me see what I can work up.

Sorry somehow I posted my comment while editing. Please see updated version, I don't think "preserve entire context" needs its own setting - just pass some or copy all.

**Comment by dannyadair** [comment]
```
if "queue_job_context" in self.env.context:
          context.update(self.env.context["queue_job_context"])
```

maybe better
```
context.update(self.env.context.get("queue_job_context") or {})
```

so you can explicitly set it to something falsy to disable it if it might have been set before (otherwise you would have to make it an empty dictionary, this way None works just as well)

**Comment by amh-mw** [comment]
> I don't think "preserve entire context" needs its own setting - just pass some or copy all.

It's a requirement for me. I have far too many context variables to copy paste them across `with_delay` invocations spread across modules. This becomes even worse when considering that I run different combinations of modules for different customers. I don't have the context serialization problem and want the **entire** context every time.

**Comment by dannyadair** [comment]
> > I don't think "preserve entire context" needs its own setting - just pass some or copy all.
> 
> It's a requirement for me. I have far too many context variables to copy paste them across `with_delay` invocations spread across modules. This becomes even worse when considering that I run different combinations of modules for different customers. I don't have the context serialization problem and want the **entire** context every time.

Gotcha. If there's a performance issue an assignment would be better than a .copy().
How about:

```
    def _job_prepare_context_before_enqueue(self): 
      """Return the context to store in the jobs
      Can be used to keep only safe keys.
      """
      if self.env.context.get("queue_job_preserve_context"):
          context = self.env.context
      else:
          context = {
              key: value
              for key, value in self.env.context.items()
              if key in self._job_prepare_context_before_enqueue_keys()
          }

      context.update(self.env.context.get("queue_job_context") or {})

      return context
```

Like this?

> the entire context every time

This way would still require inserting something like `record = record.with_context(queue_job_preserve_context)`
but it would be more explicit/conscious than a global database setting. Other people's modules creating queue jobs may see unwanted side effects.

**Comment by dannyadair** [comment]
>> I have far too many context variables to copy paste them across with_delay invocations spread across modules

> If there's a performance issue an assignment would be better than a .copy().

Sorry maybe I misunderstood. You won't have to copy individual context variables if you set
`.with_context(queue_job_context=self.env.context.copy()`
If those are just flags and other small objects it should be fine but if your context is large an(other) assignment may affect performance (the job already has to serialize everything).

**Comment by dannyadair** [comment]
>        keys = self.env.context.get("queue_job_context_keys")

oh good - I wasn't sure if that was possible.
We've had some weird side effects depending on where context gets set.
I really have to look more into @api.model I keep thinking @classmethod

**Comment by amh-mw** [comment]
I chose to remain out of the context modification game, so adding and using a new context value looks like
```python
self.with_context(a=1, queue_job_context_keys=['a']).with_delay().func()
```

I dislike the repetition, but I think it's preferable to reimplementing `with_context` for `queue_job`.

dannyadair:
> I chose to remain out of the context modification game, so adding and using a new context value looks like
> 
> ```python
> self.with_context(a=1, queue_job_context_keys=['a']).with_delay().func()
> ```
> 
> I dislike the repetition, but I think it's preferable to reimplementing `with_context` for `queue_job`.

I think that's good to differentiate and be explicit. The context is often set somewhere else.
I would expect queue_job_context_keys to only be set right before with_delay() to bring in what's needed.

amh-mw:
Added https://github.com/OCA/queue/pull/936 to more generally increase durability of context encoding.

Issue #770, opened 2025-05-19
Server action methods are no longer executed by crons


JulienMangono:
## Module
queue_job_cron

## Describe the bug
Crons no longer execute server action methods.

## To Reproduce
**Affected versions**:
18.0 

1. Add error raise to the _run_job method 
2. Wait for crons to execute
3. The exception :
 ```
 _run_job
    cron._callback(job['cron_name'], job['ir_actions_server_id'])
TypeError: IrCron._callback() missing 1 required positional argument: 'job_id'

```

**Expected behavior**
No raise and the server action must execute correctly

**Additional context**
The signature of the `_callback` method has changed in 18.0. 
There are now only 2 parameters. In this module the signature has 3 parameters.

PR #771, opened 2025-05-19
[18.0][FIX] queue_job_cron: _callback method signature has too many parameters 

JulienMangono:
PR to fix this [issue](https://github.com/OCA/queue/issues/770) 

The signature of `_callback`method has changed in 18.0. There is no more `job_id`parameter.


florentx (approved):
LG

sbidoul (approved):

PR #773, opened 2025-06-02
[14.0][IMP] queue_job: cancel job before it retries

florentx:
Use case:
- a job hangs due to some concurrent access in the database, it prevents other job from running
- after some time it fails with retryable error
- it is retried short time after, and it can continue like this for some time
- since the button "Cancel" is not authorized on running jobs, the operator cannot cancel this job

This patch allows to Cancel running jobs:
- job is started, then operator chooses to cancel it (because it already retried 2 times, for example, and is blocking other jobs)
- when button is pressed, job record is set to "Cancelled"
- job will not be interrupted
- if job ends successfully: its state is saved to "Done"
- if it fails with an error, its state stays "Cancelled" and it is not retried

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -591,6 +591,17 @@ class Job:
 
         db_record = self.db_record()
         if db_record:
+            # If job was cancelled manually, do not revive it for a retry
+            if (
+                self.retry > 0
+                and db_record.state == CANCELLED
+                and self.state in (PENDING, FAILED)
+            ):
+                self.state = CANCELLED
+                self.date_cancelled = db_record.date_cancelled
+                if not self.result and db_record.result:
+                    self.result = db_record.result
+
             db_record.with_context(_job_edit_sentinel=edit_sentinel).write(
                 self._store_values()
             )
--- a/queue_job/views/queue_job_views.xml
+++ b/queue_job/views/queue_job_views.xml
@@ -25,7 +25,7 @@
                     />
                     <button
                         name="button_cancelled"
-                        states="pending,enqueued,failed"
+                        states="pending,enqueued,started,failed"
                         class="oe_highlight"
                         string="Cancel job"
                         type="object"
--- a/queue_job/wizards/queue_jobs_to_cancelled.py
+++ b/queue_job/wizards/queue_jobs_to_cancelled.py
@@ -11,7 +11,7 @@ class SetJobsToCancelled(models.TransientModel):
 
     def set_cancelled(self):
         jobs = self.job_ids.filtered(
-            lambda x: x.state in ("pending", "failed", "enqueued")
+            lambda x: x.state in ("pending", "failed", "enqueued", "started")
         )
         jobs.button_cancelled()
         return {"type": "ir.actions.act_window_close"}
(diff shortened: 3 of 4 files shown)


guewen (approved):
Makes sense, thanks!

twalter-c2c (approved):

sbidoul (changes requested):
I'm not convinced we should do this.

From a UX perspective this is not cancelling the job, it merely prevents it to be retried. Could we have something doing that explicitly? Such as limiting the retry count?

As a side note, I still think the concept of canceling job is unclear. In my mind, a job is conceptually part of the transaction that created it and we expect it to be eventually completed. That is why there was also a Set Done button initially.

Also, last time I checked it did not work with job graphs.

florentx:
I understand that it is not perfect solution, however it helps in real situations.

Using action Cancel is a last resort solution when there's some Production issue. It is not so important that the job will finish running (trying) before it is really cancelled.

Moreover, one of the case which is tackled here does not comply with the `max_retries` parameter. So an alternative solution will not work.
These errors are PostgreSQL concurrency errors, like `SERIALIZATION_FAILURE`. They are retried indefinitely.

In my opinion this button Cancel should be convenient for the system admin which needs some tool to unblock Production.

Then when system load goes down, then Administrator can find easily the "cancelled" jobs and use button "Requeue" in order to process them, one at a time.
With button "set to done", it is not so easy to identify which ones should be requeued later (or dealt with in a different manner).

sbidoul:
But if I understand correctly, in your use case, you could cancel a running job, and it could still end up in done state? Is that not counter intuitive?

> These errors are PostgreSQL concurrency errors, like SERIALIZATION_FAILURE. They are retried indefinitely.

I did not know these were retried indefinitely. Is that intended? Shouldn't we address that instead?

florentx:
> But if I understand correctly, in your use case, you could cancel a running job, and it could still end up in done state? Is that not counter intuitive?

Yes, if the job is just started, it can finish successfully, even if operator presses button to cancel.
This was also the case previously, because  Job state was not re-checked before setting "Cancelled".
In my opinion, this behavior is acceptable, since button Cancel is for abnormal situations only.

> I did not know these were retried indefinitely. Is that intended? Shouldn't we address that instead?

Regarding the infinite retry, code is here:
https://github.com/OCA/queue/blob/18.0/queue_job/controllers/main.py#L115-L122

Maybe it's legit to keep retrying, if the database load decreases later, and it can terminate successfully. I don't know.

amh-mw:
> > I did not know these were retried indefinitely. Is that intended? Shouldn't we address that instead?
> 
> Regarding the infinite retry, code is here: https://github.com/OCA/queue/blob/18.0/queue_job/controllers/main.py#L115-L122
> 
> Maybe it's legit to keep retrying, if the database load decreases later, and it can terminate successfully. I don't know.

TIL. I have been catching SerializationFailure and rethrowing RetryableJobError myself for years! I have never seen a job actually make it all the way to "infinite" retries. Even my most congested jobs running in tens of thousands with a couple dozen workers in parallel still resolve after my `retry_pattern` spaces them out.

Alternate perspective: Would it make more sense to have the job run out of retries and fail (with some sort of notification), then an admin could restart the job later, rather than have to recover it in real time?

PR #775, opened 2025-06-02
[18.0][REF] queue_job: remove deprecated and not used methods

florentx:
These parts were marked as deprecated since Odoo 12.0.

Also the comparison methods on `Job` are misleading:
- the `sorting_key` returns a tuple containing a missing attribute `seq` --> `AttributeError`
- so it means that this comparison never worked
- and obviously, it is not used anywhere, it is just confusing
- meaningful ordering happens on `odoo.addons.queue_job.jobrunner.channels.ChannelJob`

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -9,7 +9,6 @@ import sys
 import uuid
 import weakref
 from datetime import datetime, timedelta
-from functools import total_ordering
 from random import randint
 
 import odoo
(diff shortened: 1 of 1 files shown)


guewen (approved):
Thanks

amh-mw (approved):
Code reviewed (improved code coverage on negative lines of code 😁), functionally tested (ran several hundred varied data import queue jobs) and organically tested (fooled around with sorting in user interface).

sbidoul (approved):

PR #777, opened 2025-06-02
[18.0][IMP] queue_job: prevent invalid change of job status

florentx:
Currently the check of `Job` state before manual action is loose:
- for button: check happens when the Form view is loaded. Later when operator presses button, there's no guarantee that job is the same status. It could be processed in the meantime.
- for mass-change using wizard, the check exists only on 1 wizard.

Here the check is made at the level of the action, and applies for both scenarios: button or wizard

### DIFF
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -17,6 +17,7 @@ from ..fields import JobSerialized
 from ..job import (
     CANCELLED,
     DONE,
+    ENQUEUED,
     FAILED,
     PENDING,
     STARTED,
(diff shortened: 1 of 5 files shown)


queue_job/models/queue_job.py
```diff
@@ -324,19 +325,24 @@ def _change_job_state(self, state, result=None):
             else:
                 raise ValueError(f"State not supported: {state}")
 
-    def button_done(self):
+    def button_done(self, __state_in=(WAIT_DEPENDENCIES, PENDING, ENQUEUED, FAILED)):
```
sbidoul: Why making `__state_in` a method argument?

queue_job/models/queue_job.py
```diff
@@ -324,19 +325,24 @@ def _change_job_state(self, state, result=None):
...
         result = _("Cancelled by {}").format(self.env.user.name)
-        self._change_job_state(CANCELLED, result=result)
+        records = self.filtered(lambda job_: job_.state in __state_in)
+        records._change_job_state(CANCELLED, result=result)
         return True
 
-    def requeue(self):
-        jobs_to_requeue = self.filtered(lambda job_: job_.state != WAIT_DEPENDENCIES)
-        jobs_to_requeue._change_job_state(PENDING)
+    def requeue(self, __state_in=(FAILED, DONE, CANCELLED)):
```
sbidoul: Why removing WAIT_DEPENDENCIES?

sbidoul (commented):
Good improvement. But I'm uncomfortable with the `__state_in` method parameter and it seems there is a change in requeue related to WAIT_DEPENDENCIES?

queue_job/models/queue_job.py
florentx: We simply reversed the condition: `!=` --> `in` and we excluded the other states which do not make sense here: `PENDING / ENQUEUED / STARTED`

florentx: This dunder argument is a way to have constant defined once, when module is loaded. It will not be rebuilt on each method call.
Also when reading code it is easy to identify which state transitions are allowed. And prefix with dunder `__` makes it explicit that it is internal stuff.

If it is confusing, it can be moved to the 1st line of the method.

queue_job/models/queue_job.py
florentx: This dunder argument is a way to have constant defined once, when module is loaded. It will not be rebuilt on each method call.
Also when reading code it is easy to identify which state transitions are allowed. And prefix with dunder `__` makes it explicit that it is internal stuff.

If it is confusing, it can be moved to the 1st line of the method.

sbidoul: > it can be moved to the 1st line of the method.

I would prefer that indeed.

guewen (approved):

simahawk (approved):

hoangtrann (approved):

sbidoul:
/ocabot merge patch

Thanks!

PR #779, opened 2025-06-03
[18.0][IMP] queue_job: more efficient ChannelJob sorting

florentx:
*Experimental*

- Store the sorting key on the `ChannelJob` instance, to sort more efficiently.
- Simplifiy `__lt__` implementation.

### DIFF
--- a/queue_job/jobrunner/channels.py
+++ b/queue_job/jobrunner/channels.py
@@ -2,6 +2,7 @@
 # Copyright 2015-2016 Camptocamp SA
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
 import logging
+from collections import namedtuple
 from functools import total_ordering
 from heapq import heappop, heappush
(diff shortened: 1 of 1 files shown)


guewen (approved):

florentx:
Coverage lowered because Doctests are not run:
- #780

queue_job/jobrunner/channels.py
```diff
@@ -193,18 +192,36 @@ def __eq__(self, other):
...
+    def date_created(self):
+        return self._sorting_key[2]
+
+    @property
+    def priority(self):
+        return self._sorting_key[1]
+
+    @property
+    def eta(self):
+        return self._sorting_key[0]
```
sbidoul:
```suggestion
        return self._sorting_key.eta
```

Would a namedtuple work for what you want to achieve? It would help readability.

sbidoul (approved):

PR #785, opened 2025-06-06
[17.0] [FIX] queue_job: cancel job waiting dependencies

ivantodorovich:
We should be able to cancel jobs waiting dependencies

### DIFF
--- a/queue_job/views/queue_job_views.xml
+++ b/queue_job/views/queue_job_views.xml
@@ -25,7 +25,7 @@
                     />
                     <button
                         name="button_cancelled"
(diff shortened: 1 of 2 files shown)


florentx:
I assume that you need to change it here as well?
https://github.com/OCA/queue/blob/17.0/queue_job/views/queue_job_views.xml#L28

SilvioC2C (approved):
LGTM, imho @florentx is right about updating the view too

ivantodorovich:
Absolutely. It's done

hildickethan (approved):

hoangtrann (approved):

PR #786, opened 2025-06-06
[17.0] [IMP] queue_job: add_depends on RetryableJobError

ivantodorovich:
This commit adds the posibility to add new dependencies to a job when raising a RetryableJobError.

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -576,6 +576,11 @@ class Job:
         try:
             self.result = self.func(*tuple(self.args), **self.kwargs)
         except RetryableJobError as err:
+            if err.add_depends:
+                from .delay import DelayableGraph
+
+                DelayableGraph._ensure_same_graph_uuid([self] + err.add_depends)
+                self.add_depends(err.add_depends)
             if err.ignore_retry:
                 self.retry -= 1
                 raise
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -76,11 +76,16 @@ class RunJobController(http.Controller):
         http.request.session.db = db
         env = http.request.env(user=SUPERUSER_ID)
 
-        def retry_postpone(job, message, seconds=None):
+        def retry_postpone(job, message, seconds=None, add_depends=None):
             job.env.clear()
             with registry(job.env.cr.dbname).cursor() as new_cr:
-                job.env = api.Environment(new_cr, SUPERUSER_ID, {})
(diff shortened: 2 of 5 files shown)


queue_job/job.py
```diff
@@ -576,6 +576,11 @@ def perform(self):
         try:
             self.result = self.func(*tuple(self.args), **self.kwargs)
         except RetryableJobError as err:
+            if err.add_depends:
+                from .delay import DelayableGraph
```
SilvioC2C, line 580: Curiosity: why do we import this here instead of the file's top? Any specific technical reason?

SilvioC2C (approved):
LGTM, thanks!

queue_job/job.py
ivantodorovich: To avoid a circular dependency

sbidoul:
Hi! What's the use case for this? Could this be done without changing the queue_job API?

ivantodorovich:
> Hi! What's the use case for this? Could this be done without changing the queue_job API?

Hey @sbidoul 👋🏻 

The case we're implementing this for is a connector, where jobs handling the import of records identify missing record dependencies that must be imported before, and so the jobs to import such dependencies are scheduled and the current job is retried after they have succeeded

```py
def import_sale_order(self):
    ...
    missing_products = ...
    if missing_products:
        raise RetryableJobError(
            "Missing products",
            add_depends=[
                self.with_delay().import_product(missing_product)
                for missing_product in missing_products
            ],
        )
```

I thought it'd be an useful addition to the API

> Could this be done without changing the queue_job API?

The difficult part would be that the current transaction is rolledback, and the executing `Job` instance -not accessible from outside- will be stored using a separate cursor, from methods that are not easy to override. I believe @SilvioC2C tried to achieve it with https://github.com/OCA/queue/pull/758, but finally we chose another path at that time

guewen:
Hi @ivantodorovich , creating a new job that inserts itself in a new dependency of the current job and then retry the later which such becomes in a wait_dependencies state seems complicated and I'm afraid of all the corner cases we might encounter (job already in a graph, ...). May I suggest to handle this in the job itself ? 

The code of the job method may enqueue the prerequisite job, chained with a new version of itself (same job method, same parameters), and the current job terminates, so is "done". This solution also doesn't need to handle another transaction.

guewen:
Admittedly if the job is already in a graph it may not do what we want neither

PR #797, opened 2025-06-18
[14.0][FIX] queue_job: Cannot toggle visibility of ETA column

florentx:
Since there are 2 "eta" column in the view, the Optional checkbox does not behave correctly.
Easier way to fix it is to remove "optional" on the 1st column. This is partial revert of recent PR:
- #791 

No such issue in 16/17/18 since there is only 1 `"eta"` column for these versions.


guewen (approved):
/ocabot merge minor

Issue #331, opened 2021-04-06
License Clarification Required v13 onwards 


chrisv1351:
queue_job is LGPL-3 where in other modules like queue_job_cron, base_import_async, base_export_async are still AGPL-3. Is it intentionally kept as AGPL-3 or Forgot to change ? 

I have checked in v13 repo and __manifest__.py file for each of the above mentioned modules.

pedrobaeza:
Why do you want such secondary modules to be `LGPL`? The main one has been done LGPL, as is a base module, and for being tolerant with private extensions, but the other are still AGPL on purpose, as they are not intended to serve as base for other things, and if so, they should be AGPL and released as well. If you want an specific one of them to be changed its license, you will have to ask their authors about that.

chrisv1351:
from my experience, it is better to have a harmonized licenses at branch or at repo level in order to avoid any conflicts or confusion to developer.

pedrobaeza:
That's not needed and there's no conflict. Each developer chooses the license they want for their modules while being compatible, and this scheme is compatible.

Closing as this is not an issue, but a question, and it has been solved.

PR #800, opened 2025-06-26
Make queue_job_cron LGPL

baimont:


amh-mw (changes requested):
You can't relicense a project without the agreement of every single contributor. Sampling other OCA repositories, it looks like AGPL is standard across all of them, so this pull request is likely DOA.

pedrobaeza:
Although I'm not a contributor of this module, my take on this: https://github.com/OCA/contract/pull/1259#issuecomment-3008159581

ValentinVinagre (changes requested):
I do not agree with the license change

etobella (changes requested):
I think it was made clear in previous issues....

https://github.com/OCA/queue/issues/331

For me it should be AGPL.

rousseldenis:
> You can't relicense a project without the agreement of every single contributor. Sampling other OCA repositories, it looks like AGPL is standard across all of them, so this pull request is likely DOA.

@amh-mw Acsone is the original author of the module and in particular @Cedric-Pigeon 

As for queue_job, this is a technical module and serve as base for others.

@etobella @ValentinVinagre  Please remove your blocking as you are not a contributor of this module. Many thanks

etobella:
Have you asked for the people doing the last 3 migrations for example?

@QuocDuong1306
@nguyenminhchien
@JoanMForgeFlow

There are other contributors, but just to add some of them.

If there is consensus on this topic from all contributors, I will unblock it, obviously :wink:

ValentinVinagre (commented):
Removed RC. But I do not agree with the license change

nguyenminhchien:
This PR should be closed as requested by [lmignon](https://github.com/lmignon): https://github.com/OCA/contract/pull/1259#issuecomment-3015155659

rousseldenis:
> This PR should be closed as requested by [lmignon](https://github.com/lmignon): [OCA/contract#1259 (comment)](https://github.com/OCA/contract/pull/1259#issuecomment-3015155659)

This is not the same case.

nguyenminhchien:
> > This PR should be closed as requested by [lmignon](https://github.com/lmignon): [OCA/contract#1259 (comment)](https://github.com/OCA/contract/pull/1259#issuecomment-3015155659)
> 
> This is not the same case.

image

I think they are the same subject, no?

ValentinVinagre:
@rousseldenis 

> > You can't relicense a project without the agreement of every single contributor. Sampling other OCA repositories, it looks like AGPL is standard across all of them, so this pull request is likely DOA.
> 
> @amh-mw Acsone is the original author of the module and in particular @Cedric-Pigeon
> 
> As for queue_job, this is a technical module and serve as base for others.
> 
> @etobella @ValentinVinagre Please remove your blocking as you are not a contributor of this module. Many thanks

I can't remove the block because I only have read permissions. Can you cancel my review?

rousseldenis:
> I can't remove the block because I only have read permissions. Can you cancel my review?

No, the fact you react here is sufficent :smiley:

rousseldenis:
> I think they are the same subject, no?

I checked with @lmignon. It was just about contract :sweat_smile:

rousseldenis:
@etobella 
@QuocDuong1306 and @JoanMForgeFlow are no more active. I think we have the full pack :smile:

yvaucher:
I just want to react to the claim that you cannot re-license without the approval of all contributors, it's not true due to the CLA that all contributor must have signed. It can be decided by the OCA. So it's a collective decision one person cannot put a veto. (It's nice to ask though and be clear about why)

I'm also totally in favor of keeping the default as AGPL accross OCA repo.

Here is my take, `queue_job` modules are LGPL and cron is a feature of the core. It would make sense to also have it as LGPL. I'm not against in this case. However, I don't see the need for it.

Let's assume the following dependency graph:

```mermaid
flowchart LR
    queue_job_cron["queue_job_cron"] --> queue_job["queue_job"]
    specific_addon["specific_addon"] --> queue_job

    style queue_job_cron fill:#C8E6C9,stroke:#00C853,color:#000000
    style queue_job stroke:#2962FF,fill:#BBDEFB,color:#000000
    style specific_addon fill:#E1BEE7,stroke:#AA00FF,color:#000000
```

The `specific_addon`, don't inherit on `queue_job_cron`, thus you can still configure your cron as a job even if it launches a proprietary method. Basically the only thing that you cannot do is to put the field `run_as_queue_job` in xml data inside your `specific_addon`.

EDIT: fixed field name

etobella:
@yvaucher It is true that CLA allows OCA to change the license, but this has never been used in the past. Also, it is not specified how this change of license can be launched according to this document or the bylaws. 

Before applying it, we should define the way that this specific change can be applied.

For this reason, we always ask for approval of all contributors to avoid this problem.

yvaucher:
@etobella You are right I was just nuancing @amh-mw assertion. Mostly because there is on the balance contributors that might never answer or you might have 1% of contributor that veto the change. For the contributions that got no approval it should be removed legally speaking, that's why invoking the CLA could be an option.

I also want to remind that the OCA is commited to keep a FOSS license. And the license could for instance be changed (like the change from GPL to AGPL to close a loophole).


To me the current issue is just a followup of the license change on connector and queue_job which is sad but already happened in 2019.

etobella:
Well, also changing licenses should require a complete audit of CLA signed, because we had a problem in another repo for this reason.

In any case, I would prefer to keep as many AGPL modules as possible too.

rousseldenis:
@etobella @yvaucher @baimont Don't miss to read OCA FAQ about that:

https://odoo-community.org/resources/faq

If I read correctly, contributors that are not authors are not involved in licensing change.

etobella:
Well, Certainly, it says only authors, but it states all, nothing about disappeared authors, so you should need the author (or their company at that time approval) approval from disappeared authors too (also, it is not defined what a disappeared author is....). Also, it is explained that if we want to change without all author approvals, we need to make a delegate vote.

# Can I change the license when I port a module to a new version?

No. You can only change the license of a module if you have the copyright or if all authors of the module agree to change the license.

# Can the OCA change the license of a module?

Through the [signed CLA](https://odoo-community.org/about/cla), the OCA has the right to relicense a module under any OSI certified license. But the CLA is here to protects the contributors, not to relicense their work. If that would happen, it’ll be subject to the vote of the Delegate assembly (the board alone cannot decide).


Thanks for sharing @rousseldenis

PR #812, opened 2025-08-11
Increase ideal timeout value to 5 seconds.

andevatikul:
It is good to increase ideal timeout value to 5 seconds to avoid any network latency related error/request failure.

### DIFF
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -245,7 +245,8 @@ def _async_http_get(scheme, host, port, user, password, db_name, job_uuid):
                 auth = (user, password)
             # we are not interested in the result, so we set a short timeout
             # but not too short so we trap and log hard configuration errors
-            response = requests.get(url, timeout=1, auth=auth)
+            # Set timeout to 5 seconds to avoid any network latency related error/request failure.
+            response = requests.get(url, timeout=5, auth=auth)
 
             # raise_for_status will result in either nothing, a Client Error
             # for HTTP Response codes between 400 and 500 or a Server Error


amh-mw (changes requested):
> It is good to increase ideal timeout value to 5 seconds to avoid any network latency related error/request failure.

Network latency? Isn't this request local to the current Odoo server? Or at least the local container cluster?

```python
# we are not interested in the result, so we set a short timeout
# but not too short so we trap and log hard configuration errors
```
Why wait longer for a response that is not desired?

andevatikul:
> > It is good to increase ideal timeout value to 5 seconds to avoid any network latency related error/request failure.
> 
> Network latency? Isn't this request local to the current Odoo server? Or at least the local container cluster?
> 
> ```python
> # we are not interested in the result, so we set a short timeout
> # but not too short so we trap and log hard configuration errors
> ```
> 
> Why wait longer for a response that is not desired?

**Firstly,** Yes, I am talking about local requests. Think about too much external requests that jammed or slow down the Server or Odoo instance or HTTP server processes; or, think about a scenario of Brute-force or DDoS and the request didn't get response on-time and exceeded 2-3 seconds. In these scenarios, even internal/local requests might fail.

**Secondly,** in normal system operation, if no bottlenecks, of course the requests/response will be quick even less than a sec although you increase to 5 seconds. A bit timeout increase is a safe-side which will ensure quicker response in normal case as well as will ensure safety in high-load/stress scenario. It doesn't matter if the request is internal or external.

One problem I have faced is, due to some internal Python + Odoo process + System process's misbehave, a lot of repeated, failed jobs were re-queued, timed-out due to unresponsiveness.

sbidoul:
I suspect your problem is due to an overload on the Odoo instances that process jobs with no workers available to accept new jobs. Please check that the root channel capacity is lower than the number of workers and that it leaves enough workers available for the interactive workload.

Until there is more evidence, I'm -1 on changing this timeout.

sbidoul:
Note that with #716 and #853, the requeuing occurs after 10 sec in case the server has no available worker. #859 also improves how this specific circumstance is handled. So closing this one, as this specific timeout  does not trigger a requeue anymore.

PR #813, opened 2025-08-12
[18.0][IMP] queue_job: avoid deprecation warning about `datetime.datetime.utcnow()` in `_odoo_now()`

acsonefho:
`datetime.datetime.utcnow()` is now deprecated and should be replaced by `datetime.datetime.now()` (optional TZ parameter).
As the original `_odoo_now()` doesn't contain the Timezone, the parameter `datetime.UTC` is not added into this improvement to stay a `naive datetime`.

The purpose of `_odoo_now()` is to have the current datetime in UTC timezone but staying `naive datetime`.

https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

https://docs.python.org/3/whatsnew/3.12.html#:~:text=datetime%3A%20datetime,gh%2D103857.)

**Original warning:**
image

### DIFF
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -139,7 +139,6 @@ Caveat
        of running Odoo is obviously not for production purposes.
 """
 
-import datetime
 import logging
 import os
 import selectors
@@ -181,15 +180,10 @@ def _channels():
     )
 
 
-def _datetime_to_epoch(dt):
+def _odoo_now():
     # important: this must return the same as postgresql
     # EXTRACT(EPOCH FROM TIMESTAMP dt)
-    return (dt - datetime.datetime(1970, 1, 1)).total_seconds()
-
-
-def _odoo_now():
-    dt = datetime.datetime.utcnow()
-    return _datetime_to_epoch(dt)
+    return time.time()
 
 
 def _connection_info_for(db_name):


amh-mw (changes requested):
Given https://www.odoo.com/documentation/18.0/administration/on_premise/source.html#python
> Odoo [18] requires Python 3.10 or later to run.

I went through the Python documentation for versions 3.10 through 3.13 and the guidance on replacing utcnow is consistent:

> Warning: Because naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing the current time in UTC is by calling datetime.now(timezone.utc).

So shouldn't the change be 
```diff
-    dt = datetime.datetime.utcnow()
+    dt = datetime.datetime.now(timezone.utc)
```

acsonefho:
> Given https://www.odoo.com/documentation/18.0/administration/on_premise/source.html#python
> 
> > Odoo [18] requires Python 3.10 or later to run.
> 
> I went through the Python documentation for versions 3.10 through 3.13 and the guidance on replacing utcnow is consistent:
> 
> > Warning: Because naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing the current time in UTC is by calling datetime.now(timezone.utc).
> 
> So shouldn't the change be
> 
> ```diff
> -    dt = datetime.datetime.utcnow()
> +    dt = datetime.datetime.now(timezone.utc)
> ```

Yes but if I do that, the returned `datetime` contains a timezone (and others errors occurs) due to difference between `datetime` and `naive datetime`.

sbidoul:
> Yes but if I do that, the returned datetime contains a timezone (and others errors occurs) due to difference between datetime and naive datetime.

Yet `dt = datetime.datetime.now()` is not correct. This still needs work.

acsonefho:
What do you think of this: `datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None)`

sbidoul:
Yes, that sounds better.

amh-mw (approved):
LGTM

sbidoul (approved):

benwillig (approved):
LGTM

sbidoul:
I checked that `_odoo_now()` does the same as `SELECT EXTRACT(EPOCH FROM NOW());` in psql.

That said, so does `time.time()`, so this can probably be simplified.

acsonefho:
> I checked that `_odoo_now()` does the same as `SELECT EXTRACT(EPOCH FROM NOW());` in psql.
> 
> That said, so does `time.time()`, so this can probably be simplified.

The current return type of `_odoo_now()` is `datetime` and `time.time()` is `float`.
That update probably needs others changes.

rousseldenis (approved):

FrancoMaxime (approved):
LGTM: code review

queue_job/jobrunner/runner.py
```diff
@@ -188,7 +188,8 @@ def _datetime_to_epoch(dt):
 
 
 def _odoo_now():
-    dt = datetime.datetime.utcnow()
+    # Need naive datetime
+    dt = datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None)
```
AnizR: I agree with @sbidoul 

We can remove the funciton `_datetime_to_epoch` and directly return:
```suggestion
    return time.time()
```

I tested localy and in fact, `_odoo_now()` returns a `float`:
```
>>> dt = datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None)
>>> (dt - datetime.datetime(1970, 1, 1)).total_seconds()
1757943419.659445
>>> time.time()
1757943426.0511284
>>> 
```

FrancoMaxime: I agree with @AnizR comment

SirPyTech (approved):
Thanks!

queue_job/jobrunner/runner.py
AnizR: > I agree with @sbidoul
> 
> We can remove the funciton `_datetime_to_epoch` and directly return:
> 
> I tested localy and in fact, `_odoo_now()` returns a `float`:
> 
> ```
> >>> dt = datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None)
> >>> (dt - datetime.datetime(1970, 1, 1)).total_seconds()
> 1757943419.659445
> >>> time.time()
> 1757943426.0511284
> >>> 
> ```

As discussed with @acsonefho, I have done the refactor of the function `_odoo_now()`.

queue_job/jobrunner/runner.py
```diff
@@ -181,15 +180,8 @@ def _channels():
...
-def _datetime_to_epoch(dt):
-    # important: this must return the same as postgresql
-    # EXTRACT(EPOCH FROM TIMESTAMP dt)
-    return (dt - datetime.datetime(1970, 1, 1)).total_seconds()
-
-
 def _odoo_now():
-    dt = datetime.datetime.utcnow()
-    return _datetime_to_epoch(dt)
+    return time.time()
```
sbidoul, line 186: I'd keep the comment that says it must do the same as the postgres equivalent, though.

AnizR: Indeed, done :+1:

PR #818, opened 2025-08-28
[FIX] queue_job_cron_jobrunner: use priority to select job

StefanWiselka-Solvti:
[FIX] queue_job_cron_jobrunner: use priority to select job
* use FIFO, firt createad job will be treat first
* if priority are different it take the precedent

Yet we are not using channel priority into account.

Note:  FIX has already been applied to branch 17.0.
Ref: https://github.com/OCA/queue/commit/f8fcceba13d9e78fea82bcca08a648f4b23c6664


ivantodorovich (approved):

MarcinSokolowski-Solvti (approved):
LGTM 🚀

PR #825, opened 2025-09-16
[17.0] [FIX] queue_job: fix exception msg handling

ThomasBinsfeld:
Forward port of #824 

18.0 PR : #826  

Use args[0] instead of name to get the exception message as the name attribute is deprecated

PR #824, opened 2025-09-16
[16.0] [FIX] queue_job: fix exception msg handling

ThomasBinsfeld:
Use args[0] instead of name to get the exception message as the name attribute is deprecated

17.0 forward-port PR : #825 
18.0 forward-port PR : #826

PR #826, opened 2025-09-16
[18.0] [FIX] queue_job: fix exception msg handling

ThomasBinsfeld:
Forward port of #824 (16.0) 

17.0 PR : #825 

'name' attributes for odoo's exception has been deprecated and produces a warning message. This commit fixes the behavior

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -156,7 +156,9 @@ class RunJobController(http.Controller):
         exception_name = orig_exception.__class__.__name__
         if hasattr(orig_exception, "__module__"):
             exception_name = orig_exception.__module__ + "." + exception_name
-        exc_message = getattr(orig_exception, "name", str(orig_exception))
+        exc_message = (
+            orig_exception.args[0] if orig_exception.args else str(orig_exception)
+        )
         return {
             "exc_info": traceback_txt,
             "exc_name": exception_name,
--- a/queue_job/tests/__init__.py
+++ b/queue_job/tests/__init__.py
@@ -1,3 +1,4 @@
+from . import test_run_rob_controller
 from . import test_runner_channels
 from . import test_runner_runner
 from . import test_delayable
--- a/queue_job/tests/test_run_rob_controller.py
+++ b/queue_job/tests/test_run_rob_controller.py
@@ -1,0 +1,17 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
+
+from ..controllers.main import RunJobController
+from ..job import Job
+
+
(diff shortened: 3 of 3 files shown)


FrancoMaxime (approved):
LGTM

AnizR (approved):
LGTM

amh-mw (approved):

amh-mw (changes requested):
Test coverage?

queue_job/controllers/main.py
```diff
@@ -156,7 +156,9 @@ def _get_failure_values(self, job, traceback_txt, orig_exception):
         exception_name = orig_exception.__class__.__name__
         if hasattr(orig_exception, "__module__"):
             exception_name = orig_exception.__module__ + "." + exception_name
-        exc_message = getattr(orig_exception, "name", str(orig_exception))
+        exc_message = (
+            orig_exception.args[0] if orig_exception.args else str(orig_exception)
```
amh-mw, line 160: So I just noticed above
```python
        except (FailedJobError, Exception) as orig_exception:
```
and both of those classes extend Exception, which extends BaseException, which is defined
```python
class BaseException:
    args: tuple[Any, ...]
```
so I withdraw my comment about the args[0] guard. As penance, here is a unit test:
```python
# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)

from odoo.tests import TransactionCase

from ..controllers.main import RunJobController
from ..job import Job


class TestRunJobController(TransactionCase):
    def test_get_failure_values(self):
        method = self.env["res.users"].mapped
        job = Job(method)
        ctrl = RunJobController()
        rslt = ctrl._get_failure_values(job, "info", Exception("zero", "one"))
        self.assertEqual(rslt, {
            "exc_info": "info",
            "exc_name": "Exception",
            "exc_message": "zero"
        })
```

amh-mw (changes requested):

ThomasBinsfeld:
The modified file doesn't seem covered at all, we won't write its coverage in this PR. I suggest to merge as it. ping @guewen

amh-mw:
> The modified file doesn't seem covered at all, we won't write its coverage in this PR. I suggest to merge as it. ping @guewen

Yes, that's why I offered unit test in prior comment.

ThomasBinsfeld:
> > The modified file doesn't seem covered at all, we won't write its coverage in this PR. I suggest to merge as it. ping @guewen
> 
> Yes, that's why I offered unit test in prior comment.

Missed it, sorry. I've added it in the three PR's.

amh-mw (approved):

guewen (approved):

guewen:
/ocabot merge patch
Thanks both for the code and the test :)

PR #829, opened 2025-09-23
[18.0][MIG] queue job web notify

hhgabelgaard:
Std migration


queue_job_web_notify/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     "summary": """
         This module allows to display a notification to the related user of a
         failed job. It uses the web_notify notification feature.""",
-    "version": "13.0.1.0.0",
+    "version": "18.0.1.0.0",
```
amh-mw, line 9: If the migration were based off of 16 instead of 13, it would gain some translation commits:
https://github.com/OCA/queue/commits/16.0/queue_job_web_notify

amh-mw (changes requested):

queue_job_web_notify/__manifest__.py
hhgabelgaard: @amh-mw Good point
Closing this in favor of #830

PR #852, opened 2025-11-20
[18.0][IMP] queue_job: handle same uuid in duplicated db

hoangtrann:
Because jobrunner can be initialized with instance having multiple databases, cloning a db with failed jobs would cause the jobrunner to failed and causing jobs hang.

### DIFF
--- a/queue_job/jobrunner/channels.py
+++ b/queue_job/jobrunner/channels.py
@@ -799,6 +799,62 @@ class ChannelManager:
     >>> cm.notify(db, 'S', 'S3', 3, 0, 10, None, 'done')
     >>> pp(list(cm.get_jobs_to_run(now=105)))
     []
+
+    Test handling of duplicate job UUIDs across different databases.
+    This can happen when a database is duplicated/cloned and both databases
+    notify the same job runner with the same job UUID.
+
+    >>> cm = ChannelManager()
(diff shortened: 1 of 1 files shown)


queue_job/jobrunner/channels.py
```diff
@@ -1029,22 +1085,31 @@ def notify(
...
-                _logger.debug("job %s properties changed, rescheduling it", uuid)
+            # if db_name differs, this is likely a cloned database
+            if job.db_name != db_name:
+                _logger.warning(
+                    "job %s exists in multiple databases (%s and %s). recreating it",
+                    uuid,
+                    job.db_name,
+                    db_name,
+                )
                 self.remove_job(uuid)
```
sbidoul, line 1096: I'm not quite sure this is a correct thing to do. We still end up with two jobs with the same uuid in the queue and that's going to be confusing, if only when reading the logs. I suspect it will also break things down the line.

If that case is important to you, perhaps you could have a little script that updates uuid and dbname in the jobs of the cloned database?

sbidoul (changes requested):
Another problem when duplicating a database is graph_uuid, which is also supposed to be unique. 

So at this point I don't think this is an easy thing to support.

hoangtrann:
that's fair point, let me think about this more

PR #853, opened 2025-11-21
[18.0][IMP] queue_job: requeue orphaned jobs

hoangtrann:
This pr add a secondary attempt to requeue "orphaned jobs" where it stucks in enqueued, not yet make it to queue_job_lock and selecting jobs after db initialization also misses it. Even a server restart would not unblock the job.

This is most likely due to server being shutdown or crash during job enqueue. Manually requeue the job on UI make everything running again.

### DIFF
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -378,6 +378,35 @@ class Database:
             RETURNING uuid
             """
 
+    def _query_requeue_orphaned_jobs(self):
+        """Query to requeue jobs stuck in 'enqueued' state without a lock.
+
+        This handles the edge case where the runner marks a job as 'enqueued'
+        but the HTTP request to start the job never reaches the Odoo server
+        (e.g., due to server shutdown/crash between setting enqueued and
+        the controller receiving the request). These jobs have no lock record
+        because set_started() was never called, so they are invisible to
+        _query_requeue_dead_jobs().
+        """
+        return """
+            UPDATE
+                queue_job
+            SET
+                state='pending'
+            WHERE
+                state = 'enqueued'
+                AND date_enqueued < (now() AT TIME ZONE 'utc' - INTERVAL '10 sec')
+                AND NOT EXISTS (
+                    SELECT
+                        1
+                    FROM
+                        queue_job_lock
+                    WHERE
+                        queue_job_id = queue_job.id
+                )
+            RETURNING uuid
+            """
+
     def requeue_dead_jobs(self):
         """
         Set started and enqueued jobs but not locked to pending
@@ -405,6 +434,14 @@ class Database:
 
             for (uuid,) in cr.fetchall():
                 _logger.warning("Re-queued dead job with uuid: %s", uuid)
+
+            # Requeue orphaned jobs (enqueued but never started, no lock)
+            query = self._query_requeue_orphaned_jobs()
+            cr.execute(query)
+            for (uuid,) in cr.fetchall():
+                _logger.warning(
+                    "Re-queued orphaned job (enqueued without lock) with uuid: %s", uuid
+                )
 
 
 class QueueJobRunner:
--- a/queue_job/jobrunner/runner.py
+++ b/queue_job/jobrunner/runner.py
@@ -357,52 +357,26 @@ class Database:
                         ELSE exc_info
                     END)
             WHERE
-                id in (
-                    SELECT
-                        queue_job_id
-                    FROM
-                        queue_job_lock
-                    WHERE
-                        queue_job_id in (
-                            SELECT
-                                id
-                            FROM
-                                queue_job
-                            WHERE
-                                state IN ('enqueued','started')
-                                AND date_enqueued <
-                                (now() AT TIME ZONE 'utc' - INTERVAL '10 sec')
-                        )
-                    FOR UPDATE SKIP LOCKED
-                )
-            RETURNING uuid
-            """
-
-    def _query_requeue_orphaned_jobs(self):
-        """Query to requeue jobs stuck in 'enqueued' state without a lock.
-
-        This handles the edge case where the runner marks a job as 'enqueued'
-        but the HTTP request to start the job never reaches the Odoo server
-        (e.g., due to server shutdown/crash between setting enqueued and
-        the controller receiving the request). These jobs have no lock record
-        because set_started() was never called, so they are invisible to
-        _query_requeue_dead_jobs().
-        """
-        return """
-            UPDATE
-                queue_job
-            SET
-                state='pending'
-            WHERE
-                state = 'enqueued'
+                state IN ('enqueued','started')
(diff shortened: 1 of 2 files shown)


queue_job/jobrunner/runner.py
```diff
@@ -378,6 +378,44 @@ def _query_requeue_dead_jobs(self):
...
+                                1
+                            FROM
+                                queue_job_lock
+                            WHERE
+                                queue_job_id = queue_job.id
+                        )
+                        AND date_enqueued <
+                            (now() AT TIME ZONE 'utc' - INTERVAL '10 sec')
+                    FOR UPDATE SKIP LOCKED
+                )
```
sbidoul: Would a simpler query like this work too?

```suggestion
            WHERE
                state = 'enqueued'
                AND date_enqueued < (now() AT TIME ZONE 'utc' - INTERVAL '10 sec')
                AND NOT EXISTS (
                    SELECT
                        1
                    FROM
                        queue_job_lock
                    WHERE
                        queue_job_id = queue_job.id
                )
```

sbidoul (commented):
Good catch! I kind of remember we had thought about this case when doing the original implementation with @AnizR but maybe it got lost in the various iterations we did.

queue_job/jobrunner/runner.py
sbidoul: Can you add a little test?

hoangtrann: > Would a simpler query like this work too?

I think that's absolutely works! My original thought is trying to keep it as similar to how we query dead jobs so that it increases the likelihood this can be merged.

hoangtrann: @sbidoul I've updated the query and add a test case for this

AnizR:
> Good catch! I kind of remember we had thought about this case when doing the original implementation with @AnizR but maybe it got lost in the various iterations we did.

In deed, we talked about it but I think that it got lost.

Thanks for this patch, I'll review it within the week.

AnizR (approved):

hoangtrann:
@sbidoul are we good to merge?

sersanchus (approved):

maurochip (approved):

zamberjo (approved):

sbidoul:
Re-reading this I'm slightly annoyed that it reintroduces a complexity that we had simplified before.

I wonder if it would work to add a `OR NOT EXISTS (SELECT 1 FROM queue_job_lock...)` in `_query_requeue_dead_jobs`.

hoangtrann:
@sbidoul I've adapted the query to your suggestions and tests are passing!

queue_job/jobrunner/runner.py
```diff
@@ -357,23 +357,24 @@ def _query_requeue_dead_jobs(self):
...
-                                AND date_enqueued <
-                                (now() AT TIME ZONE 'utc' - INTERVAL '10 sec')
-                        )
-                    FOR UPDATE SKIP LOCKED
+                state IN ('enqueued','started')
+                AND date_enqueued < (now() AT TIME ZONE 'utc' - INTERVAL '10 sec')
+                AND (
+                    id in (
+                        SELECT
+                            queue_job_id
```
```suggestion
                            1
```

for symmetry

queue_job/jobrunner/runner.py
```diff
@@ -357,23 +357,24 @@ def _query_requeue_dead_jobs(self):
...
+                        FROM
+                            queue_job_lock
+                        FOR UPDATE SKIP LOCKED
```
sbidoul: It needs a where clause here otherwise it locks all jobs.

sbidoul (commented):
Thanks! One important fix is needed but I think it looks better like this overall.

I'll need to check the execution plan of the new query to be sure it's efficient.

sbidoul:
Execution plan looks like this (with the missing where clause added), with one job started with no lock held, and one enqueued with no lock record. So looks good to me:

```
 Update on queue_job  (cost=0.43..7.30 rows=1 width=106) (actual time=0.146..0.711 rows=2 loops=1)
   ->  Index Scan using queue_job_state_index on queue_job  (cost=0.43..7.30 rows=1 width=106) (actual time=0.046..0.095 rows=2 loops=1)
         Index Cond: ((state)::text = ANY ('{enqueued,started}'::text[]))
         Filter: ((date_enqueued < ((now() AT TIME ZONE 'utc'::text) - '00:00:10'::interval)) AND ((SubPlan 1) OR (NOT (SubPlan 2))))
         SubPlan 1
           ->  LockRows  (cost=0.43..2.66 rows=1 width=10) (actual time=0.010..0.010 rows=0 loops=2)
                 ->  Index Scan using queue_job_lock_queue_job_id_index on queue_job_lock  (cost=0.43..2.65 rows=1 width=10) (actual time=0.008..0.008 rows=0 loops=2)
                       Index Cond: (queue_job_id = queue_job.id)
         SubPlan 2
           ->  Index Only Scan using queue_job_lock_queue_job_id_index on queue_job_lock queue_job_lock_1  (cost=0.43..2.65 rows=1 width=0) (actual time=0.004..0.004 rows=0 loops=1)
                 Index Cond: (queue_job_id = queue_job.id)
                 Heap Fetches: 0
 Planning Time: 0.247 ms
 Trigger queue_job_notify: time=0.083 calls=2
 Execution Time: 0.855 ms
 ```

**Review by sbidoul — CHANGES_REQUESTED** [review]
Marking as change requested until the missing where clause is added.

File: queue_job/jobrunner/runner.py (hunk 2 above)
Comment (sbidoul) [reply queue_job/jobrunner/runner.py]: Hm, maybe not because we lock.

File: queue_job/jobrunner/runner.py (hunk 3 above)
Comment (hoangtrann) [reply queue_job/jobrunner/runner.py]: @sbidoul I've updated the clause. I was trying to avoid redundant where clause for state and date_enqueued not realizing it introduces lock issue.

**Comment by sbidoul** [comment]
Ah no, there is a misunderstanding. What you did was good, but the missing where clause was `WHERE queue_job_lock.queue_job_id = queue_job.id` in the first `id IN `.

**Comment by sbidoul** [comment]
So like this:


```
            WHERE
                state IN ('enqueued','started')
                AND date_enqueued < (now() AT TIME ZONE 'utc' - INTERVAL '10 sec')
                AND (
                    id in (
                        SELECT
                            queue_job_id
                        FROM
                            queue_job_lock
                        WHERE
                            queue_job_lock.queue_job_id = queue_job.id
                        FOR UPDATE SKIP LOCKED
                    )
                    OR NOT EXISTS (
                        SELECT
                            1
                        FROM
                            queue_job_lock
                        WHERE
                            queue_job_lock.queue_job_id = queue_job.id
                    )
                )
            RETURNING uuid
```

hoangtrann:
@sbidoul I've reverted the changes and add the WHERE clause

sbidoul (approved):

sbidoul:
Thanks! This version reads even better than before.

PR #854, opened 2025-11-22
[18.0][DOC] queue_job: remove dead jobs caveat from roadmap

sbidoul:
The automatic dead jobs requeuer now works out of the box.


flotho (approved):
Hi,
I can confirm in production !

PR #859, opened 2025-12-11
[FIX] handle race conditions that could lead to job running twice

### LINKED ISSUE #858: After requeueing a job, first run often fails to update date_done with concurrent error
## Module
queue_job

## Describe the bug

I see in my log many cases of:

```
# good case
2025-12-09 07:22:43,904 2229881 WARNING ? odoo.addons.queue_job.jobrunner.runner: Re-queued dead job with uuid: 76c33f7c-2a1f-4564-9545-2daccf99026d 
2025-12-09 07:22:43,908 2229881 INFO ? odoo.addons.queue_job.jobrunner.runner: asking Odoo to run job 76c33f7c-2a1f-4564-9545-2daccf99026d on db ....

# bad case
2025-12-09 07:22:43,922 2229881 WARNING ? odoo.addons.queue_job.jobrunner.runner: Re-queued dead job with uuid: b8b82330-6c0d-44a6-b12c-1fde5bf8ee42 
2025-12-09 07:22:43,926 2229881 INFO ? odoo.addons.queue_job.jobrunner.runner: asking Odoo to run job b8b82330-6c0d-44a6-b12c-1fde5bf8ee42 on db ....
2025-12-09 07:22:43,954 2820876 ERROR .... odoo.sql_db: bad query: UPDATE "queue_job" SET "date_done" = '2025-12-09 07:22:43.951733', "dependencies" = '{"depends_on": [], "reverse_depends_on": []}', "exec_time" = 0.034894, "retry" = 1, "state" = 'done' WHERE id IN (22671925)
ERROR: could not serialize access due to concurrent update
```

## To Reproduce
**Affected versions**: 16.0

1. Have a job start
2. Kill the worker to make the job "dead"
3. Job gets enqueued
4. Job starts
5. Concurrent error; second retry goes through

**Expected behavior**
First run directly goes through

**Additional context**
I don't know at the moment what's locking the queue job record - if it is the previous "dead" job (unlikely, because jobrunner itself already wrote to that record), or the jobrunner itself locking the record (more likely, but also strange since the cursor should be committed and closed at that point)

sbidoul:
Do not wait for locks and start jobs that are not in the expected state.

In a nutshell this PR replaces two `SELECT FOR UPDATE` by `SELECT FOR UPDATE SKIP LOCKED`. This is because if the job to run is already locked or not in the expected state there is no need to wait: it means the job is being executed by another worker already.

Also since there is commit between the check that the job is in enqueued state and set started, and the actual start of execution, there was window there for two workers to start the same job in some rare situations. This PR should avoid this case.

Maybe fixes #858

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -16,7 +16,7 @@ from odoo.modules.registry import Registry
 from odoo.service.model import PG_CONCURRENCY_ERRORS_TO_RETRY
 
 from ..delay import chain, group
-from ..exception import FailedJobError, RetryableJobError
+from ..exception import FailedJobError, RetryableJobError, UnexpectedJobStateError
 from ..job import ENQUEUED, Job
 
 _logger = logging.getLogger(__name__)
@@ -29,13 +29,16 @@ DEPENDS_MAX_TRIES_ON_CONCURRENCY_FAILURE = 5
 class RunJobController(http.Controller):
     def _try_perform_job(self, env, job):
         """Try to perform the job."""
+
+        # Here we expect that the job is in 'enqueued' state and protected
+        # from concurrent start by the SELECT FOR UPDATE on the job record.
+        # that was done at the beginning of the runjob method.
         job.set_started()
         job.store()
         env.cr.commit()
+
         job.lock()
-
         _logger.debug("%s started", job)
-
         job.perform()
         # Triggers any stored computed fields before calling 'set_done'
         # so that will be part of the 'exec_time'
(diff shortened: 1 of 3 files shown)


sbidoul:
@thomaspaulb if you have a reproducer, for #858 you may want to test this, possibly in combination with #853

sbidoul:
BTW, for readability, I think the first part of `_try_perform_job` (until `job.lock()`) should be extracted from that function. That would make the flow easier to understand.

sbidoul:
Improved version in #866

PR #860, opened 2025-12-18
[18.0][IMP] queue_job: config param to disable queue, readme cleanup

RLeeOSI:
Adds a config param to disable the queue for testing purposes. In the [queue_job] section of your odoo conf file, include queue_job__no_delay=True to disable the job queue. This is particularly useful for odoo.sh and other environments where setting an env var is not convenient. Follows the same format as the env var QUEUE_JOB__NO_DELAY or the context key queue_job__no_delay .

Also cleans up some documentation. Moves some documentation out of runner.py into CONFIGURE.md where it belongs, and removes some outdated config params.


sbidoul (changes requested):
Hi, thanks for the contrib. Can you do separate PRs for the documentation improvements and the new feature?

RLeeOSI:
@sbidoul I can, but as part of the disable feature, the documentation needs to be updated. I can make a branch that updates the existing documentation, but then I either need to make my feature changes on top of that commit, or update the old documentation with the new feature, then resolve conflicts when both branches get merged.

IMO it's cleaner to keep it in one branch. If you prefer, I can do the generic documentation update as a separate commit, then add the feature and its documentation as a second commit.

sbidoul:
if you can do a PR with the docs update I promise to review it and merge quickly then you can followup with the new feature.

RLeeOSI:
Closing for #886

PR #840, opened 2025-10-01
[19.0] [MIG] queue_job: migrate + tests

tishmen:
Scope
- queue_job, test_queue_job only

Summary
- Migrate to Odoo 19 using the working feature branch code.
- Tests create their own users (no demo data), context handling aligned with 19.0.
- Asset paths are relative under web.assets_backend.

Pre-commit
- Verified locally; no additional changes applied here to keep parity with the feature branch.

Tests
- Command:
  ./odoo/odoo-bin -c odoo.conf -d odoo_queue_pr_queue_job_base_001 --init queue_job,test_queue_job --test-tags="/queue_job,/test_queue_job" --stop-after-init
- Result: all tests green locally.

PR #861, opened 2025-12-18
[MIG] Odoo 19.0 has removed config.misc and config.get_misc()

daemonblade:
Odoo 19+ ignores non [options] sections. Reparse the configuration with ConfigParser so that existing rc files can be kept


hoangtrann (approved):
Code LGTM

sbidoul:
This is addressed in the main migration PR #840. Can you please use that one and confirm it works for you?

PR #865, opened 2026-01-02
[17.0][IMP]queue_job: only subscribe job creator if indicated via boolean method

GuillemCForgeFlow:
We add the `_subscribe_job_creator` method in `queue.job` which will return True for the cases where we want to subscribe the job creator, False otherwise.

cc @ForgeFlow

### DIFF
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -350,8 +350,11 @@ class QueueJob(models.Model):
         # at every job creation
         domain = self._subscribe_users_domain()
         base_users = self.env["res.users"].search(domain)
+        suscribe_job_creator = self._subscribe_job_creator()
         for record in self:
-            users = base_users | record.user_id
+            users = base_users
+            if suscribe_job_creator:
+                users |= record.user_id
             record.message_subscribe(partner_ids=users.mapped("partner_id").ids)
             msg = record._message_failed_job()
             if msg:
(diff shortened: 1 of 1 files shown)


GuillemCForgeFlow:
hi @guewen, any chance you see this a good improvement to the module?
thank you in advance 🙂

sbidoul:
@GuillemCForgeFlow why is it not sufficient for you to override `_subscribe_users_domain` ?

GuillemCForgeFlow:
> @GuillemCForgeFlow why is it not sufficient for you to override `_subscribe_users_domain` ?

@sbidoul, the problem is that the user assigned to the job will always be added anyway: https://github.com/OCA/queue/blob/17.0/queue_job/models/queue_job.py#L359. There may be cases for which we don't want that to happen.
I was also thinking about solving with `queue_job_subscribe` but thought this would be best. Any ideas on how to solve alternatively are welcomed 🙂

sbidoul:
> the problem is that the user assigned to the job will always be added anyway

Ah yes.

How about this?

```python
def _subscribe_job_creator(self) -> bool
    """Whether the user that created the job should be subscribed to the job, in addition to users determined by `_subscribe_users_domain`"""
```

GuillemCForgeFlow:
@sbidoul changes applied, thank you for the feedback 🙂 👍🏿

GuillemCForgeFlow:
hi @guewen, I'd appreciate it if you have the time to review this small improvement 🙂

guewen (approved):
Thanks

GuillemCForgeFlow:
@sbidoul perhaps we can get this merged? I'll take care of the fws 👌🏿

sbidoul:
/ocabot merge patch

Thanks!

PR #866, opened 2026-01-02
[18.0] queue_job: refactor job acquisition

### LINKED ISSUE #858: After requeueing a job, first run often fails to update date_done with concurrent error

sbidoul:
In this PR we cleanly separate the job acquisition (i.e. verifying the job is in the exepected state, marking it started and locking it) from job execution.

We also avoid trying to start the job if it is already locked by using SKIP LOCKED and exiting early. Indeed, in such situations, the job is likely already being handled by another worker so there is no point trying to start it, so we exit early and let it be handled either by the other worker or the dead job requeuer.

Following-up on https://github.com/OCA/queue/pull/859#issuecomment-3641850078

maybe fixes #858

### DIFF
--- a/test_queue_job/tests/test_requeue_dead_job.py
+++ b/test_queue_job/tests/test_requeue_dead_job.py
@@ -13,23 +13,6 @@ from .common import JobCommonCase
 
 @tagged("post_install", "-at_install")
 class TestRequeueDeadJob(JobCommonCase):
-    def _get_demo_job(self, uuid):
-        # job created during load of demo data
-        job = self.env["queue.job"].search(
-            [
-                ("uuid", "=", uuid),
-            ],
-            limit=1,
-        )
-
-        self.assertTrue(
-            job,
-            f"Demo data queue job {uuid} should be loaded in order"
-            " to make this tests work",
-        )
-
-        return job
-
     def get_locks(self, uuid, cr=None):
         """
         Retrieve lock rows
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -27,15 +27,47 @@ DEPENDS_MAX_TRIES_ON_CONCURRENCY_FAILURE = 5
 
 
 class RunJobController(http.Controller):
-    def _try_perform_job(self, env, job):
-        """Try to perform the job."""
+    @classmethod
+    def _acquire_job(cls, env: api.Environment, job_uuid: str) -> Job | None:
+        """Acquire a job for execution.
+
+        - make sure it is in ENQUEUED state
+        - mark it as STARTED and commit the state change
+        - acquire the job lock
+
+        If successful, return the Job instance, otherwise return None. This
+        function may fail to acquire the job is not in the expected state or is
+        already locked by another worker.
+        """
+        env.cr.execute(
+            "SELECT uuid FROM queue_job WHERE uuid=%s AND state=%s "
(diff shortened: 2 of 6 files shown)


test_queue_job/tests/test_requeue_dead_job.py
```diff
@@ -13,23 +13,6 @@
...
-            limit=1,
-        )
-
-        self.assertTrue(
-            job,
-            f"Demo data queue job {uuid} should be loaded in order"
-            " to make this tests work",
-        )
-
-        return job
```
sbidoul, line 31: Moved to JobCommonCase

sbidoul:
This is ready to review.

It can be tested with a small number of workers and a root channel capacity greater than the number of workers and test jobs of duration >= 10 sec.

If the test graph has enough parallelism, you will see warnings about dead jobs being requeued (the jobs in state enqueued that are not picked up by workers in due time), but each job should execute only once.

Example: `odoo --workers=3`, `root:10` and `http://odoo/queue_job/create_test_job?size=20&job_duration=10`.

The `_acquire_job` method is now simple enough that it can be unit tested. Next I may try refactoring the perform job part buy its a bit scarier :) But first, merging 19.0.

hoangtrann (approved):
Code LGTM

hoangtrann:
I've also tested this locally and and it seems to work as intended
image

sbidoul:
Thanks for the review and testing @hoangtrann !

I added a couple of commits to split out the job execution logic to class method so it is reusable from outside the controller and open the door to experiments such as #871.

guewen (approved):

PR #871, opened 2026-01-03
[18.0][WIP] queue_job: run jobs in cron workers

sbidoul:
This very WIP and draft and I'm not even sure this is useful.

One benefit is simpler deployment when the time limit of job workers has to be greater than the small time limit of interactive workers. In that (common) case configur the  cron worker time limit would be sufficient instead of deploying another Odoo instance with different worker limits.

It builds on the refactoring done in #866, in particular the `_acquire_job` function and job execution class methods.

The jobrunner still does the job scheduling (so with full support of channels etc), but `/queue_job/runjob` triggers crons which acquire jobs in ENQUEUED state and run them.

Look at the last commit to understand how this works, it's pretty simple.

### DIFF
--- a/queue_job/__manifest__.py
+++ b/queue_job/__manifest__.py
@@ -20,6 +20,7 @@
         "wizards/queue_requeue_job_views.xml",
         "views/queue_job_menus.xml",
         "data/queue_data.xml",
+        "data/queue_job_executor_cron.xml",
         "data/queue_job_function_data.xml",
     ],
     "assets": {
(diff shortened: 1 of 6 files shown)


sbidoul:
One thing I don't quite understand is why this can raise `SerializationFailure`.

```sql
SELECT uuid FROM queue_job WHERE state='enqueued' LIMIT 1 FOR UPDATE SKIP LOCKED
```

PR #875, opened 2026-01-08
[18.0][FIX] test_queue_job: also add an _unregister_hook function that reverts patches

thomaspaulb:
Otherwise monkey-patches could be added twice when instantiating new test classes.

### DIFF
--- a/test_queue_job/models/test_models.py
+++ b/test_queue_job/models/test_models.py
@@ -139,6 +139,13 @@ class ModelTestQueueJob(models.Model):
         )
         return super()._register_hook()
 
+    def _unregister_hook(self):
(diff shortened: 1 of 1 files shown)


sbidoul:
Is that due to a recent change or has this been there since like forever?

thomaspaulb:
Probably since 2021? https://github.com/OCA/queue/pull/275/files

I only ran into it because I backported your change from 18.0 that adds an extra test class, thus Odoo applies the hook twice; and in 16.0 there is still a test that checks if the warning that's emitted by the patched methods is emitted only once, and it was emitted twice. But in 18.0 it wouldn't be apparent because that test was removed.

sbidoul:
/ocabot merge patch

Thanks!

PR #878, opened 2026-01-14
[16.0][FIX] base_import_async: use chunk size defined by user in front when creating import batches

AnizR:
# Goal

Define the chunk size when importing in background

# Context

When realizing big import, we have the option to define batch sizes for import (feature from odoo's base_import):
image

But, I realized that base_import_async wasn't using this field and so, it does only batches of 100 records (default value).
This field is set as `options["limit"]` by odoo's front (in .js). Therefore, I have updated the `OPT_CHUNK_SIZE` because "chunk_size" is never defined.


lmarion-source (approved):

RyanTranAVB (approved):

sbidoul:
Looks good. Thanks!

/ocabot merge patch

PR #886, opened 2026-01-26
[REF] queue_job: documentation cleanup

RLeeOSI:
Cleanup of documentation in python file, is already present in readme file. Also adds clarity on env vars and conf file options


RLeeOSI:
@sbidoul First part of the split PR

amh-mw (approved):
Confirmed that deleted comments appear in CONFIGURE.md

sbidoul:
> Confirmed that deleted comments appear in CONFIGURE.md

@guewen I think there used to be a sphinx docs generator that was picking up those comments. I suppose we can consider this is lost in time? Or is it something that's worth reviving?

guewen:
> I suppose we can consider this is lost in time?

Yeah I would say so, better put time in the readme.


Thanks for the cleaning @RLeeOSI

guewen (approved):

sbidoul:
/ocabot merge patch

Thanks!

PR #893, opened 2026-02-11
[MIG] export_async_schedule: Migration to 17.0

StephaneMangin:
depends : https://github.com/OCA/queue/pull/894


StephaneMangin:
@yvaucher @simahawk @cuongnmtm 
Hi guys, a little up about this module migration.
Thanks a lot in advance for your reviews.

yvaucher (approved):

StephaneMangin:
A improvement is incoming for this module https://github.com/OCA/queue/pull/897

export_async_schedule/models/export_async_schedule.py
```diff
@@ -62,12 +62,10 @@ class ExportAsyncSchedule(models.Model):
     )
     end_of_month = fields.Boolean()
 
-    def name_get(self):
-        result = []
+    @api.depends("model_id", "ir_export_id")
```
cuongnmtm, line 65: Should we depend on model_id.name and ir_export_id.name?

cuongnmtm (changes requested):
I noticed that you are using two spaces instead of four for XML file indentation, which introduces multiple unnecessary changes.
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#21format

If I am not mistaken, I believe the authors value in the manifest file should not be updated, as it is intended to reflect the original author of the module. We are contributors.

PR #895, opened 2026-02-12
[18.0] Add queue_job_profiler

simahawk:


queue_job_profiler/models/queue_job_function.py
```diff
@@ -0,0 +1,48 @@
...
+        string="Profiling until",
+        help="The date and time until which profiling is enabled "
+        "for this job function.",
+    )
+
+    def is_profiling_enabled(self):
+        self.ensure_one()
+        return (
+            self.profiling_enabled
+            and self.profiling_user_id == self.env.user
```
simahawk: @mmequignon as discussed together: I'm not so sure limiting by user all the times is a good idea.
We might want to monitor the same function but for multiple users at once.

mmequignon: This could be interesting, but should be optional.
I'd say if we don't set a user, it profiles for everybody, otherwise just the user.
Watchu think ?

simahawk: done ;)

mmequignon: Cool, on my way to test it. Gonna drop a review after.

queue_job_profiler/models/queue_job.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2019 Creu Blanca
```
TDu: ?

queue_job_profiler/models/queue_job.py
```diff
@@ -0,0 +1,43 @@
...
+        string="Profiled",
+        default=False,
+        help="Whether this job has been profiled or not.",
+        compute="_compute_job_is_profiled",
+    )
+
+    def _compute_job_is_profiled(self):
+        for job in self:
+            # don't care about perf as this is loaded only on the job form view
+            profile_name = job.job_function_id._profile_make_name(job)
```
TDu, line 25: The `job_function_id` is not a required field for a job !
This could crash.

divad1196:
2 similar branches exist:
- [OCA/18-qjob-profiler](https://github.com/OCA/queue/tree/18-qjob-profiler)
- [camptocamp/18-qjob-profiler](https://github.com/camptocamp/queue/tree/18-qjob-profiler)

The OCA one had 2 commits more than the camptocamp one. I took the missing commits.

@simahawk I think we can close the OCA branch ?

queue_job_profiler/models/queue_job.py
divad1196: @TDu Fixed.

It was probably due to a copy/paste. It was fixed by @simahawk in another repository.

queue_job_profiler/models/queue_job.py
divad1196: @TDu You were right.

From my understanding, this field is mostly informative, same for the output of `_profile_make_name`.

if not defined, `job_function_id` will be an empty recordset, so I changed `_profile_make_name` accordingly: https://github.com/OCA/queue/pull/895/changes/919e0d2b6f6126bb7d5cdad91dfe08d6746dfba1

Does it look okay to you?

TDu: Yes, thank you. We may also have an issue with `is_profiling_enabled` function if the job function is no set, could you check.

queue_job_profiler/models/queue_job_function.py
```diff
@@ -0,0 +1,49 @@
...
+        self.ensure_one()
+        return (
+            self.profiling_enabled
+            and (self.profiling_until and self.profiling_until >= fields.Datetime.now())
+            and (
+                not self.profiling_user_ids or self.env.user in self.profiling_user_ids
+            )
+        )
+
+    @api.constrains("profiling_enabled")
```
SilvioC2C, line 35: This constraint should track `profiling_until` too.
Also: maybe it'd be useful to check that `profiling_until` is actually set in the future, instead of the past.

queue_job_profiler/tests/test_job_profiler.py
```diff
@@ -0,0 +1,148 @@
...
+            # After the profiling_until date, profiling should be disabled for all users
+            for user in (self.env.user, self.user1, self.user2):
+                self.assertFalse(self.user_func1.with_user(user).is_profiling_enabled())
+                self.assertFalse(self.user_func2.with_user(user).is_profiling_enabled())
+
+    def test_no_queue_job_function(self):
+        job1 = self.env.user.with_delay().read(["id"])
+        job1.store()
+        job_rec1 = job1.db_record()
+        self.env["queue.job.function"]._profile_make_name(job_rec1)
```
SilvioC2C: Please add some comments to describe what's being tested.

Same applies to most other tests as well.

queue_job_profiler/models/queue_job.py
```diff
@@ -0,0 +1,44 @@
...
+
+    job_is_profiled = fields.Boolean(
+        string="Profiled",
+        default=False,
+        help="Whether this job has been profiled or not.",
+        compute="_compute_job_is_profiled",
+        compute_sudo=True,
+    )
+
+    def _compute_job_is_profiled(self):
```
SilvioC2C, line 19: Please decorate this method with `@api.depends(...)`, so that the field's value can be handled properly (eg: invalidated and recomputed correctly when needed).

queue_job_profiler/models/queue_job_function.py
```diff
@@ -0,0 +1,49 @@
...
+                    self.env._(
+                        "A profiling until date must be set when profiling is enabled."
+                    )
+                )
+
+    def _profile_make_name(self, job):
+        if not self:
+            return f"queue.job {job.uuid} - <anonymous>"
+        self.ensure_one()
+        return f"queue.job {job.uuid} - {self.name}"
```
SilvioC2C:
```suggestion
        return f"queue.job {job.uuid} - {self.name or '<anonymous>'}"
```

simahawk: this fix is not what we want, I'll followup ;)

queue_job_profiler/models/queue_job_function.py
simahawk: Not necessary now.

simahawk:
job function issue fixed, rebased and squashed

TDu (approved):

PR #896, opened 2026-02-17
Update dotfiles

sbidoul:


amh-mw (changes requested):
It looks like the .pylint changes have lost alphabetical order? This makes it difficult to determine what has changed, if anything.

sbidoul:
> It looks like the .pylint changes have lost alphabetical order? This makes it difficult to determine what has changed, if anything.

I just applied the OCA repo template, that change comes right from there.

sbidoul:
@amh-mw specifically this change to rules order comes from https://github.com/OCA/oca-addons-repo-template/pull/317

amh-mw:
I will watch that repository from now on. Thanks!

PR #898, opened 2026-02-18
[FIX] mock.patch(auto_spec=True) does not exist

gurneyalex:
There is no parameter auto_spec to mock.patch. This is likely a typo.

Running the tests with Python 3.12 give a RuntimeError
```
  File "/usr/local/lib/python3.12/unittest/mock.py", line 1311, in __init__
    _check_spec_arg_typos(kwargs)
  File "/usr/local/lib/python3.12/unittest/mock.py", line 1287, in _check_spec_arg_typos
    raise RuntimeError(
RuntimeError: 'auto_spec' might be a typo.
```

### DIFF
--- a/queue_job/tests/common.py
+++ b/queue_job/tests/common.py
@@ -96,7 +96,6 @@ def trap_jobs():
     with mock.patch(
         "odoo.addons.queue_job.delay.Job",
         name="Job Class",
-        auto_spec=True,
         unsafe=True,
     ) as job_cls_mock:
         with JobsTrap(job_cls_mock) as trap:


queue_job/tests/common.py
```diff
@@ -96,7 +96,6 @@ def button_that_uses_delayable_chain(self):
     with mock.patch(
         "odoo.addons.queue_job.delay.Job",
         name="Job Class",
-        auto_spec=True,
```
amh-mw, line 99: Per https://docs.python.org/3/library/unittest.mock.html#auto-speccing?
```diff
-         auto_spec=True,
+         autospec=True,
```

amh-mw (changes requested):

queue_job/tests/common.py
gurneyalex: yes, I thought about that too. The thing is that `auto_spec` never did anything, and things have been working so far, which is why I'm tempted to just remove the kwarg.

gurneyalex: and removing it does *not* break the tests...

sbidoul (approved):

amh-mw (approved):

gurneyalex:
I will backport this.

guewen (approved):

PR #883, opened 2026-01-23
[18.0][FIX] Don't try to enqueue dependent jobs when job failed

thomaspaulb:
Running (on 16.0) into:

```
2026-01-23 13:45:15,992 351151 ERROR custdatabase odoo.sql_db: bad query: UPDATE "stock_quant" SET "reserved_quantity" = '45212.0000', "write_date" = '2026-01-23 13:45:15.153636', "write_uid" = 1 WHERE id IN (1793638)
ERROR: could not serialize access due to concurrent update
 
2026-01-23 13:45:15,994 351151 DEBUG custdatabase odoo.addons.queue_job.controllers.main: <Job 7f20f326-3383-4c42-9cdf-7126dbd89743, priority:2> OperationalError, postponed 
2026-01-23 13:45:16,000 351151 DEBUG custdatabase odoo.addons.queue_job.controllers.main: <Job 7f20f326-3383-4c42-9cdf-7126dbd89743, priority:2> postponed 
2026-01-23 13:45:16,001 351151 DEBUG custdatabase odoo.addons.queue_job.controllers.main: <Job 7f20f326-3383-4c42-9cdf-7126dbd89743, priority:2> enqueue depends started 

  File "/data/home/odoo/odoo/auto/addons/muk_rest/models/ir_http.py", line 160, in _dispatch
    response = super()._dispatch(endpoint)
  File "/data/home/odoo/odoo/.venv/lib/python3.10/site-packages/odoo/addons/base/models/ir_http.py", line 154, in _dispatch
    result = endpoint(**request.params)
  File "/data/home/odoo/odoo/.venv/lib/python3.10/site-packages/odoo/http.py", line 761, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/data/home/odoo/odoo/auto/addons/queue_job/controllers/main.py", line 200, in runjob
    self._runjob(env, job)
  File "/data/home/odoo/odoo/auto/addons/queue_job/controllers/main.py", line 169, in _runjob
    cls._enqueue_dependent_jobs(env, job)
  File "/data/home/odoo/odoo/auto/addons/queue_job/controllers/main.py", line 86, in _enqueue_dependent_jobs
    with job.env.cr.savepoint():
  File "/data/home/odoo/odoo/.venv/lib/python3.10/site-packages/odoo/sql_db.py", line 157, in savepoint
    return _FlushingSavepoint(self)
[...]

PR #900, opened 2026-02-19
[18.0] Skip check of dependencies when a done job has no dependents

guewen:
Every time a job is done, even if it is not part of a graph, it runs a query to look for dependents to enqueue. Storing the dependent uuids in the "dependencies" field was on purpose to know that we have no further jobs in the graph and that we can skip the check entirely and have no overhead in this case.

It looks like an oversight, we can add the missing condition.

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -108,6 +108,10 @@ class RunJobController(http.Controller):
 
     @classmethod
     def _enqueue_dependent_jobs(cls, env, job):
+        if not job.should_check_dependents():
+            return
+
+        _logger.debug("%s enqueue depends started", job)
         tries = 0
         while True:
             try:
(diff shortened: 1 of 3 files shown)


queue_job/controllers/main.py
```diff
@@ -182,9 +182,10 @@ def retry_postpone(job, message, seconds=None):
                 buff.close()
             raise
 
-        _logger.debug("%s enqueue depends started", job)
-        cls._enqueue_dependent_jobs(env, job)
-        _logger.debug("%s enqueue depends done", job)
+        if job.should_check_dependents():
```
amh-mw: Would it be a better defensive posture to put this check inside `_enqueue_dependent_jobs` and let it no-op?

amh-mw (changes requested):

queue_job/controllers/main.py
guewen: Updated

amh-mw (approved):
Code review only, LGTM.

sbidoul:
Looks good. While you are at it, could you add a return at line 174 of the controller (https://github.com/OCA/queue/pull/883#issuecomment-3790518648)?

sbidoul (approved):

PR #899, opened 2026-02-19
[18.0] Add 'Allow Commit' option on job functions

guewen:
It is forbidden to commit inside a job, because it releases the job lock and can cause it to start again, while still being run, by the dead jobs requeuer. For some use cases, it may actually be legitimate, or at least be needed in the short term before actual updates in the code.

A new option on the job function, false by default, allow to run the job in a new transaction, at the cost of an additional connection + transaction overhead.

Related to #889

PR #910, opened 2026-03-17
[16.0] Add 'Allow Commit' option on job functions

lmignon:
Backport of #899


guewen (approved):

florian-dacosta (approved):
LGTM (tested)

sbidoul:
@guewen I kind of remember you could set `allow_commit` in `with_delay()` in your original implementation. Was that removed on purpose, or is it me who imagined it?

guewen:
@sbidoul It was only in your head 😂 I didn't think about that. Do you see a use case for this? For methods that accept a "commit=True" parameter?

florian-dacosta:
On adapting `attachment_queue` module, where no `queue.job.function` was defined, I first tried to pass the allow_commit to the `with_delay` method here https://github.com/OCA/server-tools/pull/3577/changes#diff-21c8389cfe9291d701d45d4860731ff11ac1ce8ddd2317a07479852488bca7b9R57) I was kind of surprised it was not implemented. But I am not sure it is necessary to change anything about this

sbidoul:
Yes that was what prompted my question ^.

guewen:
I suppose most of the time, we know ahead of time when a job method will/should commit or not, so it makes more sense on the job function. If you had to set `.with_delay(allow_commit=True)` in *all* the delayed calls of the job, it's not the place where the parameter should be IMO.

sbidoul:
@lmignon is this good to go?

queue_job/views/queue_job_function_views.xml
```diff
@@ -11,6 +11,7 @@
                     <field name="model_id" required="1" />
                     <field name="method" required="1" />
                     <field name="channel_id" />
+                    <field name="allow_commit" />
```
florian-dacosta, line 14: @lmignon 
>A little part is missing.... The new field is not declared into the queue.job.function view  

Not sure I understand your comment, this is it, right ?

lmignon: Am I so stupid... , I tried to configure the wrong odoo instance with the previous version installed... it's friday

PR #923, opened 2026-04-22
[FIX] queue_job: repoint default_env to SUPERUSER in runjob

### LINKED ISSUE #922: [19.0] `env.user` empty during flush/recompute inside queue jobs
## Module
queue_job

## Describe the bug
https://github.com/OCA/queue/blob/19.0/queue_job/controllers/main.py#L218
```
env = http.request.env(user=SUPERUSER_ID)
```
I believe needs to be
```
http.request.update_env(user=SUPERUSER_ID)
env = http.request.env
```

`runjob` is declared as `auth=none`.
https://github.com/OCA/queue/blob/19.0/queue_job/controllers/main.py#L212
`_auth_method_none` forces `uid=None` for request env and default_env
https://github.com/odoo/odoo/blob/19.0/odoo/addons/base/models/ir_http.py#L263

So `env = http.request.env(user=SUPERUSER_ID)`
only produces a local env with that superuser, the other two are still None.

The job "itself" runs fine as the specified user, but any flush/recompute on a stored field uses the default_env with uid `None`.
We see a singleton error on `is_backend_user = self.env.user.has_group('base.group_user')` (in Odoo Enterprise `compute_signing_user`) because `self.env.user` is an empty recordset.

```
File ".../odoo/orm/environments.py", line 592, in flush
    self.default_env.flush_all()
...
File ".../enterprise/account_accountant/models/account_move.py", line 76, in _compute_signing_user
    is_backend_user = self.env.user.has_group('base.group_user')
File ".../odoo/addons/base/models/res_users.py", line 1074, in has_group
    self.ensure_one()
ValueError: Expected singleton: res.users()
```

The default_env mechanism was introduced 2 years ago
https://github.com/odoo/odoo/commit/897e2f2dff618b1c322da213ec955d383369c301
but the flush() was tolerant. Ironically,
https://github.com/odoo/odoo/commit/45e1c05a75d357b6c883e125f57bb4421a3d39fa
(Jan 24 2025) states "flush as public user when there is no default_env"
but now `_auth_method_none` _always_ sets `default_env`
https://github.com/odoo/odoo/commit/45e1c05a75d357b6c883e125f57bb4421a3d39fa#diff-c5500b83255cdef56ed07cb5a74e9b0899ad378cf9a3739d9676101191c83eb2

Using `update_env` repoints the default_env correctly
https://github.com/odoo/odoo/blob/19.0/odoo/http.py#L1836

## To Reproduce
**Affected versions**: 19.0

1. Create a module with a computed attribute that accesses `self.env.user` (I added an explicit flush to force the pending compute):

```
    _inherit = "res.partner"

    x_toucher = fields.Many2one("res.users", store=True, compute="_c")

    @api.depends("name")
    def _c(self):
        self.env.user.has_group("base.group_user")  # ensure_one() fails on empty recordset
        for r in self:
            r.x_toucher = self.env.user

    def touch(self):
        self.write({"name": "x"})
        self.env.cr.flush()  # forces Transaction.flush() -> default_env.flush_all()
```

2. Enqueue a write that causes a flush: `partner.with_delay().touch()`
3. Run the jobrunner -> job fails with Expected singleton: res.users().

**Expected behavior**
`self.env.user` inside stored-field recomputes triggered during a queue job should resolve to a real user (the job's user_id or SUPERUSER), not an empty recordset.

**Additional context**
In the same commit (45e1c05), core updated the two other places that create a uid=None env to set default_env explicitly:

 - odoo/service/common.py::exp_authenticate assigns `env.transaction.default_env = env` directly.
 - addons/bus/websocket.py added a `new_env()` helper that does the same.

Neither can use `Request.update_env` because they don't run inside an HTTP Request. 

queue_job/controllers/main.py::runjob does run inside a Request, so the idiomatic fix is `http.request.update_env(user=SUPERUSER_ID)`, which has been repointing default_env since 897e2f2.

dannyadair:
runjob is declared auth="none", so _auth_method_none pins request.env and transaction.default_env to a uid=None env. `env = http.request.env( user=SUPERUSER_ID)` only creates a local superuser env, leaving the default_env as uid=None. Any flush that goes through Transaction.flush() -> default_env.flush_all() then recomputes stored fields as uid=None and fails on anything dereferencing self.env.user.

`http.request.update_env(user=SUPERUSER_ID)` additionally sets transaction.default_env to the superuser env.

Closes #922

### DIFF
--- a/queue_job/tests/test_run_job_controller_http.py
+++ b/queue_job/tests/test_run_job_controller_http.py
@@ -1,4 +1,4 @@
-# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
 
 from unittest import mock
 
--- a/queue_job/tests/test_run_job_controller_http.py
+++ b/queue_job/tests/test_run_job_controller_http.py
@@ -1,0 +1,85 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from unittest import mock
+
+from odoo import SUPERUSER_ID, http
+from odoo.tests.common import HttpCase, tagged
+
+from ..controllers.main import RunJobController
+
+
+@tagged("-at_install", "post_install")
+class TestRunJobHttp(HttpCase):
+    """End-to-end tests for the ``/queue_job/runjob`` endpoint.
+
+    ``runjob`` is declared ``auth="none"``, so ``_auth_method_none`` pins
+    ``transaction.default_env`` to an env with ``uid=None``. The controller
+    must additionally repoint ``default_env`` to a superuser env, otherwise
+    any flush during ``job.perform()`` that goes through
+    ``Transaction.flush() -> default_env.flush_all()`` recomputes stored
+    fields with ``self.env.user`` as an empty recordset. See #922.
+    """
+
+    def _capture_default_env_uid_at_acquire(self):
+        """Hit ``/queue_job/runjob`` and observe ``transaction.default_env.uid``
+        at the moment ``_acquire_job`` is entered.
+
+        ``_acquire_job`` is patched to return ``None`` so the endpoint exits
+        without performing the job. This avoids the commit inside
+        ``_acquire_job``, which is forbidden inside tests, and is sufficient
+        to observe the env state established by the request-handling
+        machinery: nothing between ``update_env`` and ``_acquire_job``
+        touches ``transaction.default_env``, so the value seen here is the
+        same value a job would see at ``perform()`` time.
+
+        HttpCase runs the HTTP server in the same process, so class-level
+        ``mock.patch`` calls take effect on the server side as well.
+        """
+        captured = {}
+
+        def spy(cls, env, job_uuid):
+            captured["default_env_uid"] = env.transaction.default_env.uid
+            return None
+
+        with mock.patch.object(RunJobController, "_acquire_job", classmethod(spy)):
+            response = self.url_open(
+                f"/queue_job/runjob?db={self.env.cr.dbname}&job_uuid=stub"
+            )
+            response.raise_for_status()
+
+        return captured.get("default_env_uid")
+
+    def test_runjob_pins_default_env_to_superuser(self):
+        """``runjob`` must set ``transaction.default_env`` to a superuser env.
+
+        Verifies that by the time ``_acquire_job`` is entered,
+        ``env.transaction.default_env.uid`` is ``SUPERUSER_ID``. This is the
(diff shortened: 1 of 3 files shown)


dannyadair:
Rebased after #924 was merged.
That PR removed the db/session.db handling but not the uid=None default_env issue this fixes, so the change is unchanged.

amh-mw (changes requested):
Seems reasonable after reading the source here and https://github.com/odoo/odoo/blob/19.0/odoo/http.py#L1856-L1867:

```python
    def update_env(self, user=None, context=None, su=None):
        """ Update the environment of the current request.

        :param user: optional user/user id to change the current user
        :type user: int or :class:`res.users record<~odoo.addons.base.models.res_users.ResUsers>`
        :param dict context: optional context dictionary to change the current context
        :param bool su: optional boolean to change the superuser mode
        """
        cr = None  # None is a sentinel, it keeps the same cursor
        self.env = self.env(cr, user, context, su)
        self.env.transaction.default_env = self.env
        threading.current_thread().uid = self.env.uid
```

~~That said, why make the change? What problem does this avoid?~~ ☕  Unit test that problem?

dannyadair:
The problem is described in https://github.com/OCA/queue/issues/922 and solved by this PR.

amh-mw:
> The problem is described in #922 and solved by this PR.

Yep, sorry! Struck through much of my comment already! Still desire a unit test.

dannyadair:
> > The problem is described in #922 and solved by this PR.
> 
> Yep, sorry! Struck through much of my comment already! Still desire a unit test.

ok with the existing tests this one would be weak and almost meaningless. (test_run_job_controller just calling class methods so we'd be mocking and verify that our two line change calls the two-line change), I 'll add something using HttpCase.

dannyadair:
ugh, sorry for the noise. _acquire_job commits which is forbidden in tests. Have to get tricky.

dannyadair:
added test

amh-mw (approved):
Code review only, LGTM.

queue_job/tests/test_run_job_controller_http.py
```diff
@@ -0,0 +1,85 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
guewen:
```suggestion
# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
```

guewen (commented):
Thanks @dannyadair, can you fix the license to match the one of the module please?

Ricardoalso (approved):
Thanks !

SilvioC2C (approved):

grindtildeath (approved):

MiquelRForgeFlow (approved):

PR #924, opened 2026-04-25
[19.0][IMP] queue_job: remove multi-db monkey patch

sbidoul:
By using the X-Odoo-Database HTTP header that is natively supported by Odoo since version 19, we achieve the same result.


guewen (approved):

hoangtrann (approved):

amh-mw (approved):
Code review only, LGTM.

Issue #889, opened 2026-02-09
commit is forbidden in queue jobs


sbidoul:
It is forbidden to call `env.cr.commit()` in delayed job functions.

It has always been bad practice, but now a commit releases the job lock and can cause it to be started again (while still running) by the dead jobs requeuer.

https://github.com/OCA/queue/pull/880 makes such commits an error to make that explicit.

I open this issue to collect feedback on this, as there might be (semi-)legitimate uses of commit.

amh-mw:
#### Use case:
Import hundreds of millions of rows from external systems with ordering (`chain`, `group`).

#### Initial Problems:
- Memory exhausted creating job execution graph.
- The `queue_job_batch` module doesn't interact with (`chain`, `group`).

#### Initial implementation:
- Custom batching module that imports multiple rows per job.
- Use of `self.env.cr.commit()` introduced to save position in batch and partial progress after exception thrown or prior to thread timeout (`SIGKILL`).
- For lightweight records without risk of `UniqueViolation` or `SerializationFailure`, batches are orders of magnitude faster than one record per job.

#### Better implementation?
- Improve `queue_job_batch` module with some sort of permanently retrying job that completes when the entire batch is successful or failed?

To be absolutely clear, **I am not arguing for** `self.env.cr.commit()`; this antiquated design originated _circa_ 12.0. I should have some cycles this quarter to work on `queue_job_batch` or whatever alternative approach seems best.

JordiMForgeFlow:
@sbidoul I understand the rationale behind the change, but seems to me that what you did is a bit risky. Any deployment that is running some queue logic doing commits is now failing due to the change, and we are forced to revert it temporarily until we can analyze the committing logics.

I am not saying I disagree with the proposal, but it could have been added first as a warning or similar rather than causing jobs to fail.

sbidoul:
@JordiMForgeFlow I understand but we think it's better to have an explicit failure than very surprising and hard to debug issues with jobs running multiple times. If you need commits in jobs, I strongly recommend pinning a queue_job version that does not have the automatic dead jobs requeuer (#716 and related).

JordiMForgeFlow:
@sbidoul makes sense, will take that into account. Maybe a communication could be send to contributors just so they know? I feel like queue is a widely used module and I am also sure that a lot of people tend to explicitly commit on queue jobs...

iggups:
[...]

PR #926, opened 2026-04-29
[16.0][FIX] queue_job: rebind args/kwargs to inner cursor in in_temporary_env

deeniiz:
Job.in_temporary_env() rebinds self.recordset to the inner cursor via the @env.setter, but leaves self.args and self.kwargs on the outer worker cursor where _prevent_commit is patched. A cr.commit() reached through any recordset argument then raises RuntimeError: Commit is forbidden in queue jobs even with allow_commit=True. Typical case: a connector exporter taking a binding and a related record, committing inside binder.bind_export(...).            
  
Introduced by #910.                                 

Fix: after the env swap, walk self.args and self.kwargs recursively (BaseModel, list, tuple, dict) and rebind each recordset with value.with_env(value.env(cr=new_env.cr)) — only the cursor changes, uid/su/context are kept. Originals restored in the finally, like self._env.
                                                                                                                                                          
Regression test in test_queue_job passes the same recordset as positional arg, list element and dict value, checks cursor identity in all three, then commits. Run through RunJobController._runjob.

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -13,8 +13,26 @@ from datetime import datetime, timedelta
 from random import randint
 
 import odoo
+from odoo.models import BaseModel
 
 from .exception import FailedJobError, NoSuchJobError, RetryableJobError
+
+
+def _rebind_to_env(value, new_env):
+    """Rebind any BaseModel inside `value` to ``new_env``'s cursor.
+
+    Recurses into lists, tuples and dicts. Preserves uid/su/context of each
+    inner env - only the cursor is swapped. Non-recordset values pass through
+    untouched.
+    """
+    if isinstance(value, BaseModel):
+        return value.with_env(value.env(cr=new_env.cr))
+    if isinstance(value, (list, tuple)):
+        return type(value)(_rebind_to_env(v, new_env) for v in value)
+    if isinstance(value, dict):
+        return {k: _rebind_to_env(v, new_env) for k, v in value.items()}
+    return value
+
 
 WAIT_DEPENDENCIES = "wait_dependencies"
 PENDING = "pending"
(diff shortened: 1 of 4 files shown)


deeniiz:
@eantones @guewen, can you please review it?

eantones:
Rebased on current 16.0; all checks green (codecov/patch 100%).

@guewen @sbidoul gentle ping — this completes the `allow_commit` option from #899/#910: currently `in_temporary_env()` rebinds `self.recordset` only, so any recordset passed as a job *argument* keeps the outer commit-forbidden cursor, and the job still raises `RuntimeError: Commit is forbidden in queue jobs` even with the option enabled. Typical hit: connector-style `export_record(backend, ...)` jobs committing bindings so external ids are not lost — the case discussed in #889. The fix rebinds args/kwargs exactly like `self.recordset` (cursor-only swap, originals restored on exit), with a regression test through `RunJobController._runjob`.

Happy to port this to 18.0 and 19.0 right after approval.

sbidoul:
I'll try to find time to look into this. It will not before end of next week though.

With this and #923, I can't help thinking there is something subtly wrong in the way we manage job envs and some refactoring could make things clearer and less fragile. Not sure.

guewen:
I was looking at this this morning, and on one side it seems to me in the current state, this is the only way forward, but I have the same worry as @sbidoul , there is something that looks wrong if we have to do that.


On the details (really), I would have added args and kwargs as "lazy" methods that rebind on the fly, something along the lines of

```py
    @property
    def args(self):
        return _rebind_env(self._args, self.env.cr) # also _rebind_env doesn't the env, only cr
```

Thanks for the investigation and correction

guewen (approved):

eantones:
Hi @guewen, you got me thinking with your "lazy" idea. I tried it and I think it's actually the better solution, so I implemented it on a separate branch: https://github.com/nuobit/queue/pull/1

args/kwargs stay in _args/_kwargs and are rebound on access (only the cr is swapped, uid/su/context are kept), so in_temporary_env doesn't need to touch them anymore. Same test, coverage stays at 100%.

I can update this PR to that version if you prefer. The 18.0/19.0 ports will follow whatever shape this one ends up with.

cc @sbidoul

benwillig (approved):
LGTM

guewen:
Hi @eantones , I had a response in draft and not posted it, sorry :/
I prefer the implementation in https://github.com/nuobit/queue/pull/1 yes!
Thanks for trying this approach

deeniiz:
Superseded by #967.

PR #927, opened 2026-04-30
[18.0][FIX] queue_job: remove create/delete permissions

guewen:
Jobs are always created through sudo() in the Job class, and always deleted by the cron, never by users.

See discussion on https://github.com/OCA/queue/pull/802#issuecomment-4341581616


sbidoul (approved):

amh-mw (approved):
Functionally tested locally. LGTM.

PR #931, opened 2026-04-30
[18.0][FIX] queue_job: Restore doctesting.

### LINKED ISSUE #780: [18.0]  Tests "doctest" are no longer run since Odoo 18
It can be seen on Github CI for 18.0 branch and related PRs.
For example the doctests embedded in `queue_job.jobrunner.channels` are not run.

This is due to changes in Odoo test loader:
- odoo/odoo#148660
- odoo/odoo#157121

~Maybe it is because of the 2nd PR above.~

It is because Odoo test loader does not call anymore standard `unittest.TestLoader().loadTestsFromModule(...)`

amh-mw:
Fixes #780

### DIFF
--- a/queue_job/tests/common.py
+++ b/queue_job/tests/common.py
@@ -1,15 +1,10 @@
 # Copyright 2019 Camptocamp
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-import doctest
-import logging
 import typing
 from contextlib import contextmanager
 from itertools import groupby
 from operator import attrgetter
 from unittest import TestCase, mock
-
-from odoo.tests.case import TestCase as _TestCase
-from odoo.tests.common import MetaCase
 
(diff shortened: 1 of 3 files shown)


amh-mw:
I'm not in love with this. Verbose output is ugly, so I turned it back off, but otherwise it is just a single line in the logs for potentially hundreds of doctests. I was expecting code coverage to increase more, but it looks like it only improved from the deletions in `tests.common`

guewen (approved):

sbidoul:
> I'm not in love with this. Verbose output is ugly, so I turned it back off, but otherwise it is just a single line in the logs for potentially hundreds of doctests. I was expecting code coverage to increase more, but it looks like it only improved from the deletions in `tests.common`

It's better than no doctests at all. Thanks for tackling this!

/ocabot merge patch

PR #932, opened 2026-05-09
[IMP] queue_job_batch: improve perf

hoangtrann:
In an instance with a lot of queue jobs, lacking index on `job_batch_id` and proper `read_group` on job batch can lead to full seq scan on db and cause performance degrade.

Before the changes, on a local database with roughly 100k rows and 10% of it having `job_batch_id`

|Function|Execution Time (before)|Execution Time (after)|
|-|-|-|
|`_compute_job_count`|6.971ms|1.611ms|

On a production database where the performance issue occurs, the time it takes to execute the query is actually much worst

### DIFF
--- a/queue_job_batch/models/queue_job.py
+++ b/queue_job_batch/models/queue_job.py
@@ -9,7 +9,7 @@ from odoo.addons.queue_job.job import identity_exact
 class QueueJob(models.Model):
     _inherit = "queue.job"
 
-    job_batch_id = fields.Many2one("queue.job.batch")
+    job_batch_id = fields.Many2one("queue.job.batch", index="btree_not_null")
 
     @api.model_create_multi
     def create(self, vals_list):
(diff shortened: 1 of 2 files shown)


guewen (approved):

dnplkndll:
@hoangtrann 

took a look.  I was wondering about the index cost versus options and research returned:

 Consider index="btree_not_null" rather than index=True — most  queue.job rows have NULL job_batch_id, so a partial index gives the  same speedup at a fraction of the disk + write cost.

 The same N+1 pattern lives in _get_state (hotter path — called  per job completion via check_state). Worth applying the same  read_group refactor in this PR or a follow-up?

dnplkndll (approved):
LGTM

hoangtrann:
@dnplkndll suggestions sound good, I've added both

PR #936, opened 2026-05-19
[IMP] queue_job: Handle TypeError when JSON-encoding context.

amh-mw:
This nulls out unencodable values when serializing the job context, instead of raising an error.

### DIFF
--- a/queue_job/fields.py
+++ b/queue_job/fields.py
@@ -103,7 +103,10 @@ class JobEncoder(json.JSONEncoder):
             }
         elif isinstance(obj, lazy):
             return obj._value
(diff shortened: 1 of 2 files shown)


amh-mw:
I am of the opinion that `queue_job` should not be in the context-management game, so this is an effort to increase durability when the context may not have been curated.

sbidoul:
I don't know. I personally prefer fast and explicit failures than silent degradation that would be harder to diagnose.

What is the use case?

amh-mw:
I'm currently running a variant of #768 where I ripped out all of the context management code. Then when I adopted  `queue_job_batch`, I started getting JSON encoding errors.

PR #939, opened 2026-05-28
[MIG] base_import_async: Migration to 19.0

REANOVA-GIT:
## Description

`base_import_async` is still flagged `installable: False` with `version: 18.0.1.0.0` on the `19.0` branch. This PR bumps it to V19 and makes it installable, with the API changes required by V19 and the pylint fixes triggered by the strict OCA mandatory checks now applying to the module.

### Code

- Bumped `__manifest__.py`: `version` `18.0.1.0.0` → `19.0.1.0.0`, `installable` `False` → `True`.
- Removed `@api.returns("ir.attachment")` decorator on `_create_csv_attachment` (decorator no longer supported in V19; the method already returns an `ir.attachment` recordset).
- Replaced `_()` with `self.env._()` everywhere (W8161 `prefer-env-translation`, see [odoo/odoo#174844](https://github.com/odoo/odoo/pull/174844)).
- Switched to the kwarg form of `self.env._()` instead of `_("...") % {...}` formatting (W8301 `translation-not-lazy`). Renamed `%(from)s` / `%(to)s` to `%(row_from)s` / `%(row_to)s` because `from` is a Python reserved word and can't be used as a kwarg name.
- Removed now-unused imports (`_`, `api`).
- Annotated `BaseImportImport` with `# pylint: disable=no-wizard-in-models` — the class extends core `base_import.import` (which itself lives in core Odoo's `base_import/models/`), so following the upstream layout is the right thing here. Open to feedback if reviewers prefer moving the file to `wizard/`.

### Repo-level

- Removed `^base_import_async/|` from `.pre-commit-config.yaml` "NOT INSTALLABLE ADDONS" exclude list.
- Added `odoo-addon-base_import_async==19.0.*` to `setup/_metapackage/pyproject.toml` dependencies.

## Context

Discovered while migrating a large Odoo V12 → V19 deployment (Reanova ERP) where `base_import_async` is a dependency of a vendored module (`customer_file_importation`).

## Tests

Tested on a V19 install with a 35k-row partner CSV: import is correctly queued as a `queue.job` and processed in the background. The `ir.attachment` returned by `_create_csv_attachment` is properly linked to the queue job (`res_model="queue.job"`).

## Note on `queue_job_cron` (dropped from this PR)

This PR originally also bumped `queue_job_cron` to 19.0, but it was dropped because its `test_queue_job_cron.TestQueueJobCron.test_queue_job_cron_callback` fails on V19 with:

```
AssertionError: Cannot commit or rollback a cursor from inside a test,
this will lead to a broken cursor when trying to rollback the test.
Please rollback to a specific savepoint instead or open another cursor
if really necessary
```

V19 strengthened the test framework to forbid `cr.commit()` / `cr.rollback()` from within `TransactionCase`. The `_callback` test triggers an internal `cr.commit()` in `ir.cron._callback`, which the V19 test framework now treats as a hard failure.

That migration needs its own PR with a test refactor (savepoint-based or by mocking the cron callback's commit) and is out of scope here.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,6 @@
 exclude: |
   (?x)
   # NOT INSTALLABLE ADDONS
-  ^base_import_async/|
   ^queue_job_batch/|
   ^queue_job_cron/|
   ^queue_job_cron_jobrunner/|
--- a/base_import_async/__manifest__.py
+++ b/base_import_async/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Asynchronous Import",
     "summary": "Import CSV files in the background",
-    "version": "18.0.1.0.0",
+    "version": "19.0.1.0.0",
     "author": "Akretion, ACSONE SA/NV, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "website": "https://github.com/OCA/queue",
@@ -20,6 +20,6 @@
             "base_import_async/static/src/xml/import_data_sidepanel.xml",
         ],
     },
-    "installable": False,
+    "installable": True,
     "development_status": "Production/Stable",
 }
--- a/base_import_async/models/base_import_import.py
+++ b/base_import_async/models/base_import_import.py
@@ -9,7 +9,7 @@ import csv
 from io import BytesIO, StringIO, TextIOWrapper
 from os.path import splitext
 
-from odoo import _, api, models
+from odoo import models
 from odoo.models import fix_import_export_id_paths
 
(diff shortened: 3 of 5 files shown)


RaphaelVicini:
Thanks for this PR!
During functional review, I noticed that async imports can fail and we dont have the possibility to “Set value as empty” for unmatched values.

Example error:
`No matching record found for name 'luxembor' in field 'Country'`

RaphaelVicini:
@RaphaelVicini @REANOVA-GIT

PR #940, opened 2026-05-30
[IMP] queue_job: handle non-existing method job

RyanTranAVB:
When a Job is previously created with a valid method, at any point if job not yet executed and the method is no longer valid, job runner can be stuck trying to enqueue it. This commit allows to skip these jobs with proper exception and one can clean/fix them if needed.

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -17,7 +17,7 @@ from odoo.service.model import PG_CONCURRENCY_ERRORS_TO_RETRY
 from odoo.tools import config
 
 from ..delay import chain, group
-from ..exception import FailedJobError, RetryableJobError
+from ..exception import FailedJobError, JobMethodNotFound, RetryableJobError
 from ..job import ENQUEUED, Job
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


queue_job/controllers/main.py
```diff
@@ -78,7 +78,29 @@ def _acquire_job(cls, env: api.Environment, job_uuid: str) -> Job | None:
...
+                WHERE uuid  = %s AND state NOT IN ('done', 'cancelled', 'failed')
+                """,
+                (exc_name, exc_message, exc_message, job_uuid),
+            )
+            _logger.warning(
+                "Job %s references a non-existent method and was marked as failed",
+                job_uuid,
+            )
+            if not config["test_enable"]:
+                env.cr.commit()
```
amh-mw, line 103: Both the raw SQL and the ORM are using the same cursor; the same savepoint?

amh-mw (changes requested):

queue_job/controllers/main.py
amh-mw: Perhaps job.in_temporary_env() and then ORM methods instead of raw SQL?

RyanTranAVB: that's good point, I was overthinking it cause `JobMethodNotFound` wont cause any rollback, so it's safe using the same cursor I guess.

`in_temporary_env` won't work here because it's failing during loading the job as `getattr` fails

PR #941, opened 2026-06-04
[18.0][FIX] Adapt queue_job_profiler to job aquisition refactor

TDu:
That module was develop before this change

* https://github.com/OCA/queue/pull/866

The new module tests were not affected by the change when merging.

### DIFF
--- a/queue_job_profiler/controllers/main.py
+++ b/queue_job_profiler/controllers/main.py
@@ -7,24 +7,28 @@ from odoo.addons.queue_job.controllers.main import RunJobController as BaseContr
 
 
 class RunJobController(BaseController):
-    def _try_perform_job(self, env, job):
-        if self._profiler_is_enabled(env, job):
-            return self._profiler_perform_job(env, job)
+    @classmethod
+    def _try_perform_job(cls, env, job):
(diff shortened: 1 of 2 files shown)


amh-mw (approved):
Code review only, LGTM. Bonus points still available for a unit test that would have noticed the method signature change.

simahawk:
@TDu can you try to add the test?
As you are there, can you pls rewrite the commit msg to something like

```
[FIX] queue_job_profiler: controller hook

Broken by job acquisition refactoring #941
```
?

TDu:
About adding a test that would have cached the change and make the module fail. I agree it would have been good to have one, but I am not sure of the pertinence of adding it afterwards.
For such test to make sense it would need to call  the `_run_job` function and render the test suite more complex. 

But I have updated the tests to use the RunJobController class and not an instance of it.
And remove the `commit` mock, because it is not needed IMO https://github.com/OCA/queue/blob/1164f054e00d29dbcd53d1009dd37ba9bf75b6de/queue_job/controllers/main.py#L108

simahawk (approved):
@TDu pls squash

PR #942, opened 2026-06-04
[18.0][ADD] queue_job_is_running

Saran440:
This module adds a warning banner and a new boolean field (`is_job_running`) to help users understand that a document's status is running. Developers can also hide buttons to prevent users from clicking them and creating duplicate jobs.

Screenshot 2026-06-04 165005


queue_job_alert_running/readme/USAGE.md
```diff
@@ -0,0 +1,30 @@
...
+3. Call `action_mark_done()` inside the background job when finished:
+
+```python
+def _process_in_background(self):
+    try:
+        # ... do work ...
+        pass
+    finally:
+        self.action_mark_done()
+```
```
amh-mw: This approach leaves room for a downstream developer to mess up error handling. What about a method decorator instead?

```suggestion
def action_process(self):
    self.with_delay()._process_in_background()

@queue_job_is_running
def _process_in_background(self):
    # ... do work ...
```

Failed jobs seem likely to goof up the transaction, so the decorator would need to `savepoint` and `roll back`?

Alternate approach: `is_job_running` as computed field?

amh-mw (changes requested):
Conceptually, I really like this module, as I have a number of long-running jobs that would enjoy the visibility.

Saran440:
@amh-mw Thank you for your feedback. I will check it.

Saran440:
Update `is_job_running` to be a compute field.
Other module just inherit `queue.job.status.mixin` and it ready to use.

Example Code
```
class AccountMove(models.Model):
    _name = "account.move"
    _inherit = ["account.move", "queue.job.status.mixin"]

    def action_confirm_queue(self):
        return self.with_delay().action_post()

```

queue_job_alert_running/models/queue_job_status_mixin.py
```diff
@@ -0,0 +1,56 @@
...
+
+#: queue.job states that are not terminal yet (done / cancelled / failed).
+RUNNING_STATES = (WAIT_DEPENDENCIES, PENDING, ENQUEUED, STARTED)
+
+
+class QueueJobStatusMixin(models.AbstractModel):
+    _name = "queue.job.status.mixin"
+    _description = "Queue Job Status Mixin"
+
+    is_job_running = fields.Boolean(compute="_compute_is_job_running")
```
amh-mw: Change Boolean `is_job_running` field to Char `jobs_running` to include the name(s) of running job(s)?

queue_job_alert_running/models/queue_job_status_mixin.py
```diff
@@ -0,0 +1,56 @@
...
+        """True when at least one non-terminal queue.job targets this record."""
+        target_ids = set(self.ids)
+        if not target_ids:
+            self.is_job_running = False
+            return
+
+        jobs = (
+            self.env["queue.job"]
+            .sudo()
+            .search_fetch(
```
amh-mw: Instead of `search_fetch` here, bypass ORM and use [PostgreSQL native JSON operators](https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-JSONB-OP-TABLE) to check for record intersection instead of querying all records for the model? Something like `records->'ids' ?| %(self.ids)s`? (Pseudo-SQL, certainly wrong.) That might want an index.

amh-mw (changes requested):
Rename the module to `queue_job_is_running` or `queue_job_banner`?

Saran440:
Fixed.
1. Rename module to `queue_job_is_running`
2. Change field name `is_job_running` from boolean to `running_job_names` char
3. Compute with Raw SQL for performance

@amh-mw Thank you for your time :)

Issue #857, opened 2025-12-08
[RFC] mechanism to configure the root channel capacity independently of other subchannels


sbidoul:
**Is your feature request related to a problem?**

The queue_job channels configurations (ODOO_QUEUE_JOB_CHANNELS) is used to configure the root channel capacity, which in turn is closely related to the number of available Odoo workers. It is also used to configure subchannels which is used to control concurrency across different job types.

When configuring different environments (prod, staging, ...) the number of available Odoo workers varies across environment so we want to change the root channel capacity per environment. However, to do that we also need to update the full subchannels configuration, which leads to duplication in configurations.

**Describe the solution you'd like**

A mechanism to configure the root channel capacity independently of the subchannels configuration.

**Describe alternatives you've considered**

Copy the full configuration in each environment.

amh-mw:
I use Terraform to configure my containers, with the environment getting set like:

```terraform
{"name": "ODOO_QUEUE_JOB_CHANNELS", "value": "root:${max(local.web_workers - 2, 1)}:subcapacity=1"}
```

Not sure what could be better than configuration management variable interpolation.

I had previously attempted to use `server_environment` from [OCA/server-env](https://github.com/OCA/server-env), but suffered from copy pasta as described. A more general YAML merge or interpolation feature might fit well there.

hoangtrann:
To me, managing root capacity is always about how much resources we want to borrow for queue job. It needs to tie with the web workers so something as @amh-mw suggest is definitely nice. Scaling subchannel capacity can be percentage of root to avoid configuration duplication. Capacity can be defined as fixed number or percentage, similar to what AWS has with their scaling policy. If we are introducing independent capacity to subchannel, then properly we need a `max_queue_workers` to not exceeding the total web workers in any case, and control what contribute to the total capacity.

sbidoul:
Another reason to configure the root channel independently is if we want to update it dynamically in an autoscaling scenario.

PR #951, opened 2026-07-21
[DRAFT] queue_job: Configure channel capacity relative to parent.

amh-mw:


amh-mw:
May be an avenue of escape from #857.

PR #954, opened 2026-07-23
[IMP] queue_job: add MaxRetryJobError

simahawk:
Prior to this change it was impossible to catch the specific case of max retry error w/o ispecting the attributes of the exception.


amh-mw (approved):
Code review only, LGTM.

sbidoul:
Looks good but curious in which case you want to catch that exception.

PR #937, opened 2026-05-19
[IMP] queue_job_batch: Add batch execution_time field.

amh-mw:

PR #734, opened 2025-01-09
[16.0][IMP] queue_job: Add error handler when job fails

trisdoan:

PR #955, opened 2026-07-24
[19.0][IMP] queue_job: Implement on fail hook

grindtildeath:
Allows to execute a model function when the job fails and will not be retried.

### DIFF
--- a/queue_job/controllers/main.py
+++ b/queue_job/controllers/main.py
@@ -186,6 +186,7 @@ class RunJobController(http.Controller):
                 vals = cls._get_failure_values(job, traceback_txt, orig_exception)
                 job.set_failed(**vals)
                 job.store()
+                job.on_fail(vals)
                 buff.close()
             raise
 
--- a/queue_job/models/queue_job.py
+++ b/queue_job/models/queue_job.py
@@ -490,3 +490,6 @@ class QueueJob(models.Model):
             time.sleep(job_duration)
         if commit_within_job:
             self.env.cr.commit()  # pylint: disable=invalid-commit
+
+    def _test_on_fail(self, **kw):
+        pass
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -431,6 +431,11 @@ class Job:
         self.job_config = (
             self.env["queue.job.function"].sudo().job_config(self.job_function_name)
         )
+        on_fail_method_name = self.job_config.on_fail_method_name
+        if on_fail_method_name:
+            if not _is_model_method(getattr(self.recordset, on_fail_method_name, None)):
+                raise TypeError("Job accepts only methods of Models")
+            self.on_fail_method_name = on_fail_method_name
 
         self.state = PENDING
 
@@ -868,6 +873,11 @@ class Job:
         for k, v in kw.items():
             if v is not None:
                 setattr(self, k, v)
+
+    def on_fail(self, fail_vals):
+        on_fail_func = getattr(self.recordset, self.on_fail_method_name, None)
+        if on_fail_func:
+            on_fail_func(**fail_vals)
 
     def __repr__(self):
         return f"<Job {self.uuid}, priority:{self.priority}>"
--- a/queue_job/models/queue_job_function.py
+++ b/queue_job/models/queue_job_function.py
@@ -29,7 +29,8 @@ class QueueJobFunction(models.Model):
         "related_action_func_name "
         "related_action_kwargs "
         "job_function_id "
-        "allow_commit",
+        "allow_commit "
+        "on_fail_method_name",
     )
 
     def _default_channel(self):
@@ -48,6 +49,10 @@ class QueueJobFunction(models.Model):
         comodel_name="ir.model", string="Model", ondelete="cascade"
     )
     method = fields.Char()
+    on_fail_method = fields.Char(
+        help="Model function to be called if the job is failed and will not be "
+        "retried.",
+    )
 
     channel_id = fields.Many2one(
         comodel_name="queue.job.channel",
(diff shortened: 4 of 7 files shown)


simahawk:
Just for the record, there were 2 other attempts to manage this
- https://github.com/OCA/queue/pull/674
- https://github.com/OCA/queue/pull/734

Your approach seem cleaner tho.

queue_job/controllers/main.py
```diff
@@ -186,6 +186,7 @@ def retry_postpone(job, message, seconds=None):
                 vals = cls._get_failure_values(job, traceback_txt, orig_exception)
                 job.set_failed(**vals)
                 job.store()
+                job.on_fail_hook(vals)
```
simahawk: shall we pass the exception obj too?

queue_job/models/queue_job.py
```diff
@@ -490,3 +491,6 @@ def _test_job(
             time.sleep(job_duration)
         if commit_within_job:
             self.env.cr.commit()  # pylint: disable=invalid-commit
+
+    def _test_on_fail_hook(self, **kw):
```
simahawk: If only for testing,this is not needed

simahawk (commented):
overall LGTM

florentx (commented):
CI fails on `test_queue_job` add-on.

simahawk:
IMO the test failure is correct because it runs in a temporary env that is not rolled back.
You should mock `in_temporary_env` to use the same cursor.

grindtildeath:
@simahawk That is what I did, but it was still :red_circle:  because of the logger in the controller, explicitely muting it makes it :green_circle: 

@guewen Can you please have a look? :pray:

queue_job/models/queue_job.py
```diff
@@ -77,6 +77,7 @@ class QueueJob(models.Model):
 
     model_name = fields.Char(string="Model", readonly=True)
     method_name = fields.Char(readonly=True)
+    on_fail_method_name = fields.Char(readonly=True)
```
amh-mw: I would appreciate consistency in naming: on_fail_hook + on_fail_method + on_fail_method_name is going to make some greps miss.

amh-mw (changes requested):

queue_job/models/queue_job.py
grindtildeath: Hello @amh-mw

What would you suggest?

As I take naming quite seriously, here was my rationale:
- on_fail_method is a reference to the method (eg when call `with_delay` )
- on_fail_method_name is the name of the method (which must be on the same model than the queue job)
- on_fail_hook is the internal function of the job that will be calling the on_fail_method

queue_job/models/queue_job.py
grindtildeath: I get `AttributeError: <class 'odoo.addons.queue_job.models.queue_job.QueueJob'> does not have the attribute '_test_on_fail_hook'` if I remove it

queue_job/controllers/main.py
grindtildeath: Is there any added value to pass the raw exception instead of what's been extracted from it in `_get_failure_values` ❓

queue_job/models/queue_job.py
amh-mw: I might rename all of them (plus `on_fail_func` that I just noticed) to `on_fail`. They're on different classes or are method arguments, so I am 85% confident they won't collide. 

In honor of this pull request, I'm going to rename `_on_done` in https://github.com/OCA/queue/pull/937 to `check_done` to reserve the `on_` prefix for hooks.

queue_job/controllers/main.py
simahawk: yes, you can inspect the exception for additional info. Not a must have tho.

queue_job/models/queue_job.py
simahawk: `on_fail` instead of `on_fail_hook` make sense. For the rest, is true that `func` is the term that is already present. But not a blocker from my POV.

queue_job/models/base.py
```diff
@@ -26,6 +26,7 @@ def with_delay(
         description=None,
         channel=None,
         identity_key=None,
+        on_fail_method=None,
```
guewen: Does it make sense to set this on a per-call basis? I think a "job function" will always have the same "on fail method", and would then configure it on `queue.job.function`, it simplifies the thinking model (always the same behavior) and it simplifies the code (need to be added only in `job_config` (no propagation in all calls, no new field on `queue.job`, ...)

grindtildeath: @guewen Thanks for checking in.

I guess both implementations have their pros and cons.  :shrug: 

Anyway, since the first use I have for it doesn't require such flexibility, I did the changes in [6651362](https://github.com/OCA/queue/pull/955/commits/6651362155ea54cd04789d53cca1df96200fccee) 

Finally about the naming, since we already have a mix of _method, _func, and others I tried my best to be consistent

simahawk (approved):
LGTM

amh-mw (approved):

queue_job/models/base.py
guewen: Thanks @grindtildeath

guewen: There is a failing test, probably a small thing in the assertion:

```python
   File "/__w/queue/queue/test_queue_job/tests/test_delay_mocks.py", line 303, in test_mock_with_delay
    self.assertDictEqual(
AssertionError: {'priority': 15, 'eta': 15, 'max_retries': [105 chars]f30>} != {'channel': 'root.test', 'description': 'Te[129 chars]: 15}
  {'channel': 'root.test',
   'description': 'Test',
   'eta': 15,
   'identity_key': <function identity_exact at 0x7f18df221f30>,
   'max_retries': 1,
+  'on_fail_method': None,
   'priority': 15}
```

ivantodorovich (approved):
LG but ci is red

simahawk:
Rebased and squashed. @guewen good for you?

guewen:
/ocabot merge minor

Yes, thanks!

PR #959, opened 2026-08-14
[IMP] queue_job_cron - allow commit within a cron that runs as queue job

baimont:
Related to https://github.com/OCA/queue/pull/899

Works as the new option on the job function but on the cron, false by default, allow to run the job in a new transaction


baimont:
take a look, @sbidoul ?

Issue #765, opened 2025-04-29
[RFC] Hot reload channels configuration (size, pause, ...) stored in DB


guewen:
**Is your feature request related to a problem?**

The configuration of the channels is stored in a configuration file/environment variable and need a restart of the jobrunner to be applied. This proposition aims to bring the possibility to change the configuration on the fly from the UI, which also offers new options such as pausing/resuming a channel.

**Describe the solution you'd like**

Add fields on `queue.job.channel`: capacity, sequential, throttle, running (boolean).

In the jobrunner startup, read configuration from model `queue.job.channel`, fallback on the current implementation based on file/environment variable.

In the `queue.job.channel` model, when one of the configuration field that concerns the jobrunner is changed, send a `pg_notify` in the same postgres channel (`queue_job`), but instead of a job uuid, send `reload`.

In the Jobrunner, when the payload is `reload`, go through a cycle close databases/replace the channel manager by a fresh one/configure channels/initialize databases (maybe not necessary to close/initialize database but we have to go through the steps to load all pending jobs in the channels).

Add the "paused" support in the Channel:
* New `is_paused=False` parameter in Channel
* A method is_paused() returns True if the current channel or any parent is paused
* `has_capacity()` returns False if `is_paused()` returns True

**Describe alternatives you've considered**

/

**Additional context**

The cycle to update the channels can probably be optimized to prevent reloading all at once, but changes on the configuration should be are enough not to be a problem. Maybe if the only changes are in pause/resume, we can change the channels in place instead of recreating all the channels.

guewen:
I missed the biggest point which is a problem for this implementation: channels configuration is loaded before loading databases, and it may be different from database to database.

sbidoul:
Sometimes I have this idea of building a small web UI to view channels, some counters, and why not manage channels. This could take the form of a little web server thread in the job runner worker, serving a simple single page app.

Then we could think about persistence of the channels config.

flotho:
@AmalBenAissa @DorianMAG @JulienMartinez @gaelTorrecillas

simahawk:
> I missed the biggest point which is a problem for this implementation: channels configuration is loaded before loading databases, and it may be different from database to database.

Question: why is the loading before db is ready needed? In any case you cannot run jobs until the instance is up and running and able to answer requests. No?

sbidoul:
[...]

PR #960, opened 2026-08-14
[18.0][IMP] queue_job: add pause option on channels

guewen:
Add a new "paused" option on channel configuration. A paused channel is equivalent to a channel with no capacity: it yields no job and also blocks all its subchannels.

It still needs a restart of the jobrunner, so it does not allow dynamic pause/resume, but hot reload may be added later in OCA/queue#765.

It still is an improvement over setting the capacity at 0, because it keeps the capacity in the configuration (removing the option in the config restores the initial capacity).

### DIFF
--- a/queue_job/jobrunner/channels.py
+++ b/queue_job/jobrunner/channels.py
@@ -402,9 +402,14 @@ class Channel:
     with a capacity of 1. It is also possible to dedicate a channel with a
     limited capacity for application-autocreated subchannels
     without risking to overflow the system.
+
+    A paused channel does not process any job until it is resumed. All subchannels
+    are blocked with their parent channel.
     """
 
-    def __init__(self, name, parent, capacity=None, sequential=False, throttle=0):
+    def __init__(
+        self, name, parent, capacity=None, sequential=False, throttle=0, paused=False
+    ):
         self.name = name
         self.parent = parent
         if self.parent:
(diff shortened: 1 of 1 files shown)


queue_job/jobrunner/channels.py
```diff
@@ -434,11 +440,13 @@ def configure(self, config):
         * capacity
         * sequential
         * throttle
+        * paused
         """
         assert self.fullname.endswith(config["name"])
         self.capacity = config.get("capacity", None)
         self.sequential = bool(config.get("sequential", False))
         self.throttle = int(config.get("throttle", 0))
+        self.paused = int(config.get("paused", False))
```
amh-mw: bool?

amh-mw (changes requested):
Code review only, LGTM.

queue_job/jobrunner/channels.py
guewen: Yes of course, updated

amh-mw (approved):

sbidoul (approved):

hoangtrann:
loving the idea. are we able to pause and got it applied right away or would it require a restart?

guewen:
> loving the idea. are we able to pause and got it applied right away or would it require a restart?

@hoangtrann right now it needs a restart. Trying something with hot reload here https://github.com/OCA/queue/pull/962 it somehow works with caveats but still need some work, and it changes the channel config to be per db instead of global (in my wip, you can choose between the current server-side static config that do not support hot reload or the per db config that supports hot reload).

PR #830, opened 2025-09-23
[18.0][MIG] queue_job_web_notify from v16

hhgabelgaard:
Std migration from 16.0

Replaces #829

PR #961, opened 2026-08-14
[19.0][MIG] queue job web notify

hhgabelgaard:
Forward port of #830


amh-mw (approved):
Code review only, LGTM. Bonus points would be awarded for writing the first unit test for this module.

PR #966, opened 2026-08-19
[19.0][MIG] queue_job_subscribe

hhgabelgaard:
Std migration


imlopes (commented):
I think it could be better to split this PR in 2, one by module.

ivantodorovich (approved):
agree about the split, one module per PR

hhgabelgaard:
Replaced by #975

PR #967, opened 2026-08-20
[16.0][FIX] queue_job: rebind args/kwargs lazily to the job's current cursor

eantones:
Supersedes #926.

Job.in_temporary_env() rebinds self.recordset to the inner cursor via the @env.setter, but leaves self.args and self.kwargs on the outer worker cursor where _prevent_commit is patched. A cr.commit() reached through any recordset argument then raises RuntimeError: Commit is forbidden in queue jobs even with allow_commit=True. Typical case: a connector exporter taking a binding and a related record, committing inside binder.bind_export(...).

Introduced by #910.

Fix: args and kwargs are stored in _args/_kwargs and exposed as properties that rebind any recordset inside (recursing into lists, tuples and dicts) to self.env.cr on access — only the cursor changes, uid/su/context are kept. in_temporary_env() goes back to a plain env swap.

Regression test in test_queue_job passes the same recordset as positional arg, list element and dict value, checks cursor identity in all three, then commits. Run through RunJobController._runjob.

The same issue exists on 18.0 and 19.0; ports will follow once this is merged.

### DIFF
--- a/queue_job/job.py
+++ b/queue_job/job.py
@@ -13,8 +13,29 @@ from datetime import datetime, timedelta
 from random import randint
 
 import odoo
+from odoo.models import BaseModel
 
 from .exception import FailedJobError, NoSuchJobError, RetryableJobError
+
+
+def _rebind_to_cr(value, cr):
+    """Rebind any BaseModel inside ``value`` to the cursor ``cr``.
+
+    Recurses into lists, tuples and dicts. Preserves uid/su/context of each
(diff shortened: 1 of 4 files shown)


eantones:
Hi @guewen, no worries about the delay. Here it is as its own PR, same code as the branch you preferred, unchanged. Thanks again for the suggestion.

guewen (approved):

guewen:
Thanks!

PR #968, opened 2026-08-21
[18.0][FIX] queue_job: rebind args/kwargs lazily to the job's current cursor

eantones:
18.0 port of #967, as promised.


guewen (approved):
Thanks

PR #974, opened 2026-08-31
[19.0][FIX] queue_job: skip on_fail when unconfigured

Ricardoalso:
Job.on_fail_method_name was only set as an instance attribute when the job function configured on_fail_method, so on_fail() raised AttributeError for every job that didn't opt into it.

Introduced in https://github.com/OCA/queue/pull/955


guewen (approved):
Thanks

Issue #965, opened 2026-08-18
[RFC] expose jobrunner status as prometheus metrics


sbidoul:
It would be useful to expose the jobrunner status as prometheus metrics for monitoring.

metrics could include, for each channel being a label:
- capacity
- jobs in channel
- running jobs
- failed jobs
- paused
- throttle

This means starting a little web server next to the jobrunner and protecting access to the Channels data structure with proper multi-thread locking.

PR #976, opened 2026-08-31
Expose jobrunner metrics

sbidoul:
This is a first draft towards #965 

This turns out to be fairly simple. A few things worth noting:

- The http server runs in the same thread and event loop as listen/notify job events, so there are no concurrency issues while accessing the channels stats
- I added the set of jobs waiting for dependencies on each channel for completeness and convenient access to that stat, similar to what is done for failed jobs.
- I used the `prometheus-client` external library. It's currently required even if the metrics server is not configured. This could be improved by mocking the lib and making it mandatory only for users enabling the metrics server. Or make it a required dependency (which is my preference, for simplicity).
- This still needs tests and docs.

Test by setting `ODOO_QUEUE_JOB_METRICS_PORT=8080` before running Odoo then `watch curl -s http://localhost:8080`.

Assisted-By: Pamina (my cat)

### DIFF
--- a/queue_job/jobrunner/channels.py
+++ b/queue_job/jobrunner/channels.py
@@ -2,6 +2,8 @@
 # Copyright 2015-2016 Camptocamp SA
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
 import logging
+import math
+import weakref
 from collections import namedtuple
 from functools import total_ordering
 from heapq import heappop, heappush
@@ -9,6 +11,7 @@ from weakref import WeakValueDictionary
 
 from ..exception import ChannelNotFound
 from ..job import CANCELLED, DONE, ENQUEUED, FAILED, PENDING, STARTED, WAIT_DEPENDENCIES
+from . import metrics
 
 NOT_DONE = (WAIT_DEPENDENCIES, PENDING, ENQUEUED, STARTED, FAILED)
 JobSortingKey = namedtuple("SortingKey", "eta priority date_created seq")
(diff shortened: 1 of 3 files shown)


queue_job/jobrunner/channels.py
```diff
@@ -413,10 +415,38 @@ def __init__(self, name, parent, capacity=None, sequential=False, throttle=0):
...
         self._failed = set()
+        self._waiting_dependencies = set()
         self._pause_until = 0  # utc seconds since the epoch
         self.capacity = capacity
         self.throttle = throttle  # seconds
         self.sequential = sequential
+        self._metrics_labels = {
+            "channel": self.fullname,
+            "root": not bool(self.parent),
+            "leaf": not bool(self.children),
```
sbidoul, line 436: Maybe move this to configure? Because the `leaf` label is probably not correct if/when children are added after the parent.

queue_job/jobrunner/channels.py
```diff
@@ -409,14 +412,70 @@ def __init__(self, name, parent, capacity=None, sequential=False, throttle=0):
...
+            return
+        metrics.channel_capacity.remove_by_labels(self._metrics_labels)
+        metrics.channel_pending.remove_by_labels(self._metrics_labels)
+        metrics.channel_running.remove_by_labels(self._metrics_labels)
+        metrics.channel_failed.remove_by_labels(self._metrics_labels)
+        metrics.channel_waiting_dependencies.remove_by_labels(self._metrics_labels)
+
+    def _capacity_gauge(self) -> float:
+        if self.capacity is None:
+            return math.inf
```
AnizR, line 465: I wonder how will it look in the output metric.
Maybe return -1.0 if capacity is not set?

AnizR (commented):
Seems promising 🚀

queue_job/jobrunner/channels.py
sbidoul: Yes I have not tested that yet.

PR #977, opened 2026-08-31
[19.0][MIG] queue_job_batch, test_queue_job_batch: Migration to 19.0

desarrollonextpro:
Migration of `queue_job_batch` and `test_queue_job_batch` to 19.0. Part of #831.

## Changes required by 19.0

- `res.groups.category_id` no longer exists: the security group now uses `privilege_id` pointing at `queue_job.privilege_queue_job`, introduced by the `queue_job` migration.
- Search view: the `<group expand="0" string="Group By">` wrapper was removed in 19.0; the group-by filters are declared flat after a `<separator/>`.
- `Store.add()` no longer accepts a dict and `_to_store()` takes two positional arguments: the batch now declares `_to_store_defaults()` like the core models, which also removes 15 lines of manual field reading.
- `res.users`: `store.add({...})` replaced by `add_global_values()`.
- `webclient` controller: in 19.0 it is a `classmethod` with the `(store, name, params)` signature, and the systray is requested by name rather than through a keyword flag.
- `Store.esm.js`: `initMessagingParams` does not exist in 19.0. Replaced by the core pattern, a patch on `initialize()` calling `fetchStoreData()`. Worth noting: the old patch overrode a getter nothing calls, so the systray request was silently never sent.
- `QueueJobBatchMenu.esm.js`: `fetchData()` replaced by `fetchStoreData()`.
- The manifest declared `installable` inside the `assets` dict; cleaned up.

## Not included on purpose

The multi-company `ir.rule` still uses `('company_id', 'child_of', [user.company_id.id])` while `queue_job` moved to `('company_id', 'in', company_ids)` in `decd9c93` (13.0). That is a pre-existing inconsistency, not a 19.0 breakage; changing a security rule inside a migration would widen the scope. It should go in a separate `[FIX]`.

## Testing

`test_queue_job_batch` installs and its tests pass (3 tests, 0 failed), covering `get_new_batch`, the `job_batch` context key, `check_state`, `completeness` and `set_read`.

The frontend was **not** validated in a browser: the systray menu rendering, the `ColumnProgress` bars and the bus-driven counter update still need manual verification.

cc @etobella


etobella:
I think you didn't follow all OCA migration guidelines, as I don't see previous commits. Can you check?

PR #985, opened 2026-09-15
[FIX] queue_job: validate JobSerialized base_type again

HoloborodkoBohdan:
Odoo 19 renamed `Field._setup_attrs` to `_setup_attrs__` (`odoo/orm/fields.py`), and the old name has no remaining reference in core. `JobSerialized` still overrode `_setup_attrs`, so the override was never called and an unsupported `base_type` was silently accepted instead of raising `ValueError`.

This renames the override and adds a test: `base_type=dict` sets up fine, `base_type=int` raises. The test fails on the current 19.0 branch.

All `JobSerialized` fields in the repository use `dict`, `tuple` or `models.BaseModel`, so re-enabling the check does not affect existing fields.


Jordi-Buitrago (approved):
LGTM!

The fix looks correct and the regression test covers the issue.
Tested locally.

amh-mw (approved):
Code review only, LGTM. Confirmed upstream `_setup_attrs__` rename. [^1]

[^1]: https://github.com/odoo/odoo/commit/13373d44965acbfa4fbac90b0053e3179fefe154

PR #986, opened 2026-09-15
[FIX] queue_job: notify managers holding the group through implication

HoloborodkoBohdan:
In Odoo 19 res.users.group_ids only holds explicitly assigned groups; implied ones are exposed through all_group_ids. Users who got Job Queue Manager through another group were no longer subscribed to failed jobs.

`_subscribe_users_domain` searched `[("group_ids", "=", group.id)]`. In Odoo 19 `res.users.group_ids` only holds explicitly assigned groups; implied groups are exposed through `all_group_ids` (searched as `group_ids.all_implied_ids`).

As a result, users who get *Job Queue Manager* through another group were no longer subscribed to failed jobs. The domain now uses `[("all_group_ids", "in", group.ids)]`.

Added a test with a direct manager and a manager through an implying group; it fails on the current 19.0 branch.


amh-mw (approved):
Code review only, LGTM. Confirmed upstream `all_group_ids` behavior change. [^1]

[^1]: https://github.com/odoo/odoo/commit/34a50e83e654017520c0add727215bd7b528ae19

